PR #3, opened 2023-02-27
Bring back the actions

jbygdell:
* Static code tests
* Build containers on PR
* Scan PR containers for vulnerabilities
* Integration test
* Code linting for GO code


.github/workflows/ghcr-actions.yml
```diff
@@ -0,0 +1,42 @@
...
+          keep-at-least: 1
+          token: ${{ secrets.GHCR_TOKEN }}
+          timestamp-to-use: updated_at
+
+      - name: Delete all containers older than treee months, using a wildcard
+        uses: snok/container-retention-policy@v2.0.0
+        with:
+          image-names: sensitive-data-archive
+          filter-tags: v*
+          cut-off: Three months ago UTC
```
pontus, line 37: Aren't these "released" versions, do we really want to clean those?

jbygdell: Given that we get dependabot update every week, images that are more than 3 months old can be considered obsolete.

pontus: I haven't kept track, how many of those are security updates?

My understanding is that most installations currently upgrade the pipeline more seldom than every three months. So, if these are the "released" artifact, I'm afraid we may run into breakages because of harsh cleaning (e.g. a pod failing, being rescheduled on another node that does not have image available and can't pull it because it's no longer available).

I also think we should keep "released" versions around for historical reason (e.g. being able to see the actual image used at a certain point in time) and the resource requirement for keeping those around should be very, very low.

(If these aren't the primary produced artifact, this does not apply.)

jbygdell: If we see that we have to many images around we might prune the patch version related to older minor tags and only keep the major and minor versions.

blankdots (commented):
any reason to have it like: `ghcr.io/${{ github.repository }}:sha-${{ github.sha }}-rabbitmq` 
and not like `ghcr.io/${{ github.repository }}:rabbitmq-sha-${{ github.sha }}`

.github/workflows/ghcr-actions.yml
pontus: No objection.

jbygdell:
> any reason to have it like: `ghcr.io/${{ github.repository }}:sha-${{ github.sha }}-rabbitmq` and not like `ghcr.io/${{ github.repository }}:rabbitmq-sha-${{ github.sha }}` ?

Not really, just that the common semantics seems to be `IMAGENAME:VERSION-VARIANT`

blankdots (dismissed):

pontus (approved):

PR #45, opened 2023-03-22
Add missing command in workflow

pontus:


jbygdell (approved):

pontus:
Fixes missing command for `docker pull`. 

I'm not sure from output from action (like https://github.com/neicnordic/sensitive-data-archive/actions/runs/4488694263/jobs/7893618008) if this is enough or if `github.event.number` is missing for the actual merge run and something like https://github.com/marketplace/actions/find-current-pull-request will also be needed.

PR #61, opened 2023-04-06
Rework postgres image

### LINKED ISSUE #13: Update database image to use Postgresql 14 or higher
*This will be a breaking change* - automatic upgrade will not be possible

A/C
- [ ]  base image uses Postgresql 14 or higher

This makes it easier to handle the certificates since the key doesn't need to be owned by the process UID anymore.

jbygdell:
* uses `postgres:15-alpine3` as baseimage
* reworked the entrypoint script
* Includes all changes from `sda-db` up to commit `e0d9e1`
* added a test suite for testing the queries on a per service basis.
* Added a generic runner script that can be reused for other test suites.
* shellcheck all scripts

Closes #13


postgresql/entrypoint.sh
```diff
@@ -0,0 +1,142 @@
...
+# If already initiliazed, then run
+if [ -s "$PGDATA/PG_VERSION" ]; then
+	migrate "$@"
+
+	setup_hba_conf
+
+	exec "$@"
+fi
+
+# Otherwise, do initilization (as postgres user)
```
norling, line 108:
```suggestion
# Otherwise, do initialization (as postgres user)
```

norling (approved):
Looks good apart from some very minor things! :) I'd also prefer "from" to be uppercase in the queries, but that's more of an opinion than a problem ^^

norling (approved):

PR #94, opened 2023-05-01
Begin the merge process

jbygdell:
The merged code resides under `sda` 

This PR merges the following:
* sda-common
* sda-s3inbox

A new functionality test for the SDA suite is created, currently it only tests the `s3inbox` in "happy mode" :smile:
*  Uploads 2 files
* Reuploads one file with the same name
* Reuploads one file with another name
* Validates the number of messages in the inbox queue
* validates the number of rows in the database

Merging this PR will allow us to archive the `sda-s3proxy` and `sda-common` repositories.


blankdots (approved):

.github/integration/sda-integration.yml
```diff
@@ -0,0 +1,133 @@
...
+      retries: 3
+    image: ghcr.io/neicnordic/sensitive-data-archive:PR${PR_NUMBER}-rabbitmq
+    ports:
+      - "15672:15672"
+    restart: always
+    volumes:
+      - rabbitmq_data:/var/lib/rabbitmq
+
+  minio:
+    image: minio/minio:latest
```
norling, line 60: Should we use `latest`? They have changed stuff to break our envs before.

norling (approved):
Some thoughts in comments! Then, this should probably have been split into smaller PRs where one is focused on just moving things, and then have changes (unrelated to the move) in separate PRs.
Also - before moving further in the merge, should we take the time to make a good integration testing system? There's different testing styles used all over (bash, docker, suite, testing.T, etc.) and this might be a good place to create and enforce a standard!

jbygdell:
> Some thoughts in comments! Then, this should probably have been split into smaller PRs where one is focused on just moving things, and then have changes (unrelated to the move) in separate PRs. Also - before moving further in the merge, should we take the time to make a good integration testing system? There's different testing styles used all over (bash, docker, suite, testing.T, etc.) and this might be a good place to create and enforce a standard!

Yes, static tests (go test) should use `suite` mainly since only `helper_test.go` and `userauth_test.go` doesn't use it yet.

Integration tests should run against the real containers, and having the tests as shell scripts is a convenient way of simulating user actions.

norling:
Yeah, I don't mind using bash scripts, I was mostly noting that there isn't a good general solution to integration tests. Most importantly we should really have a system where we can restore the system to a known state before all tests to make them independent.

PR #150, opened 2023-05-25
Don't run tests om push to `main`

jbygdell:
Push to main only happens when we merge a PR and the tests have already been run in the PR branch.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,6 +2,7 @@ name: Code tests
 
 on:
   push:
+    branches-ignore: [main]
 
 jobs:
 


kjellp:
Maybe a small but still possible risk of not re-running tests on actual push to master:
- tests pass at a fresh PR branch
- time passes and changes to master happen 
- when the actual merge happens quite delayed, they may not work correctly with the further evolved master.

Currently we are double checking the results og the "on push to master" tests after a PR merge, exactly for this reason.

jbygdell:
> Maybe a small but still possible risk of not re-running tests on actual push to master:
> 
> * tests pass at a fresh PR branch
> * time passes and changes to master happen
> * when the actual merge happens quite delayed, they may not work correctly with the further evolved master.
> 
> Currently we are double checking the results og the "on push to master" tests after a PR merge, exactly for this reason.

There is a very small chance that scenario will happen when the `Require branches to be up to date before merging` rule is enabled.

blankdots (approved):

norling (approved):

kostas-kou (approved):

Issue #14, opened 2023-03-02
Merge Golang based apps


jbygdell:
Expected structure:
```
cmd/<APPNAME-1>
cmd/<APPNAME-2>
shared/<COMPONENTNAME-1>
shared/<COMPONENTNAME-2>
Dockerfile
```

- [x] auth
- [ ] download
- [x] re-encryption service

A/C
- [ ] use similar setups as in sda-pipeline
- [ ] ensure all tests keep working

PR #249, opened 2023-08-09
Merge FEGA related components from pipeline into SDA

jbygdell:
This merges the components needed for a federated setup from `sda-pipeline` into `sda` i.e. finalize, ingest, intercept, mapper, verify.
* Backup is merged with finalize in order to simplify the message flow.
* Static code tests uses dockerized services instead of mocked backends for DB, MQ and S3.
* Functionality tests track the entire path from upload to complete ingestion and dataset release/deprecation.
* Removes code and functionality tests for the `sda-pipeline` folder.
* Updates the helm charts

related to #14


pahatz:
Could you elaborate a bit on how these reflect on the commits of the PR?

> Static code tests uses dockerized services instead of mocked backends for DB, MQ and S3.
> Functionality tests track the entire path from upload to complete ingestion and dataset release/deprecation.

also why is this needed?

> Removes code and functionality tests for the sda-pipeline folder.

This could maybe benefit from being an epic or at least multiple PRs but it's a bit late now :)

I was also expecting a whole different thing from the title of the ticket. Are we archiving the sda-pipeline repo after this? Would that mean that we're not keeping any history of the merged repo?

blankdots (commented):
> Backup is merged with finalize in order to simplify the message flow.

so the `finalize` service also performs the backup if a backup storage is configured, do i understand this correctly ?
seems to also include the `sftp` backup type which i assume is for mirroring

When deploying shall we set up one or more `finalize` services, each doing different functionalities ? i assume one (maybe 2 for mirroring use case ?), but want to be sure 

shall we include this in the description and maybe adjust `finalize.md` ?

jbygdell:
> > Backup is merged with finalize in order to simplify the message flow.
> 
> so the `finalize` service also performs the backup if a backup storage is configured, do i understand this correctly ? seems to also include the `sftp` backup type which i assume is for mirroring

Precisely, the merge is primarily because there is no dedicated `backup` message.
`sftp` is included since it exists as a storage backed.

> 
> When deploying shall we set up one or more `finalize` services, each doing different functionalities ? i assume one (maybe 2 for mirroring use case ?), but want to be sure

This PR is targeted at the FEGA usecase and as such `mirroring` is not in the scope.
If a dedicated `mirroring` step is required in addition to backup we might have to bring the `backup` service back under a different name.

> shall we include this in the description and maybe adjust `finalize.md` ?

Yes the readme needs to be updated.

jbygdell:
> Could you elaborate a bit on how these reflect on the commits of the PR?
> 
> > Static code tests uses dockerized services instead of mocked backends for DB, MQ and S3.
> > Functionality tests track the entire path from upload to complete ingestion and dataset release/deprecation.
> 
> also why is this needed?

Static tests usually use mocked backeds, but mocking an MQ backend is all but impossible. Even the libray repo uses live backends in their tests.
 
> > Removes code and functionality tests for the sda-pipeline folder.
> 
> This could maybe benefit from being an epic or at least multiple PRs but it's a bit late now :)
> 
> I was also expecting a whole different thing from the title of the ticket. Are we archiving the sda-pipeline repo after this? Would that mean that we're not keeping any history of the merged repo?

How to deal with the history was solved when this repo was created, and all standalone `sda-XXX` repos that are merged into this will be archived when the merge of a repo is completed. Currently only `sda-common` and `sda-s3proxy` is fully merged.

.github/workflows/functionality.yml
```diff
@@ -72,41 +69,6 @@ jobs:
             bash -x "$runscript";
           done
 
-  sda-pipeline:
```
pontus, line 75: I don't see where we do functionality/integration testing of sda (formerly sda-pipeline) now?

pontus:
Also unsure if I understand why sda-pipeline still remains with migrated contents here,

.github/workflows/functionality.yml
jbygdell: They are triggerede here: https://github.com/neicnordic/sensitive-data-archive/blob/80f8fbf2326fc860fd2c1620afb7c32f4959834e/.github/workflows/build_pr_container.yaml#L226

jbygdell:
> Also unsure if I understand why sda-pipeline still remains with migrated contents here,

Not every component is migrated yet and until that the old folder will still be around.

.github/workflows/functionality.yml
pontus: Thanks.

pontus (approved):

aaperis (commented):
looks nice so far, but my feeling is that some refactoring could have been split off to a separate PR so that reviewing would be easier/faster :-). I am half-way through it though

jbygdell:
Will split into separate parts.

PR #276, opened 2023-08-31
[sda-auth] update to go1.20

aaperis:
Update to go1.20 and update modules from #264, #265 while avoiding the toolchain bug in dependabot.

Also add a small fix in the compose file to make life easier.

### DIFF
--- a/sda-auth/dev-server/docker-compose.yml
+++ b/sda-auth/dev-server/docker-compose.yml
@@ -48,7 +48,7 @@ services:
     build:
       context: ../
       args:
-        GOLANG_VERSION: ${GOLANG_VERSION-1.18}
+        GOLANG_VERSION: ${GOLANG_VERSION-1.20}
       dockerfile: Dockerfile
     depends_on:
       oidc:
--- a/sda-auth/dev-server/docker-compose.yml
+++ b/sda-auth/dev-server/docker-compose.yml
@@ -19,6 +19,11 @@ services:
       - USERINFO_ROUTE=/userinfo
     ports:
       - 9090:9090
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:9090/.well-known/openid-configuration"]
+      interval: 5s
+      timeout: 10s
+      retries: 4
   cega:
     container_name: cega
     image: egarchive/lega-base:release.v0.2.0
@@ -47,7 +52,7 @@ services:
       dockerfile: Dockerfile
     depends_on:
       oidc:
-        condition: service_started
+        condition: service_healthy
       cega:
         condition: service_started
     environment:
--- a/.github/workflows/code-linter.yaml
+++ b/.github/workflows/code-linter.yaml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: ['1.19']
+        go-version: ['1.20']
     steps:
       - name: Set up Go ${{ matrix.go-version }}
         uses: actions/setup-go@v4
--- a/.github/workflows/functionality.yml
+++ b/.github/workflows/functionality.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: [1.19]
+        go-version: ['1.20']
     steps:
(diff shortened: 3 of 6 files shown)


jbygdell (approved):

pontus:
Do we want to do this? I did some research and the toolchain addition is a bug that should be fixed now (dependabot is having issues with parts of the sensitive-data-archive for other reasons and thus asking it to recreate the PR doesn't currently work).

I have a slight preference for not doing this as bumping to 1.21 involves work on most platforms.

kusalananda:
> Do we want to do this? I did some research and the toolchain addition is a bug that should be fixed now (dependabot is having issues with parts of the sensitive-data-archive for other reasons and thus asking it to recreate the PR doesn't currently work).
> 
> I have a slight preference for not doing this as bumping to 1.21 involves work on most platforms.

What about doing _the other updates_ that `dependabot` is suggesting, but leaving the `toolchain` update for later?

kusalananda:
In the morning meeting, we discussed moving to 1.20 (which I presume means 1.20.7, the latest 1.20 release), and applying the `dependabot` changes not related to the `toolchain`.  I believe @aaperis said he'll do something along these lines later.

sda-auth/dev-server/docker-compose.yml
```diff
@@ -19,6 +19,11 @@ services:
       - USERINFO_ROUTE=/userinfo
     ports:
       - 9090:9090
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:9090/.well-known/openid-configuration"]
```
kusalananda, line 23: Is this a real URL that will actually return something?

kusalananda (approved):

kusalananda:
But see failing test: https://github.com/neicnordic/sensitive-data-archive/actions/runs/6036698927/job/16379492436?pr=276

sda-auth/dev-server/docker-compose.yml
```diff
@@ -48,7 +48,7 @@ services:
     build:
       context: ../
       args:
-        GOLANG_VERSION: ${GOLANG_VERSION-1.18}
+        GOLANG_VERSION: ${GOLANG_VERSION-1.20.7}
```
kusalananda, line 51: I don't know if using "1.20.7" is over-specifying it.  I noticed that if one pulls the `golang:1.20` image from Dockerhub, one gets Go 1.20.7.

sda-auth/dev-server/docker-compose.yml
aaperis: yes, this is the `/.well-known/openid-configuration` endpoint of the`mockoidc`

sda-auth/dev-server/docker-compose.yml
aaperis: there was an error in the tests where the action tried to pull `1.2` instead of `1.20` so I thought that over specifying would fix it but I hadn't noticed that the error was from not having quotes in `go-version: [1.20]`. Anyhow, for documentation reason it's not bad to have this in the compose I suppose.

kusalananda: I saw that error too, and I do not understand the cause of it.  I tried to run the build locally but didn't have any problems.  It seemed like the CI tests treated `1.20` as a floating point value for whatever reason and decided to use it without the trailing zero.

aaperis: indeed, adding quotes around `1.20` solved it.

kusalananda: Weird.

sda-auth/dev-server/docker-compose.yml
```diff
@@ -43,11 +48,11 @@ services:
...
```
jbygdell, line 51:
```suggestion
        GOLANG_VERSION: ${GOLANG_VERSION-1.20}
```

jbygdell:
> But see failing test: https://github.com/neicnordic/sensitive-data-archive/actions/runs/6036698927/job/16379492436?pr=276

That test seems to fail randomly, it will be removed with the merging of the `sda-pipeline` folder.

dbampalikis (approved):

PR #295, opened 2023-09-13
Added missing properties for posix mapper

kkochel:
When running the application in federated and inbox mode as posix, the mapper container is not started.

Log error:
`{"level": "fatal", "msg": "stat /inbox/: no such file or directory", "time": "2023-09-13T09:56:45Z"}`

The error is generated by the following changes to the mapper:
```
  inbox, err := storage.NewBackend(conf.Inbox)
  if err != nil {
           log.Fatal(err)
  }
```


jbygdell:
Included in #297

PR #297, opened 2023-09-13
test

jbygdell:


.github/integration/scripts/charts/deploy_charts.sh
```diff
@@ -39,11 +39,18 @@ if [ "$1" == "sda-mq" ]; then
 fi
 
 if [ "$1" == "sda-svc" ]; then
+    inbox=s3
+    if [ "$3" == "true" ] && [ "$4" == "posix" ]; then
```
pontus, line 43: Why do we ignore inbox type if not using tls?

jbygdell: The sftp inbox that uses posix as backend needs TLS.

pontus: How about excluding the bad case (tls=false, storage=posix) instead (https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#excluding-matrix-configurations) of testing something other than what one would assume?

.github/workflows/build_pr_container.yaml
```diff
@@ -249,6 +249,13 @@ jobs:
         version: ["1.26", "1.27"]
         tls: ["true", "false"]
         storage: ["posix", "s3"]
+        exclude:
```
pontus, line 252: I haven't tested, but my reading is that it should be enough to just specify the combination to exclude (tls=false, storage=posix) and it would exclude those cases for both 1.26 and 1.27, but this should of course work as well.

pontus (approved):

dbampalikis (approved):

PR #299, opened 2023-09-14
[sda-pipeline][intercept] fix queue name

jbygdell:
Mixed case queue names is not allowed anymore

### DIFF
--- a/sda-pipeline/cmd/intercept/intercept.go
+++ b/sda-pipeline/cmd/intercept/intercept.go
@@ -124,7 +124,7 @@ func main() {
 			}
 
 			routing := map[string]string{
-				msgAccession: "accessionIDs",
+				msgAccession: "accession",
 				msgCancel:    "ingest",
 				msgIngest:    "ingest",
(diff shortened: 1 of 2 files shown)


pontus (dismissed):

kusalananda (approved):

sda-pipeline/cmd/intercept/intercept.go
```diff
@@ -124,7 +124,7 @@ func main() {
 			}
 
 			routing := map[string]string{
-				msgAccession: "accessionIDs",
+				msgAccession: "accession",
```
nanjiangshu, line 127: Is that true that the name `accessionIDs` used only for the queue name?

nanjiangshu (approved):
LGTM!

kostas-kou (approved):

sda-pipeline/cmd/intercept/intercept.go
jbygdell: `Mixed case queue names is not allowed anymore`

PR #302, opened 2023-09-15
[FEGA usecase] Merge internal components from `sda-pipeline` into `sda`

jbygdell:
* Merges all `internal` components into `sda`
* Updates s3inbox to work with the inernal components
* Removes unused stuff from broker.


sda/internal/helper/helper.go
```diff
@@ -153,6 +154,50 @@ func CreateRSAkeys(prPath, pubPath string) error {
...
+	}
+	pubKeyBytes := ssh.MarshalAuthorizedKey(publicKey)
+	err = os.WriteFile(path+"/id_rsa.pub", pubKeyBytes, 0600)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
```
aaperis, line 200: It seems this function is only called by `storage_test.go`, so better move it there? Especially since x509.EncryptPEMBlock is deprecated and therefore we wouldn't want to actually use this function anywhere else.

sda/internal/helper/helper.go
```diff
@@ -153,6 +154,50 @@ func CreateRSAkeys(prPath, pubPath string) error {
 	return nil
 }
 
+func CreateSSHKey(path string) error {
```
aaperis, line 157: It seems this function is only called by storage_test.go, so better move it there? Especially since x509.EncryptPEMBlock is deprecated and therefore we wouldn't want to actually use this function anywhere else.

jbygdell: Since the `helper.go` seems primarily used by the tests it's better to move the one function that's used in `proxy.go` to that file instead.

sda/internal/database/database.go
```diff
@@ -62,6 +62,15 @@ var FastConnectRate = 5 * time.Second
 // database during the after FastConnectTimeout.
 var SlowConnectRate = 1 * time.Minute
 
+// dbRetryTimes is the number of times to retry the same function if it fails
+var RetryTimes = 5
+
+// hashType returns the identification string for the hash type
+func hashType(_ hash.Hash) string {
+	// TODO: Support/check type
```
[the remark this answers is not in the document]
jbygdell: #321

aaperis (approved):

blankdots (commented):
small thing

blankdots (approved):

pontus (approved):

sda/internal/config/config.go
```diff
@@ -6,34 +6,24 @@ import (
...
 
-var (
-	requiredConfVars = []string{
-		"aws.url", "aws.accessKey", "aws.secretKey", "aws.bucket",
-		"broker.host", "broker.port", "broker.user", "broker.password", "broker.vhost", "broker.exchange", "broker.routingKey",
-	}
-)
+const POSIX = "posix"
+const S3 = "s3"
+const SFTP = "sftp"
```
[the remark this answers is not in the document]
jbygdell: In order to not force people to redo the review I'll add this to the next PR.

PR #318, opened 2023-09-20
Add test for Postgres migration scripts

jbygdell:
* Adds a case to the postgres test suite
* Fixes bad migration scripts

### DIFF
--- a/.github/integration/postgres.yml
+++ b/.github/integration/postgres.yml
@@ -26,6 +26,39 @@ services:
     volumes:
       - ./scripts:/scripts
 
+  migration_helper:
+    container_name: helper
+    command:
+      - /usr/bin/tar
+      - -zxf
+      - /pgdata.tgz
+      - -C
+      - /pgdata/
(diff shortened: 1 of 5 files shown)


.github/integration/postgres.yml
```diff
@@ -26,6 +26,39 @@ services:
     volumes:
       - ./scripts:/scripts
 
+  migration_helper:
+    container_name: helper
+    command:
+      - /usr/bin/tar
+      - -zxf
+      - /pgdata.tgz
```
blankdots, line 34: the `pgdata.tgz` here is to fix a specific version of the db and test the migration from there, do i understand it correctly?

I tried to unzip it, but not sure I follow what is there, a small comment somewhere would help.

jbygdell: As per commit message, it is a backup of `sda-db:v0.1.2` i.e. where the migration scripts where introduced.

blankdots (approved):

pontus (approved):

PR #327, opened 2023-09-28
Add support for multiple JWT public keys

### LINKED ISSUE #201: [s3-inbox] Rework handling of public keys
When importing public keys from a JWK endpoint we only reads the first key. https://github.com/neicnordic/sda-s3proxy/blob/a3f4d9b3ec9c906692bb604c5317b7dccfa9c978/userauth.go#L171
This means we have no way of handling key rotations.
Fixing this means we no longer can refer to a key by the name of the issuer of a token.

A/C
* Keys are imported as a list in memory, (can be separated into one list for RSA and one list for EC to minimize any errors)
* Multiple keys from the same issuer can be present, both RSA and EC for example
* Instead of looking for a key by name we try to validate the token against all keys in our list until we find one that matches.
* We deny access only after we have tested all keys present of a given type.

jbygdell:
* Upgrades `lestrrat-go/jwx` to `version 2`
* Deprecates `golang-jwt` in favour for `lestrrat-go/jwx/`
* Deprecates `hostname->pubKey` mapping in favor for `jwt.KeySet`, this allows for multiple keys matching each hostname, in essense gives us support for key rotation of the JWT signer
* JWT validation will now be performed against all configured keys both local path and OIDC server

closes #201


pontus:
Looks mostly good to me, but I'd prefer support for several JWKS URLs. I also don't see JWKS being refreshed anywhere.

jbygdell:
> Looks mostly good to me, but I'd prefer support for several JWKS URLs. I also don't see JWKS being refreshed anywhere.

All in due time, multiple URLs require changes to the config struct and thus should be a separate PR.

```yaml
server:
  jwtpubkeyurl: "http://oidc:8080/jwk"
```

Needs to change to:

```yaml
server:
  jwtpubkeyurl:
    - "http://oidc:8080/jwk"
    - "http://auth/jwk"
```

Auto refresh should actually be the easier of the two to implement.
And since those changes are orthogonal doing them one at the time makes more sense.

pontus (approved):

pontus:
Not sure I see refresh as orthogonal but I'm fine with merging without it as well.

jbygdell:
> Not sure I see refresh as orthogonal but I'm fine with merging without it as well.

What I meant was that refresh is orthogonal to adding multiple URLs.

sda/cmd/s3inbox/s3inbox.md
```diff
@@ -0,0 +1,107 @@
...
+  - `warn` (or `warning`)
+  - `error`
+  - `fatal`
+  - `panic`
+
+## Service Description
+
+The s3inbox proxies uploads to an S3 compatible storage backend.
+
+1. Parses and validates the JWT token (`access_token` in the S3 config file) against teh public keys, either local files on sidk or OIDC JWK endpoints.
```
pontus, line 98:
```suggestion
1. Parses and validates the JWT token (`access_token` in the S3 config file) against the public keys, either locally provisioned or from OIDC JWK endpoints.
```

typo fix, some fuzziness added.

aaperis (approved):
Looks good! I think it's worth adding a jwt rotation test in out integration workflow as a future step :-)

pontus (approved):

blankdots (commented):
I am fine with the changes, but could we split the helper internal into multiple files e.g. for creating folders, handling keys, etc.?
I see having one file, would grow quite a lot the code base - or do we split as we get more helper functions in ?

jbygdell:
> I am fine with the changes, but could we split the helper internal into multiple files e.g. for creating folders, handling keys, etc.? I see having one file, would grow quite a lot the code base - or do we split as we get more helper functions in ?

Since the stuff in the `helper` only are used for the static tests i'll think we can split it if starts to grow any more, or if someone would like to spend time during a refactoring sprint to do that.

blankdots (approved):

PR #344, opened 2023-10-11
Resume multipart uploads

### LINKED ISSUE #255: [s3inbox] Resume multipart uploads
Currently interrupted multipart uploads cannot be resumed.

DoD: Resume an interrupted upload to a ceph s3 backend using `s3cmd`.

MalinAhlberg:
This PR closes #255.

Adds support for:
- resuming an aborted multipart upload (`s3cmd --upload-id Id put FILE s3://BUCKET`)
- listing current multipart uploads (`s3cmd multipart s3://BUCKET [Id]`)
- aborting a multipart upload (`s3cmd abortmp s3://BUCKET/OBJECT Id`)

I added support for `multipart` and `abortmp` although it was not clear from the issue if that's intended.
It appeared to me as these commands should work, when looking at the code, see for example [this snippet](https://github.com/neicnordic/sensitive-data-archive/blob/a7310ee54ce9ea62d6b98a7680bba57de9980721/sda/cmd/s3inbox/proxy.go#L88)
```
case Put, List, Other, AbortMultipart:
    // Allowed
```

If that's not what's intended, the two last commits can be reverted.

I have tested this manually against `https://s3.sto3.safedc.net` and with minio in the docker compose setup of `sda-s3proxy/dev_utils`.


sda/cmd/s3inbox/proxy.go
```diff
@@ -280,7 +280,15 @@ func (p *Proxy) prependBucketToHostPath(r *http.Request) {
 	// Restructure request to query the users folder instead of the general bucket
 	switch r.Method {
 	case http.MethodGet:
-		if strings.Contains(r.URL.String(), "?delimiter") {
+		switch {
+		case strings.Contains(r.URL.String(), "?uploadId"):
+			// resume multipart upload
+			r.URL.Path = "/" + bucket + r.URL.Path
+		case strings.Contains(r.URL.String(), "?uploads"):
```
jbygdell, line 287: This particular case seems to be missing from the `proxy_test.go` file

MalinAhlberg: Added in ec8ae55

MalinAhlberg:
Note: There are no integration tests added. When an upload is killed from a script, it will not print the `upload_id`, which would be needed to test resuming the upload. Minio does not support listing ongoing multipart uploads, which would be the other way of getting the `upload_id`. I could not come up with any way of solving this.

jbygdell (approved):

aaperis (approved):
Looks great! :-)

PR #360, opened 2023-10-17
Add files api

### LINKED ISSUE #436: [api] API call that can give information about the status of each file in the ingestion pipeline
As a submitter
I want to have an API to get information about the state of files in the system
So that I can see that things are happening and remind myself what files I've uploaded.

Given a specific submission or user. The API should give back results for each file where it is in the system. In the inbox, whether ingestion has started, if there are errors and so forth. It's basically reporting the [status](https://github.com/neicnordic/sda-db/blob/schema-update-4/initdb.d/01_main.sql#L112-L123) and metadata logged in the [`file_event_log`](https://github.com/neicnordic/sda-db/blob/schema-update-4/initdb.d/01_main.sql#L127-L143) table.

The authentication for this service could/should be similar to the download or upload services.

This is an API for the users

### LINKED ISSUE #444: [api] Add authentication to API service
As an sda-user
I want to be able to authenticate with the sda-pipeline API
in order to be able to get information about my submission

The authentication should follow the sda-download way, where we use a JWK endpoint for checking the public key and validate the token

There is a discussion in this [PR](https://github.com/neicnordic/sda-pipeline/pull/540) about this issue

MalinAhlberg:
This PR closes https://github.com/neicnordic/sensitive-data-archive/issues/435. It is based on https://github.com/neicnordic/sda-pipeline/pull/540

- merges `api` from repo `sda-pipeline` into `sda`
- adds the endpoint `files` to the API service for retrieving information about the files a user has submitted
- moves `userauth` from `cmd` to `internal`, for reuse in `api`
- updates and refactoring of the `api` code to use the common code 

Testing:
When running the containers from [sda-s3-integration.yml](https://github.com/neicnordic/sensitive-data-archive/blob/7d77fb53005fbde9ce07c445aae174e662614a75/.github/integration/sda-s3-integration.yml), this should work:
```
 # call to the oidc server 
token="$(curl http://localhost:8080/tokens | jq -r '.[0]')" 
 # call to the api
curl -k -L "http://localhost:8090/files" -H "Authorization: Bearer $token" 
```
If the integration tests (`docker compose  -f ~/NBIS/sloth/sensitive-data-archive/.github/integration/sda-s3-integration.yml run integration_test`) are run, the curl response should be
```[{"inboxPath":"requester_demo.org/data/file1.c4gh","fileStatus":"uploaded","createAt":"2023-11-13T10:12:43.144242Z"}]```

Updates of the status of the user's file in the psql db should be reflected in the output of the call, so that only the last event is shown.

fixes #444 
fixes #436

### CODE REVIEW

File: sda/cmd/api/api_test.go [hunk 1]
```diff
@@ -0,0 +1,511 @@
...
+	}
+	Conf.API.MQ, err = broker.NewMQ(Conf.Broker)
+	assert.NoError(t, err)
+
+	Conf.Database = database.DBConf{
+		Host:     "localhost",
+		Port:     dbPort,
+		User:     "lega_in",
+		Password: "lega_in",
+		Database: "lega",
```
jbygdell, line 243:
```suggestion
		User:     "postgres",
		Password: "rootpasswd",
		Database: "sda",
```

sda/cmd/api/api_test.go
```diff
@@ -0,0 +1,511 @@
...
```
jbygdell, line 286:
```suggestion
		User:     "postgres",
		Password: "rootpasswd",
		Database: "sda",
```

sda/cmd/api/api_test.go
```diff
@@ -0,0 +1,511 @@
...
+		}
+	})
+	if err != nil {
+		log.Fatalf("Could not start resource: %s", err)
+	}
+
+	dbHostAndPort := postgres.GetHostPort("5432/tcp")
+	dbPort, _ = strconv.Atoi(postgres.GetPort("5432/tcp"))
+	// TODO this url or postgres://postgres:rootpasswd@%s/sda?sslmode=disable"
+	databaseURL := fmt.Sprintf("postgres://lega_in:lega_in@%s/lega?sslmode=disable", dbHostAndPort)
```
jbygdell, line 83:
```suggestion
	databaseURL := fmt.Sprintf("postgres://postgres:rootpasswd@%s/sda?sslmode=disable", dbHostAndPort)
```

sda/cmd/api/api_test.go
```diff
@@ -0,0 +1,511 @@
...
+
+	Conf = c
+
+	log.Print("Setup DB for my test")
+	Conf.Database = database.DBConf{
+		Host:     "localhost",
+		Port:     dbPort,
+		User:     "lega_in",
+		Password: "lega_in",
+		Database: "lega",
```
jbygdell, line 388:
```suggestion
		User:     "postgres",
		Password: "rootpasswd",
		Database: "sda",
```

sda/cmd/api/api_test.go
```diff
@@ -0,0 +1,511 @@
...
+
+}
+
+func TestDatabasePingCheck(t *testing.T) {
+	database := database.SDAdb{}
+	assert.Error(t, checkDB(&database, 1*time.Second), "nil DB should fail")
+
+	database.DB, _, err = sqlmock.New()
+	assert.NoError(t, err)
+	assert.NoError(t, checkDB(&database, 1*time.Second), "ping should succeed")
```
jbygdell, line 402:
```suggestion
	db, err := database.NewSDAdb(Conf.Database)
	assert.NoError(suite.T(), err)
	assert.NoError(suite.T(), checkDB(db, 1*time.Second), "ping should succeed")
```

You have a real database, no need for any mock instances.

sda/cmd/api/api_test.go
```diff
@@ -0,0 +1,511 @@
...
+			"MQ_VHOST=test",
+			"NOTLS=true",
+		},
+	}, func(config *docker.HostConfig) {
+		// set AutoRemove to true so that stopped container goes away by itself
+		config.AutoRemove = true
+		config.RestartPolicy = docker.RestartPolicy{
+			Name: "no",
+		}
+	})
```
jbygdell, line 116:
```suggestion
	rabbitmq, err := pool.RunWithOptions(&dockertest.RunOptions{
		Repository: "rabbitmq",
		Tag:        "3-management-alpine",
	}, func(config *docker.HostConfig) {
		// set AutoRemove to true so that stopped container goes away by itself
		config.AutoRemove = true
		config.RestartPolicy = docker.RestartPolicy{
			Name: "no",
		}
	})
```

Even though we are not interacting with the MQ for now we can at least make it future proof.

jbygdell (changes requested):

MalinAhlberg:
Thanks for the very detailed review, @jbygdell :+1: !  I fixed almost all of the comments, except for the one in [`config.go`](https://github.com/neicnordic/sensitive-data-archive/pull/360#discussion_r1391067653), since the pattern 
```
err = ....
if err != nil {
```
is used for all other cases in that section.

jbygdell:
A short readme file needs to be created

```
# Api

short one line description

## Service Description

More in depth description
```

jbygdell:
> > A short Readme.md file needs to be created
> 
> In `sda/cmd/api`?

Yes, name it `api.md` so that it matches the other readme files

MalinAhlberg:
> A short Readme.md file needs to be created

See be12ee5

jbygdell (approved):

sda/cmd/api/api.md
```diff
@@ -0,0 +1,19 @@
+# API
+The API service provides data submitters with functionality to controll
```
dbampalikis, line 2:
```suggestion
The API service provides data submitters with functionality to control
```

sda/cmd/api/api.md
```diff
@@ -0,0 +1,19 @@
...
+their submissions. Users are authenticated with a JWT. 
+
+## Service Description
+
+Endpoints:
+  - `/files`
+  
+    1. Parses and validates the JWT token against the public keys, either locally provisioned or from OIDC JWK endpoints.
+    2. The `sub` field from the token is extracted ad used as the user's identifier
```
dbampalikis, line 11:
```suggestion
    2. The `sub` field from the token is extracted and used as the user's identifier
```

dbampalikis (approved):
~Really nice! Tested it manually (without using the integration test) and it works great! Also tested with another user and the result is an empty json array, which feels correct. Not sure if we would like to add a message such as `no files found` or something, but that's just an extra thing I guess~ Updated the review after Alex's finding

aaperis:
I made a test with two files. I uploaded the first one:
```
./sda-cli upload -config s3cfg --encrypt-with-key c4gh.pub.pem test_file_demo

curl -sk -L "http://localhost:8090/files" -H "Authorization: Bearer $token" | jq
```
and the `files api` responded nicely:
```json
[
  {
    "inboxPath": "test_dummy.org/test_file_demo.c4gh",
    "fileStatus": "uploaded",
    "createAt": "2023-11-17T14:07:48.825563Z"
  }
]
```

Then, I uploaded another file:
```
./sda-cli upload -config s3cfg test_file_demo.c4gh -targetDir asd/testtest.c4gh
```
```
./sda-cli list -config s3cfg 
The provided token expires in less than 24 hours
Consider renewing the token.
1.00MB   asd/testtest.c4gh/test_file_demo.c4gh 
1.00MB   test_file_demo.c4gh 
```

Then
```
curl -sk -L "http://localhost:8090/files" -H "Authorization: Bearer $token" | jq
```
but the `files api` returns only info on the last uploaded file:
```json
[
  {
    "inboxPath": "test_dummy.org/asd/testtest.c4gh/test_file_demo.c4gh",
    "fileStatus": "uploaded",
    "createAt": "2023-11-17T14:11:47.143238Z"
  }
]
```
We sat down with @dbampalikis and tested a correction to the `db` query, see suggestion in Dimi's comments :-)

dbampalikis (changes requested):
Really nice, apart for the comment about the database.
Also tested with another user and the result is an empty json array, which feels correct. Not sure if we would like to add a message such as no files found or something, but that's just an extra thing.
Finally, maybe we should have tests for multiple files?

MalinAhlberg:
Thanks @aaperis and @dbampalikis , really good that you spotted this :see_no_evil:  :pray: ! I commited your fix along with a test case for this, see dd0e91f.

MalinAhlberg:
(force pushed to rebase on main)

dbampalikis (approved):
LGTM! I didn't test again but everything looks good

dbampalikis (approved):

sda/cmd/api/api.go
```diff
@@ -0,0 +1,183 @@
...
+	signal.Notify(sigc, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
+	go func() {
+		<-sigc
+		shutdown()
+		os.Exit(0)
+	}()
+
+	srv := setup(Conf)
+	if Conf.API.ServerCert != "" && Conf.API.ServerKey != "" {
+		log.Infof("Web server is ready to receive connections at https://%s:%d", Conf.API.Host, Conf.API.Port)
```
pontus, line 54:
```suggestion
		log.Infof("Starting web server at https://%s:%d", Conf.API.Host, Conf.API.Port)
```

Less confusing if something fails at start, does not give impression that server was running, then failed.

sda/cmd/api/api.go
```diff
@@ -0,0 +1,183 @@
...
+		if err := srv.ListenAndServeTLS(Conf.API.ServerCert, Conf.API.ServerKey); err != nil {
+			shutdown()
+			log.Fatalln(err)
+		}
+	} else {
+		log.Infof("Web server is ready to receive connections at http://%s:%d", Conf.API.Host, Conf.API.Port)
```
pontus, line 60:
```suggestion
		log.Infof("Starting web server at http://%s:%d", Conf.API.Host, Conf.API.Port)
```

Less confusing if something fails at start, does not give impression that server was running, then failed.

sda/internal/database/database.go
```diff
@@ -48,6 +48,12 @@ type SyncData struct {
 	Checksum string
 }
 
+type SubmissionFileInfo struct {
+	InboxPath string `json:"inboxPath"`
+	Status    string `json:"fileStatus"`
+	CreateAt  string `json:"createAt"`
```
pontus, line 54: I suspect it might also be helpful to provide the uploaded size (size the user saw before upload), and I don't see any problems with that (e.g. information leakage is not a thing since you only see your files).

sda/internal/database/db_functions.go
```diff
@@ -622,3 +622,61 @@ func (dbs *SDAdb) getArchivePath(stableID string) (string, error) {
...
+			break
+		}
+		time.Sleep(time.Duration(math.Pow(2, float64(count))) * time.Second)
+	}
+
+	return files, err
+}
+
+// getUserFiles is the actual function performing work for GetUserFiles
+func (dbs *SDAdb) getUserFiles(userID string) ([]*SubmissionFileInfo, error) {
```
pontus, line 645: Is it over-complicating things to think about filtering or paging?

I suspect the primary users of the API for which this is provided are the ones who would upload hundreds of thousands or millions of files and could benefit from not having to grab all and filter locally.

sda/cmd/api/api.go
```diff
@@ -0,0 +1,183 @@
...
+
+func main() {
+	Conf, err = config.NewConfig("api")
+	if err != nil {
+		log.Fatal(err)
+	}
+	Conf.API.MQ, err = broker.NewMQ(Conf.Broker)
+	if err != nil {
+		log.Fatal(err)
+	}
```
pontus, line 34: Is MQ being used in this microservice or is there any usage foreseen?

sda/cmd/api/api_test.go
```diff
@@ -0,0 +1,509 @@
...
+			return err
+		}
+		res.Body.Close()
+
+		return nil
+	}); err != nil {
+		if err := pool.Purge(oidc); err != nil {
+			log.Panicf("Could not purge oidc resource: %s", err)
+		}
+		log.Panicf("Could not connect to oidc: %s", err)
```
jbygdell, line 192:
```suggestion
		if err := pool.Purge(oidc); err != nil {
			log.Panicf("Could not purge oidc resource: %s", err)
		}
		if err := pool.Purge(postgres); err != nil {
			log.Fatalf("Could not purge resource: %s", err)
		}
		if err := pool.Purge(rabbitmq); err != nil {
			log.Fatalf("Could not purge resource: %s", err)
		}
		log.Panicf("Could not connect to oidc: %s", err)
```
The previously started containers needs to be removed as well.

sda/cmd/api/api_test.go
```diff
@@ -0,0 +1,509 @@
...
+		Password: "rootpasswd",
+		Database: "sda",
+		SslMode:  "disable",
+	}
+	Conf.API.DB, err = database.NewSDAdb(Conf.Database)
+	assert.NoError(suite.T(), err)
+
+}
+
+func TestDatabasePingCheck(t *testing.T) {
```
jbygdell, line 377:
```suggestion
func (suite *TestSuite) TestDatabasePingCheck() {
```

sda/cmd/api/api_test.go
```diff
@@ -0,0 +1,509 @@
...
+	}
+	if err := pool.Purge(rabbitmq); err != nil {
+		log.Fatalf("Could not purge resource: %s", err)
+	}
+	if err := pool.Purge(oidc); err != nil {
+		log.Fatalf("Could not purge resource: %s", err)
+	}
+}
+
+func TestShutdown(t *testing.T) {
```
jbygdell, line 210:
```suggestion
func (suite *TestSuite) TestShutdown() {
```

jbygdell: All tests should be part of the suite

sda/cmd/api/api_test.go
```diff
@@ -0,0 +1,509 @@
...
+
+	// make sure all conections are alive
+	assert.Equal(t, false, Conf.API.MQ.Channel.IsClosed())
+	assert.Equal(t, false, Conf.API.MQ.Connection.IsClosed())
+	assert.Equal(t, nil, Conf.API.DB.DB.Ping())
+
+	shutdown()
+	assert.Equal(t, true, Conf.API.MQ.Channel.IsClosed())
+	assert.Equal(t, true, Conf.API.MQ.Connection.IsClosed())
+	assert.Equal(t, "sql: database is closed", Conf.API.DB.DB.Ping().Error())
```
jbygdell, line 241: Replace `t,` with `suite.T(),` in all asserts

sda/cmd/api/api_test.go
```diff
@@ -0,0 +1,509 @@
...
+	Conf.API.DB.Close()
+	res, err = http.Get(ts.URL + "/ready")
+	assert.NoError(t, err)
+	assert.Equal(t, http.StatusServiceUnavailable, res.StatusCode)
+	defer res.Body.Close()
+
+	// reconnect should be fast so now this should pass
+	res, err = http.Get(ts.URL + "/ready")
+	assert.NoError(t, err)
+	assert.Equal(t, http.StatusOK, res.StatusCode)
```
jbygdell, line 314: Replace t, with suite.T(), in all asserts

sda/cmd/api/api_test.go
```diff
@@ -0,0 +1,509 @@
...
+
+}
+
+func TestDatabasePingCheck(t *testing.T) {
+	emptyDB := database.SDAdb{}
+	assert.Error(t, checkDB(&emptyDB, 1*time.Second), "nil DB should fail")
+
+	db, err := database.NewSDAdb(Conf.Database)
+	assert.NoError(t, err)
+	assert.NoError(t, checkDB(db, 1*time.Second), "ping should succeed")
```
jbygdell, line 383: Replace t, with suite.T(), in all asserts

sda/cmd/api/api_test.go
```diff
@@ -0,0 +1,509 @@
...
+	assert.Equal(t, false, Conf.API.MQ.Connection.IsClosed())
+	assert.Equal(t, nil, Conf.API.DB.DB.Ping())
+
+	shutdown()
+	assert.Equal(t, true, Conf.API.MQ.Channel.IsClosed())
+	assert.Equal(t, true, Conf.API.MQ.Connection.IsClosed())
+	assert.Equal(t, "sql: database is closed", Conf.API.DB.DB.Ping().Error())
+}
+
+func TestReadinessResponse(t *testing.T) {
```
jbygdell, line 244:
```suggestion
func (suite *TestSuite) TestReadinessResponse() {
```

sda/internal/database/db_functions.go
```diff
@@ -622,3 +622,61 @@ func (dbs *SDAdb) getArchivePath(stableID string) (string, error) {
...
+	for count := 1; count <= RetryTimes; count++ {
+		files, err = dbs.getUserFiles(userID)
+		if err == nil {
+			break
+		}
+		time.Sleep(time.Duration(math.Pow(2, float64(count))) * time.Second)
+	}
+
+	return files, err
+}
```
jbygdell, line 642: This should have a test case in `db_functions_test.go`.
Preferably testing multiple files with more than 1 status event for each file.

jbygdell (changes requested):

sda/cmd/api/api.go
jbygdell: Future proofing, the vision is for an admin to be able to trigger ingestion / map IDs with this API.

sda/internal/database/db_functions.go
jbygdell: Down the road we at least need to limit this to files that doesn't have stable IDs yet.

As for filtering we could support POST request to the files endpoint with a path filter in the payload, but that is outside the scope of this PR.

```json
{"filter": "/subpath/target"}
```

sda/internal/database/db_functions.go
MalinAhlberg: see 8caa156

aaperis:
I tested this and it works great for multiple files each with different statuses. The size of the uploaded file will always be empty though since this is never entered in the database currently. In my opinion, this part could be removed for now. We could modify `s3inbox` to update the `submission_file_size` upon successful upload if we ever want this in the future.

However, we can't rely too heavily on info provided by `s3inbox`  because this is not used by all nodes. On the other hand, returning info that is collected in the database when `ingest` finishes is not that useful for the user imo so thus we compromise for now.

MalinAhlberg:
> I tested this and it works great for multiple files each with different statuses. The size of the uploaded file will always be empty though since this is never entered in the database currently. In my opinion, this part could be removed for now. We could modify `s3inbox` to update the `submission_file_size` upon successful upload if we ever want this in the future.
> 
> However, we can't rely too heavily on info provided by `s3inbox` because this is not used by all nodes. On the other hand, returning info that is collected in the database when `ingest` finishes is not that useful for the user imo so thus we compromise for now.

Great that you looked into this, @aaperis :pray: ! I agree, I think we should revert commit ca3185e and leave that for later. If we want to be sure to remeber it, we can create a new issue about it. Is that ok with @pontus ?

pontus:
Yes, I'm fine with delaying file sizes (but will create an issue for that as I think it's very desirable).

jbygdell (approved):

pontus:
Looks good.

I have some thoughts about the TLS-configuration but I think I'll make an issue for general overhaul at some point instead.

pontus (approved):

PR #362, opened 2023-10-18
Correlation IDs are not sent for Dataset events.

jbygdell:
```
Currently, the messages that contain a correlation id are those that need to be correlated:
* inbox
* ingest
* verify
* accession
* completed

The mapping message is totally isolated from this flow,  
it is triggered when HD accepts the submission, and there is no further message/action correlated to it.
```

This PR affects `mapper` and ratifies the assumption we had during the spring that all messages would have correlation IDs

Only the first commit is of any importance, the rest is just to get the old test suite to pass for these changes.


blankdots:
i think we also need to mark the files ready, although not 100%, it seem that most of the integration tests fail because of that

blankdots (approved):

aaperis (approved):

PR #383, opened 2023-10-27
[s3inbox] set correlation ID when updating file status

### LINKED ISSUE #381: S3proxy should add correlation id to the database.
Since the s3proxy sends the "Uploaded" message it has the correlation id for that file, it should be in the file_event_log already here.

jbygdell:
Set the correlation ID when setting the file status to `uploaded`

closes #381


kusalananda (approved):

kjellp (commented):
Is it safe to use the same UUID for FileUUID and CorrelationID here?

jbygdell:
> Is it safe to use the same UUID for FileUUID and CorrelationID here?

Yes, since the FileUUID is used when sending the message, (Line 179 above).

kjellp (approved):
Ok!

PR #398, opened 2023-11-06
Update charts

jbygdell:
- Updates the charts to match the latest changes in the `sda` code path.
- Sets the app version in the chart to be used as image tag
- Updates the `app-version` when we push a new tag and also bump the chart version before we do a new release
- Adds debug steps on failure in `Build PR container` action.


charts/sda-svc/Chart.yaml
```diff
@@ -1,6 +1,7 @@
 apiVersion: v2
 name: sda-svc
-version: "0.20.7"
+version: "0.21.0"
+appVersion: "v0.0.163"
 kubeVersion: ">= 1.19.0-0"
```
blankdots, line 5: should this be bumped in all charts ? to `1.25`

blankdots (commented):
should the tags be `""` empty string ?

charts/sda-svc/Chart.yaml
jbygdell: Soon, we need to remove everything that has to do with PSP first, but that will be a different PR since it will have implications for the current deployments.

blankdots (approved):

charts/sda-svc/templates/s3-inbox-secrets.yaml
```diff
@@ -7,11 +7,10 @@ metadata:
   name: {{ template "sda.fullname" . }}-inbox
 type: Opaque
 data:
-  dbPassword: {{ include "dbPassInbox" . | b64enc }}
-  dbUser: {{ include "dbUserInbox" . | b64enc }}
-  mqPassword: {{ include "mqPassInbox" . | b64enc }}
-  s3InboxAccessKey: {{ .Values.global.inbox.s3AccessKey | b64enc }}
-  s3InboxSecretKey: {{ .Values.global.inbox.s3SecretKey | b64enc }}
+  dbPassword: {{ required "DB password is required" (include "dbPassInbox" .) | quote | trimall "\"" | b64enc }}
```
dbampalikis, line 10: Maybe it's obvious, but why do we require `quote | trimall "\""` in this case?

dbampalikis (approved):
As far as I can tell, it looks fine. However, I didn't test it, so I am approving to move forward with the BP deployment of the sync+sync-api

charts/sda-svc/templates/s3-inbox-secrets.yaml
jbygdell: The DB password can contain characters  that causes the string to quoted, and we don't want to keep the quotation characters when we quote it before base64 encoding it. At least that's from the old helm versions, seems like a perfect thing to create a test for.

PR #399, opened 2023-11-06
add update-docs workflow

aaperis:
This is needed for migrating the automatic documentation update from `sda-pipeline` to this repository.

### DIFF
--- a/.github/workflows/update-docs.yml
+++ b/.github/workflows/update-docs.yml
@@ -22,4 +22,7 @@ jobs:
           -H "Accept: application/vnd.github.v3+json" \
           https://api.github.com/repos/neicnordic/neic-sda/actions/workflows/aggregate.yml/dispatches \
           -d "{\"ref\": \"master\", \"inputs\":{\"repository\": \"$GITHUB_REPOSITORY\"}}")
-          [ -z "$result" ] && true || (echo "$result" && false)
+          if [ -n "$result" ]; then
+              echo "$result"
+              false
+          fi
--- a/.github/workflows/update-docs.yml
+++ b/.github/workflows/update-docs.yml
@@ -1,0 +1,25 @@
+name: Update documentation
+
+# NOTE: This action requires that a github secret exists in this repository
+#       named NEIC_SDA_ACCESS, containing a github token with the
+#       "Actions: Read and Write" permission on the neicnordic/neic-sda
+#       repository.
+
+on:
+  push:
+    branches:
+    - main
+
+jobs:
+
+  build:
+    name: Trigger documentation rebuild in neic-sda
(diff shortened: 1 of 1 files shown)


blankdots (approved):

pontus (dismissed):

dbampalikis (dismissed):

.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,25 @@
...
+  build:
+    name: Trigger documentation rebuild in neic-sda
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+          result=$(curl -X POST \
+          -H "Authorization: Bearer ${{secrets.NEIC_SDA_ACCESS}}" \
+          -H "Accept: application/vnd.github.v3+json" \
+          https://api.github.com/repos/neicnordic/neic-sda/actions/workflows/aggregate.yml/dispatches \
+          -d "{\"ref\": \"master\", \"inputs\":{\"repository\": \"$GITHUB_REPOSITORY\"}}")
```
kusalananda, line 24: Indent for readability, put URL last, and sort out quoting of the JSON payload.

```suggestion
          result=$(
              curl -X POST \
                 -H "Authorization: Bearer ${{secrets.NEIC_SDA_ACCESS}}" \
                 -H "Accept: application/vnd.github.v3+json" \
                 -d '{"ref": "master", "inputs": {"repository": "'"$GITHUB_REPOSITORY"'"}}' \
                 https://api.github.com/repos/neicnordic/neic-sda/actions/workflows/aggregate.yml/dispatches
          )

```

.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,25 @@
...
+          [ -z "$result" ] && true || (echo "$result" && false)
```
kusalananda, line 25: `&& true` is a no-op. Write it with a `if` so it's clear what we want to happen.

```suggestion
          if [ -n "$result" ]; then
              echo "$result"
              false
          fi
```

kusalananda (commented):
Nitpicking, for readability. Ignore it if it's not worth the effort.

.github/workflows/update-docs.yml
aaperis: To be frank, I don't find `'"'"$a"'"'` more readable as compared to `"\"$a\""` so I would prefer to keep the backslash version :sweat_smile:

.github/workflows/update-docs.yml
aaperis: Looks more reasonable :-)

kusalananda (approved):

pontus (approved):

blankdots (approved):

PR #448, opened 2023-11-16
[charts] Added missing values in integration test values.yaml

dbampalikis:
The PR adds a couple of missing values in the `values.yaml` file of the integration scripts, so that the `helm template` command doesn't fail when testing locally.


dbampalikis:
Already added in the `update_charts` PR

PR #507, opened 2023-12-11
Add group entitlement to sda-auth

dbampalikis:
This pull requests adds the `eduperson_entitlement` to the scopes requested by sda-auth during login with LS-AAI.

### DIFF
--- a/sda-auth/elixir.go
+++ b/sda-auth/elixir.go
@@ -19,7 +19,7 @@ type ElixirIdentity struct {
 	Token                string
 	Profile              string
 	Email                string
-	EdupersonEntitlement string
+	EdupersonEntitlement []string
 	ExpDate              string
 }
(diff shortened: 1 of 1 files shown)


jbygdell (approved):
A test case for this would be nice, but maybe the changes are so large that it is better to do them in a separate PR. For example by using the mock LS-AAI from GDI.

pontus (approved):

dbampalikis:
> A test case for this would be nice, but maybe the changes are so large that it is better to do them in a separate PR. For example by using the mock LS-AAI from GDI.

I tried to work on this with the mock LS-AAI provided in GDI but didn't manage to add the scope. I agree, we can look at it in a new PR and fix the test suite in general, since it doesn't seem to take care of many cases

PR #535, opened 2023-12-15
Store file errors in db

### LINKED ISSUE #349: [sda] Store file errors in the database
As a sda-admin

I want to store the file errors message in the database

The record should be stored in the schema `file_event_log`, item `error` as text string

**Note** This issue needs to be refined. e.g. How should the error be stored
Schema can be found at https://github.com/neicnordic/sensitive-data-archive/blob/main/postgresql/initdb.d/01_main.sql#L111-L144

1. Add option to put data in the error field in the go function that updates the status of files.
2. Go through the sda-pipeline and add a call to this function everywhere something goes wrong with files.

Store:
- information that should be relayed back to user. File errors
- 

Don't store:
- Database connection errors
-

jbygdell:
This PR closes #349.

**Description**
This PR adds functionality for `ingest` and `verify` to store file related errors in the database.
Currently these errors are considered fatal:
* file not found
* unable to decrypt the file (ingest)

**How to test**
```cmd
make integrationtest-sda
```


blankdots (approved):

.github/integration/tests/sda/22_error_test.sh
```diff
@@ -0,0 +1,133 @@
...
+        --arg payload_encoding base64 \
+        --arg payload "$verify_payload" \
+        '$ARGS.named'
+)
+
+curl -k -u guest:guest "$URI/api/exchanges/sda/sda/publish" \
+    -H 'Content-Type: application/json;charset=UTF-8' \
+    -d "$verify_body"
+
+# check database to verify file status
```
pontus, line 120:
```suggestion
# check database to verify file status
RETRY_TIMES=0
```

Without resetting the counter, timeout won't work.

sda/cmd/ingest/ingest.go
```diff
@@ -257,8 +258,14 @@ func main() {
 				}
 
 				file, err := inbox.NewFileReader(message.FilePath)
-				if err != nil {
+				if err != nil { //nolint:nestif
 					log.Errorf("Failed to open file to ingest reason: (%s)", err.Error())
+					if strings.Contains(err.Error(), "no such file or directory") || strings.Contains(err.Error(), "NoSuchKey:") {
```
pontus, line 263: I'm not sure I understand the thinking why we test for specific errors here.

To me, we're in a branch that will send to the `error` queue for informing/allowing manual interaction. I believe those errors all make sense to add to the file event log - the exception would if there's confidentiality reasons, but I don't see any such.

jbygdell: S3 and NFS can have newtork hickups and in those cases we don't want to stop the ingestion and report an error, we just try again a bit later.

pontus: That makes sense, but shouldn't that be the same for e.g. whatever to send to the `error` queue?

jbygdell: There needs to be some refactoring later on, because if we `nack` a message we should not send it to the error queue.

I didn't want to include those changes here in order to keep the PR contained in scope.

.github/integration/tests/sda/22_error_test.sh
```diff
@@ -0,0 +1,134 @@
...
+    RETRY_TIMES=$((RETRY_TIMES + 1))
+    if [ "$RETRY_TIMES" -eq 30 ]; then
+        echo "::error::Time out while waiting for verify to set file error"
+        exit 1
+    fi
+    sleep 2
+done
+
+echo ""
+echo "file error test completed successfully"
```
pahatz, line 134: Is this for the newline? Maybe:
```suggestion
echo -e "\nfile error test completed successfully"
```

aaperis (approved):

blankdots (approved):

PR #590, opened 2024-01-11
Update charts to comply with K8s v1.26

### LINKED ISSUE #461: [Charts] update all charts to be > v1.25 compatible
* Remove pod security policies
* Check if the TLS permission init containers can be removed. (most likely)
* Check if we comply with the [restricted](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted) pod security standard

jbygdell:
This PR closes #461.

**Description**
This PR deprecates the now obsolete `pod security policy` and includes changes that makes the charts compliant with the `restricted` policy of the new `pod security standards`

**How to test**


charts/sda-svc/templates/auth-deploy.yaml
```diff
@@ -146,7 +151,7 @@ spec:
           value: {{ template "tlsPath" . }}/tls.key
         {{- end }}
         - name: S3INBOX
-          value: {{ .Values.global.ingress.hostName.s3Inbox }}
+          value: {{ .Values.global.ingress.hostName.inbox }}
```
pontus, line 154: Values in this PR still has `s3Inbox`? Documentation update?

charts/sda-svc/templates/syncapi-deploy.yaml
```diff
@@ -164,15 +169,15 @@ spec:
...
-                name: {{ required "An certificate issuer or a TLS secret name is required for syncapi" .Values.syncapi.tls.secretName }}
-          {{- end }}
+        {{- if or .Values.global.tls.clusterIssuer .Values.global.tls.issuer }}
+          secret:
+            defaultMode: 0440
+            secretName: {{ template "sda.fullname" . }}-syncapi-certs
+        {{- else if .Values.syncapi.tls.secretName }}
+          secret:
+            defaultMode: 0440
+            secretName: {{ required "An certificate issuer or a TLS secret name is required for syncapi" .Values.syncapi.tls.secretName }}
```
pontus, line 179: Since we're checking for `.Values.syncapi.tls.secretName` in the helm conditional, the `required doesn't do anything here?

charts/sda-svc/templates/verify-deploy.yaml
```diff
@@ -210,21 +198,15 @@ spec:
...
-            medium: Memory
-            sizeLimit: 10Mi
+        {{- if or .Values.global.tls.clusterIssuer .Values.global.tls.issuer }}
+          secret:
+            defaultMode: 0440
+            secretName: {{ template "sda.fullname" . }}-verify-certs
+        {{- else if .Values.verify.tls.secretName }}
+          secret:
+            defaultMode: 0440
+            secretName: {{ required "An certificate issuer or a TLS secret name is required for verify" .Values.verify.tls.secretName }}
```
pontus, line 209: Otiose `required`

pontus (dismissed):
Looks good (visual inspection, hasn't tried installing).

charts/sda-svc/templates/syncapi-deploy.yaml
jbygdell: Rather the opposite since the `if else` isn't a bool check it gets ignored

pontus: I don't think I understand what you mean here - the helm flow control introduction (https://helm.sh/docs/chart_template_guide/control_structures/#ifelse) specifically lists the sets of values for the pipeline that are considered as false. Of course, `required` isn't the same but I can't think of any cases where `required` will trigger an error here.

But this should be harmless and not a hindrance for merge.

jbygdell: `if .Values.syncapi.tls.secretName` will only trigger if `if .Values.syncapi.tls.secretName` is unset or bool `false`, not if it is an empty string. (which is the default in the values file)

pontus: That sounds odd, is this some non-standard behaviour activated somewhere?

This shows what I expect, which I think aligns with the helm documentation:

```
pontus@lalage:/tmp$ cat testdir/templates/testfil 
# Testing a few evaluations

{{ if "" }}
# Empty string is true
{{ end }}

{{ if 0 }}
# 0 is true
{{ end }}

{{ if .Values.DoesNotExist }}
# {{ .Values.DoesNotExist }} from unset is true
{{ end }}

{{ if false }}
# false is true
{{ end }}

pontus@lalage:/tmp$ helm template testdir
---
# Source: testchart/templates/testfil
# Testing a few evaluations
pontus@lalage:/tmp$ 
```

blankdots (dismissed):

pontus (approved):

pontus (approved):

blankdots (approved):

PR #604, opened 2024-01-18
md5 not required in schemas

blankdots:
This PR closes just a part of the MD5 validation for new, new versions of the EGA JSON schemas will render MD5 requirement as null.

**Description**

MD5 is not required in new versions of the schemas (and even now), so might as well remove it from the required schema validation

**How to test**
Same tests as before

### DIFF
--- a/sda/internal/schema/schema_test.go
+++ b/sda/internal/schema/schema_test.go
@@ -193,7 +193,7 @@ func TestValidateJSONIngestionAccessionRequest(t *testing.T) {
 		User:     "JohnDoe",
 		FilePath: "path/to file",
 		DecryptedChecksums: []Checksums{
-			{Type: "sha256", Value: "da886a89637d125ef9f15f6d676357f3a9e5e10306929f0bad246375af89c2e2"},
+			{Type: "md5", Value: "68b329da9893e34099c7d8ad5cb9c940"},
 		},
 	}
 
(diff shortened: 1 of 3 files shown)


sda/internal/schema/schema_test.go
```diff
@@ -193,7 +193,7 @@ func TestValidateJSONIngestionAccessionRequest(t *testing.T) {
 		User:     "JohnDoe",
 		FilePath: "path/to file",
 		DecryptedChecksums: []Checksums{
-			{Type: "sha256", Value: "da886a89637d125ef9f15f6d676357f3a9e5e10306929f0bad246375af89c2e2"},
+			{Type: "sha256", Value: "da886a89637d125ef9f15f6d676357f3a9e5e10306929f0bad246375afc2e2"},
```
jbygdell, line 196: Why changing only 2 out of 11?

sda/schemas/isolated/ingestion-completion.json
```diff
@@ -130,17 +130,6 @@
                     }
                 ]
             ],
-            "contains": {
-                "type": "object",
-                "properties": {
-                    "type": {
-                        "const": "sha256"
```
jbygdell, line 137: So all checksums are optional from now on? since this one is `sha256`

blankdots: that is a mistake

sda/internal/schema/schema_test.go
blankdots: that test expects to fail as it misses md5, at least that is how i read it, personally i would remove it.
so the other direction i took is to fail the sha256 value

blankdots: i changed that to be md5 instead of sha256 but i am fine with any other suggestion

dbampalikis (approved):
Looks good to me, even though I didn't test it

jbygdell (approved):

PR #619, opened 2024-01-24
[Hotfix] race condition related to `syncapi`

### LINKED ISSUE #618: Fix sync to avoid race conditions

jbygdell:
This PR closes #618 .

**Description**
This should fix a race condition seen when the `syncapi` sends the messages for a dataset ingestion all at once.

**How to test**

### DIFF
--- a/sda/cmd/finalize/finalize.go
+++ b/sda/cmd/finalize/finalize.go
@@ -96,9 +96,21 @@ func main() {
 
 				continue
 			}
-			if status == "disabled" {
+
+			switch status {
+			case "disabled":
 				log.Infof("file with correlation ID: %s is disabled, stopping work", delivered.CorrelationId)
 				if err := delivered.Ack(false); err != nil {
+					log.Errorf("Failed acking canceled work, reason: %v", err)
+				}
+
+				continue
+
+			case "verified":
+			case "enabled":
+			default:
+				log.Warnf("file with correlation ID: %s is not verified yet, stopping work", delivered.CorrelationId)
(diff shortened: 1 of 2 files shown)


sda/cmd/finalize/finalize.go
```diff
@@ -96,12 +96,24 @@ func main() {
 
 				continue
 			}
-			if status == "disabled" {
+
```
pontus, line 99: Should we really have special handling for `disabled`? They could still be reenabled by a message that hasn't been processed yet?

I do see why we wouldn't want to process the same message over and over, but are we sure that's not a use case we need to handle?

pontus:
I'm not sure if I understand the test correctly - to me it seems it wouldn't fail for the out-of-order case we're trying to fix either?

sda/cmd/finalize/finalize.go
jbygdell: The `disabled` case is due to the async handling in FEGA where we could get a cancellation message for a file that was currently being ingested.

pontus: I believe I understand the context but see a (possible although unlikely) similar race condition between whatever message/service is responsible for changing the state `disabled` -> `enabled` and the message for finalize.

But I also see it as low risk, so probably fine to (consciously) ignore.

pontus (approved):

dbampalikis (approved):
I didn't test it, but it looks good to me! I am not merging in case anyone else wants to check it

aaperis (approved):
Looks reasonable :-)

PR #622, opened 2024-01-24
Rabbitmq consumer timeout

### LINKED ISSUE #621: Timeout on message handling
**Describe the bug**
The total time for completing the work on a message is only 30min, that is not enough when ingestiong files larger than 100GB

**Steps to reproduce**
Upload and start ingestion of a file >= 100GB

**Expected behavior**
File should finish ingestion and not restart due to timeout.

**Additional context**

**Estimation of size**: small

jbygdell:
This PR closes #621 

**Description**
Rabbitmq can set the time allowed before a message is supposed to be either acked or nacked on a global level. This time is the same regardless if the consumer retrieves 1 or 100 messages as it is counted for each retrieval event.

This PR does the following:
* sets the default number of messages delivered to each worker to 1, to ensure that each message gets the allotted time
* adds a config option to the RabbitMQ for setting the global consumer timeout, the default value is raised from 30 min to 4 h
* adds the config option for the consumer timeout to the helm chart

**How to test**

### DIFF
--- a/.github/integration/tests/rabbitmq/20_configuration_test.sh
+++ b/.github/integration/tests/rabbitmq/20_configuration_test.sh
@@ -1,0 +1,15 @@
+#!/bin/sh
+set -ex
+
+MQHOST="localhost:15672"
+if [ -f /.dockerenv ]; then
+	MQHOST="mq:15671"
+fi
+
+timeout=$(curl -s --cacert /tmp/certs/ca.crt -u guest:guest "https://$MQHOST/api/queues/sda/from_cega" | jq '."consumer_details"[]."consumer_timeout"')
+if [ "$timeout" -ne 1000 ]; then
+    echo "active timeout is wrong, expected: 1000, actual: $timeout"
+    exit 1
+fi
+
+echo "configuration test completed successfully"
--- a/.github/integration/rabbitmq-federation.yml
+++ b/.github/integration/rabbitmq-federation.yml
@@ -39,8 +39,9 @@ services:
 
   federation_test:
     command:
-      - /bin/sh
-      - /tests/rabbitmq/10_federation_test.sh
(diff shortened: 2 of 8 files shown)


.github/integration/tests/rabbitmq/20_configuration_test.sh
```diff
@@ -0,0 +1,15 @@
...
+
+MQHOST="localhost:15672"
+if [ -f /.dockerenv ]; then
+	MQHOST="mq:15671"
+fi
+
+timeout=$(curl -s --cacert /tmp/certs/ca.crt -u guest:guest "https://$MQHOST/api/queues/sda/from_cega" | jq '."consumer_details"[]."consumer_timeout"')
+if [ "$timeout" -ne 1000 ]; then
+    echo "active timeout is wrong, expected: 1000, actual: $timeout"
+    exit 1
```
aaperis, line 12: Shouldn't this number be `14400000` instead of `1000`? What am I missing?

jbygdell: If you  look in `rabbitmq-federation.yml` you will see that for the tests we set the timeout to 1s , just to test that the modified value actually gets enabled.

aaperis: Oh, I missed line102 there!

aaperis (dismissed):
Looks good :-)

pontus (dismissed):

blankdots (dismissed):
I am fine with this, and i think we need to find a way to include this information into our neic-sda readthedocs.

blankdots (approved):

blankdots (approved):

pontus (approved):

PR #670, opened 2024-02-14
Make linter happier by not naming unused parameters

pontus:
This hopefully fixes linter warnings in sda-downloads for unused variables being named.


jbygdell (approved):
Yeah, this is because all of these tests are stubs.

Issue #400, opened 2023-11-06
[sda-auth] Add htsget to s3cmd


dbampalikis:
As an sda-user
I want to have the htsget endpoint in the configuration file after logging in
In order to be able to download files from the htsget endpoint

PR #695, opened 2024-02-27
htsget download endpoints

MalinAhlberg:
This PR closes NBISweden/LocalEGA-SE-Deployment#725 and neicnordic/sda-download#372.

It is based on the work in https://github.com/neicnordic/sda-download/pull/327.

**Description**
- Adds endpoint `/s3-encrypted` which is similar to `/s3` but returns an encrypted file.
- Adds support for returning parts of encrypted files, either with url param `endCoordinate` or - for encrypted files - with the http header `Range: bytes=0-1000`. The later overrides the first, and is strictly interpreted, while `endCoordinate` will be used to find the nearest crypt4gh data block boundary and cut the file there.
- Minor modifications to work with htsget-rs (see https://github.com/GenomicDataInfrastructure/starter-kit-htsget/tree/feature/rust-htsget)

**How to test**
- get a token: `token=$(curl -s -k https://localhost:8080/tokens | jq -r '.[0]')`
- to download the complete file:  `curl 'http://localhost:8443/s3-encrypted/DATASET0001/htsnexus_test_NA12878.bam' -H "Authorization: Bearer $token" -o test.bam.c4gh `
- to download the first 3000 bytes:  `curl 'http://localhost:8443/s3-encrypted/DATASET0001/htsnexus_test_NA12878.bam' -H "Authorization: Bearer $token" -o test3000.bam.c4gh -H "Range: bytes=0-3000`
- to download the first 3000 bytes, and adding extra bytes to align to the data package boundaries :  `curl 'http://localhost:8443/s3-encrypted/DATASET0001/htsnexus_test_NA12878.bam?endCoordinate=3000' -H "Authorization: Bearer $token" -o test3000ok.bam.c4gh`
- to get the file info `curl --head -i 'http://localhost:8443/s3-encrypted/DATASET0001/htsnexus_test_NA12878.bam' -H "Authorization: Bearer $token"`

*Further comments*
- `startCoordinate` is not implemented (see neicnordic/sensitive-data-archive#696)
- also see neicnordic/sensitive-data-archive#400 and neicnordic/sensitive-data-archive#698 for future work that will be required


Makefile
```diff
@@ -92,7 +92,7 @@ lint-sda-auth:
 	@cd sda-auth && golangci-lint run $(LINT_INCLUDE) $(LINT_EXCLUDE)
 lint-sda-download:
 	@echo 'Running golangci-lint in the `sda-download` folder'
-	@cd sda-download && golangci-lint run $(LINT_INCLUDE) $(LINT_EXCLUDE)
+	@cd sda-download && /home/malin/.yarn/bin/bin/golangci-lint run $(LINT_INCLUDE) $(LINT_EXCLUDE)
```
pontus, line 95:
```suggestion
	@cd sda-download && golangci-lint run $(LINT_INCLUDE) $(LINT_EXCLUDE)
```

sda-download/.github/integration/tests/common/70_check_download.sh
```diff
@@ -0,0 +1,96 @@
...
+# download encrypted full file, check that it can be decrypted correctly 
+curl --cacert certs/ca.pem -H "Authorization: Bearer $token" "https://localhost:8443/s3-encrypted/$dataset/$file" --output full3.bam.c4gh
+full_file_size=$(stat -c %s full3.bam.c4gh)  # Get the size of the file
+expected_encrypted_size=65688
+
+if [ "$file_size" -ne "$expected_encrypted_size" ]; then
+    echo "Incorrect file size for full encrypted file"
+    exit 1
+fi
+
```
pontus, line 53: Try decryption of partial requested file?

sda-download/api/api.md
```diff
@@ -0,0 +1,54 @@
+# API
```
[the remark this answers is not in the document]
MalinAhlberg: Thanks for the review :+1: !
Yes... I thought it would be nice to have it under `api/` for consistency with the services in `sda/cmd`. But you are right that it's better to have things in one place, so I merged the `docs/API.md` with the new one in
c49f140abc9f2597733b53b1bc17cdcd424b108c. Please have look!

sda-download/.github/integration/tests/common/70_check_download.sh
MalinAhlberg: Thanks for reviewing!
I cleaned up the test file a bit, decryption of partial file is now [here](https://github.com/neicnordic/sensitive-data-archive/pull/695/files#diff-2368e566965256f35cd63e8867ada6d0290c2ab37430f948bf3b6fc7c2d043deR74).  The decrypted file size is also checked there.

sda-download/.github/integration/tests/common/70_check_download.sh
```diff
@@ -0,0 +1,96 @@
...
+    echo "Too big file size for partial decrypted file"
+    exit 1
+fi
+
+crypt4gh decrypt --sk c4gh.sec.pem < part2.bam.c4gh > part2.bam
+status=$?
+if [[ $status != 0 ]]; then
+    echo "Partial encrypted file could not be decrypted"
+    exit 1
+fi
```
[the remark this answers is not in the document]
MalinAhlberg: File size is now checked. Checking the file content is a good idea, but I think it's out of scope for this PR.

sda-download/api/sda/sda.go
```diff
@@ -305,3 +347,32 @@ var sendStream = func(reader *streaming.Crypt4GHReader, writer http.ResponseWrit
...
+
+				return a, b
+			}
+		}
+	}
+	// Adapt end coordinate to follow the crypt4gh block boundaries
+	headlength := bytes.NewReader(fileDetails.Header)
+	bodyEnd := int64(fileDetails.ArchiveSize)
+	if end > 0 {
+		var packageSize float64 = 65564 // 64KiB+28, 28 is for chacha20_ietf_poly1305
```
MalinAhlberg, line 370: @pontus @blankdots is the packageSize robust enough, or should the `+28` be more variable?

sda-download/.github/integration/tests/common/70_check_download.sh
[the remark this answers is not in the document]
MalinAhlberg: But @pontus do you think that the expected partial file sizes are reasonable?
The `part_expected_size=65688` seems good to me, one data package+header, I'm not sure about the size of the decrypted file (`part_decrypted_size=65536`).

sda-download/api/s3/s3.go
```diff
@@ -294,7 +333,11 @@ func Download(c *gin.Context) {
 		ListBuckets(c)
 
 	case c.Param("filename") != "":
-		GetObject(c)
+		if strings.Contains(c.Request.URL.String(), "encrypted") {
```
pontus, line 336: So, downloading a file called `this-is-not-encrypted` should always trigger encrypted objects? 

Probably fine with something like `strings.HasPrefix(c.Request.URL.Path, "/s3-encrypted")` but haven't tested.

sda-download/.github/integration/tests/common/70_check_download.sh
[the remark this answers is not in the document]
pontus: If you request data between coordinates 0 and 1000, I believe you should get a decrypted size of 1000.

sda-download/api/sda/sda.go
pontus: Good enough for now. In the future, we should probably offer this as convenience functions in crypt4gh (where you'd call with a file header and get back e.g. this size).

pontus:
I'm not sure what's in scope for this PR, should header reencryption be included or not? Some things may be premature without that.

sda-download/.github/integration/tests/common/70_check_download.sh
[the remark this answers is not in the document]
pontus: (see https://github.com/neicnordic/sensitive-data-archive/pull/695#issuecomment-1978953453)

sda-download/api/s3/s3.go
MalinAhlberg: :see_no_evil:
Good catch. Fixed in 6db0315.

sda-download/.github/integration/tests/common/70_check_download.sh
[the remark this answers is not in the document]
MalinAhlberg: Good point, for future work. I added a comment about it in #696.

MalinAhlberg:
> I'm not sure what's in scope for this PR, should header reencryption be included or not? Some things may be premature without that.

Valid points! The scope of this PR is to be able to communicate and do requests with htsget + sda-download. Some future work is mentioned under *Further comments* in the PR body.
We use [these headers](https://github.com/neicnordic/sensitive-data-archive/pull/695/files#diff-e2ed4b43ad03c7e2d659f27710da9574602e495566a116f1babffb904b07b038R125) in communication with htsget now, for the header encryption.

So, no, all things are not in place yet, but maybe good as a first step?

blankdots:
I would say, if we can keep the changes in the scope of `htsget download endpoints` that is ok, but removing the `c4gh.password` and adding info and options about re-encryption e.g. `Client-public-key` or `Server-public-key`  maybe we should do the changes with that, as it will be easier to trace.

the reason i am mentioning this is because is while for htsget re-encryption might work in a certain way and that is ok, for serving decrypted filles we might need to generate  fresh key pair every time (see changes in sda.go from https://github.com/neicnordic/sda-download/pull/330/files) .

sda-download/api/api.md
```diff
@@ -0,0 +1,161 @@
...
+*Partial file retrieval*: 
+- `startCoordinate` Start byte position in the file. If the request is for an encrypted file, the position will be adjusted to align with the nearest data block boundary.
+- `endCoordinate` End byte position in the file. If the request is for an encrypted file, the position will be adjusted to align with the nearest data block boundary.
+
+Headers:
+
+- `Authorization: Bearer <token>` 
+- `Range: bytes=<start>-<end>`  exact positions. Overrides parameter coordinates.
+- `Client-public-key: <key>` used for re-encrypting the header of the file before sending it.
+- `Server-public-key: <key>` used in communication with htsget, for re-encrypting the header of the file.
```
blankdots, line 126: nitpicking 
- can't see to seem to find how are these handled
- do these come in the same request that they need to be separate ?
- there was a diagram how these requests looked like https://github.com/neicnordic/sda-download/blob/fbc45ed5885524032c5b0e0dac859c17a7f1af85/biohackathon/paper.md#htsget a new one how this would look like would help

blankdots: reason I am asking is that:
- i think it can be the same header name if those come in separate requests
- if it is htsget server that includes those headers the request i think we can have that header be an array
- (opinion) in general maybe it should be an array so that we allow multiple keys

MalinAhlberg: Here is the [new diagram](https://raw.githubusercontent.com/umccr/htsget-rs/crypt4gh/docs/crypt4gh/htsget-rs-crypt4gh.drawio.png), for context [see  this](https://github.com/umccr/htsget-rs/blob/crypt4gh/docs/crypt4gh/ARCHITECTURE.md#the-protocol).

MalinAhlberg: We do not use these headers yet, but the plan is to do so as soon as we have the re-encryption server connected (#698).

blankdots: then maybe we can include them with PR for that issue.

sda-download/internal/config/config.go
```diff
@@ -163,7 +163,7 @@ func NewConfig() (*Map, error) {
 		}
 	}
 	requiredConfVars := []string{
-		"db.host", "db.user", "db.password", "db.database", "c4gh.filepath", "c4gh.passphrase", "oidc.configuration.url",
```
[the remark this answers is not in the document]
blankdots: i am ok with this.

sda-download/api/api.go
```diff
@@ -45,6 +45,8 @@ func Setup() *http.Server {
 	router.GET("/files/:fileid", SelectedMiddleware(), sda.Download)
 	router.GET("/s3/*path", SelectedMiddleware(), s3.Download)
 	router.HEAD("/s3/*path", SelectedMiddleware(), s3.Download)
+	router.HEAD("/s3-encrypted/*path", SelectedMiddleware(), s3.Download)
+	router.GET("/s3-encrypted/*path", SelectedMiddleware(), s3.Download)
```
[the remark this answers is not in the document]
pontus: Thanks for pointing that out, I kind of forgot!

So, just to sync up - to me it makes the most sense to have a single instance either deliver only re-encrypted or decrypted (for use cases download to off-platform client vs on-platform, in bigpicture terms). 

So, a configurable which one it is.

I don't believe adding this as a switch now should make it more difficult to develop the functionality you need?

(I don't see a problem with delivering re-encrypted content to those eligible to receive decrypted content, but for bigpicture use case we MUST NOT deliver decrypted to those only eligible to receive re-encrypted.)

sda-download/api/api.md
blankdots: talked with @dbampalikis and understood why these headers are like this, but still splitting them from this PR makes more sense

sda-download/.github/integration/tests/common/70_check_download.sh
```diff
@@ -0,0 +1,102 @@
...
+curl --cacert certs/ca.pem -H "Authorization: Bearer $token" "https://localhost:8443/s3/$dataset/$file" --output full1.bam
+file_size=$(stat -c %s full1.bam)  # Get the size of the file
+
+if [ "$file_size" -ne "$expected_size" ]; then
+    echo "Incorrect file size for full decrypted file"
+    exit 1
+fi
+
+# test that start, end=0 returns the whole file
+curl --cacert certs/ca.pem -H "Authorization: Bearer $token" "https://localhost:8443/s3/$dataset/$file?startCoordinate=0&endCoordinate=0" --output full2.bam
```
nanjiangshu, line 27: As I tested, when start=0 and end=0, the command did not return a full file, instead it was terminated without returning any data. Did I make a mistake or is that because the `startCoordinate` has not been implemented?

sda-download/.github/integration/tests/common/70_check_download.sh
```diff
@@ -0,0 +1,102 @@
...
+crypt4gh decrypt --sk c4gh.sec.pem < full3.bam.c4gh > full3.bam
+cmp --silent full1.bam full3.bam
+status=$?
+if [[ $status != 0 ]]; then
+    echo "Full encrypted files is not correct when decrypting"
+    exit 1
+fi
+
+# download full encrypted file, test that start, end=0 returns the whole file
+curl --cacert certs/ca.pem -H "Authorization: Bearer $token" "https://localhost:8443/s3-encrypted/$dataset/$file?startCoordinate=0&endCoordinate=0" --output full4.bam.c4gh
```
nanjiangshu, line 55: I think it will be good to add these tests also to the file `./dev_utils/run_integration_test.sh` and `dev_utils/run_integration_test_no_tls.sh`

sda-download/api/api.md
```diff
@@ -0,0 +1,161 @@
...
+Further, it enables the endpoints `/s3` and `/s3-encrypted`, used for htsget.
+
+The response can be restricted to only contain a given range of a file, and the files can be returned encrypted or decrypted.
+
+All endpoints require an `Authorization` header with an access token in the `Bearer` scheme.
+```
+Authorization: Bearer <token>
+```
+### Authenticated Session
+The client can establish a session to skip time-costly visa validations for further requests. Session is based on the `SESSION_NAME=sda_session_key` (configurable name) cookie returned by the server, which should be returned in later requests.
```
nanjiangshu, line 16:
```suggestion
The client can establish a session to bypass time-costly visa validations for further requests. The session is established based on the `SESSION_NAME=sda_session_key` (configurable name) cookie returned by the server, which should be included in later requests.
```

sda-download/api/api.md
```diff
@@ -0,0 +1,161 @@
...
+- `/files/:fileid`
+
+**htsget**:
+
+- `/s3/<datasetid>/<fileid>`
+- `/s3-encrypted/<datasetid>/<fileid>`
+
+### Data out API
+#### Datasets
+The `/metadata/datasets` endpoint is used to display the list of datasets the given token is authorised to access, that are present in the archive.
```
nanjiangshu, line 35:
```suggestion
The `/metadata/datasets` endpoint is used to display the list of datasets that the given token is authorised to access, and that are present in the archive.
```

sda-download/api/api.md
```diff
@@ -0,0 +1,161 @@
...
+##### Response
+```
+[
+    "dataset_1",
+    "dataset_2"
+]
+```
+#### Files
+##### Request
+Files contained by a dataset are listed using the `datasetName` from `/metadata/datasets`.
```
nanjiangshu, line 49:
```suggestion
The files contained in a dataset are listed using the `datasetName` obtained from `/metadata/datasets` endpoint.
```

sda-download/api/api.md
```diff
@@ -0,0 +1,161 @@
...
+```
+GET /metadata/datasets/{datasetName}/files
+```
+**Scheme Parameter**
+The `?scheme=` query parameter is optional. When a dataset contains a scheme, it may sometimes be problematic with reverse proxies.
```
nanjiangshu, line 54:
```suggestion
The `?scheme=` query parameter is optional. When a dataset contains a scheme, it may sometimes encounter issues with reverse proxies.
```

sda-download/api/api.md
```diff
@@ -0,0 +1,161 @@
...
+The scheme can be split from the dataset name, and supplied in a query parameter.
```
nanjiangshu, line 55:
```suggestion
The scheme can be separated from the dataset name and supplied in a query parameter.
```

sda-download/api/api.md
```diff
@@ -0,0 +1,161 @@
...
+```
+?startCoordinate=0&endCoordinate=100
+```
+
+### S3 requests, for htsget
+
+**Parameters**:
+
+*Partial file retrieval*: 
+- `startCoordinate` Start byte position in the file. If the request is for an encrypted file, the position will be adjusted to align with the nearest data block boundary.
```
nanjiangshu, line 118:
```suggestion
- `startCoordinate`: start byte position in the file. If the request is for an encrypted file, the position will be adjusted to align with the nearest data block boundary.
```

sda-download/api/api.md
```diff
@@ -0,0 +1,161 @@
...
+- `endCoordinate` End byte position in the file. If the request is for an encrypted file, the position will be adjusted to align with the nearest data block boundary.
```
nanjiangshu, line 119:
```suggestion
- `endCoordinate`: end byte position in the file. If the request is for an encrypted file, the position will be adjusted to align with the nearest data block boundary.
```

sda-download/api/s3/s3.go
```diff
@@ -188,6 +188,26 @@ func ListObjects(c *gin.Context) {
 	})
 }
 
+func getFileInfo(c *gin.Context) (fileInfo *database.FileInfo, err error) {
+	// Get file info for the given file path (or abort)
+	filename := c.Param("filename")
+	if !strings.HasSuffix(c.Param("filename"), ".c4gh") {
+		filename = c.Param("filename") + ".c4gh"
```
nanjiangshu, line 195: Why append the suffix ".c4gh" to the filename? How if it's an encrypted file?

nanjiangshu (commented):
Great job @MalinAhlberg! I looked through the PR and added a few comments. However, I encountered some problems to run tests that are described in the PR description. 
1. I have to use the command 
`token=$(curl -s -k http://localhost:8000/tokens | jq -r '.[0]')` to obtain the token, this is because I started the stack with `notls`, but still I think port  number `8080` might be erroneous. 

2. I don't know how the data file `htsnexus_test_NA12878.bam` can be input in to the archive, and therefore, I haven't tested the endpoint `s3-encrypted`. 

Apart from that, when I was setting up the stack for sda-download according to README at https://github.com/neicnordic/sensitive-data-archive/tree/main/sda-download/dev_utils, I found the instructions were not clear enough. 
E.g. the `download` container won't start with docker compose without the files `c4gh.pub.pem` and `c4gh.sec.pem` and these two files are created by integration test file. I think the README should be modified accordingly.

sda-download/api/s3/s3.go
MalinAhlberg: The file extension was previously added by default (see [the remove line here](https://github.com/neicnordic/sensitive-data-archive/commit/32478c8f5e8fd77c4997b7ac7b21b5ef676de5f2#diff-481dfe13739d77ba443566773ab0564d718e309473b081baebbbc4acd355a7a8R193)), we now check that we don't get an extra suffix, thus allowing the client to require `file.c4gh` .

sda-download/api/api.md
MalinAhlberg: Headers are removed in 8cece7d

sda-download/api/sda/sda.go
```diff
@@ -188,61 +241,58 @@ func Download(c *gin.Context) {
...
+			c.String(http.StatusInternalServerError, "an error occurred")
 
-	if start == 0 && end == 0 {
-		c.Header("Content-Length", fmt.Sprint(fileDetails.DecryptedSize))
-	} else {
-		// Calculate how much we should read (if given)
-		togo := end - start
-		c.Header("Content-Length", fmt.Sprint(togo))
+			return
+		}
```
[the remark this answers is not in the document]
MalinAhlberg: Fixed in c568788034902da088ce146264845f51f6efc811

sda-download/api/sda/sda_test.go
```diff
@@ -504,3 +505,36 @@ func TestDownload_Fail_OpenFile(t *testing.T) {
...
+	_, end = calculateEncryptedCoords(from, headerSize-10, "", fileDetails)
+	assert.GreaterOrEqual(t, end, headerSize)
+
+	// end should not be larger than file length + header
+	_, end = calculateEncryptedCoords(from, fullSize+1900, "", fileDetails)
+	assert.Equal(t, fullSize, end)
+
+	// range 0-0 should give whole file
+	start, end = calculateEncryptedCoords(0, 0, "", fileDetails)
+	assert.Equal(t, end-start, fullSize)
```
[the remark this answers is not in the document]
MalinAhlberg: Added in 3b4692eb9612443c4170fedbe3365e25a854d0fc

sda-download/api/sda/sda.go
```diff
@@ -305,3 +347,32 @@ var sendStream = func(reader *streaming.Crypt4GHReader, writer http.ResponseWrit
...
+	if end > 0 {
+		var packageSize float64 = 65564 // 64KiB+28, 28 is for chacha20_ietf_poly1305
+		togo := end - start
+		bodysize := math.Max(float64(togo-headlength.Size()), 0)
+		endCoord := packageSize * math.Ceil(bodysize/packageSize)
+		bodyEnd = int64(math.Min(float64(bodyEnd), endCoord))
+	}
+
+	return start, headlength.Size() + bodyEnd
+}
```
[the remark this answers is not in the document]
MalinAhlberg: In 43c8c2210acd91cba52465cbddeb1e3455390b8a

sda-download/.github/integration/tests/common/70_check_download.sh
MalinAhlberg: As far as I know, the `dev_utils` is to be removed. I'm not sure how the tests should be structured in the future (@jbygdell?)

sda-download/api/api.md
```diff
@@ -0,0 +1,161 @@
+# API
+
+
+The Download API service provides functionality for downloading files from the Archive.
+It implements the [Data Out API](https://neic-sda.readthedocs.io/en/latest/dataout/#rest-api-endpoints).
+
+Further, it enables the endpoints `/s3` and `/s3-encrypted`, used for htsget.
+
+The response can be restricted to only contain a given range of a file, and the files can be returned encrypted or decrypted.
```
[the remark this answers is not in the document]
MalinAhlberg: Good point! I updated this in 0414773df3c649b288bb2db14c0bad686d383f28

MalinAhlberg:
Thanks for the careful review, @nanjiangshu! I added of your suggestions, see above.
Regarding the instructions, I realize that the description of this PR could have been better. For the general case, I think we are still lacking a streamed-line way of testing and running things locally. There are also refactorings coming for the `dev_utils`. I think it would be great if you could add your ideas to [this issue](https://app.zenhub.com/workspaces/giant-sloth-6041ec3fd9297d0016321f73/issues/gh/nbisweden/localega-se-deployment/708).

nanjiangshu (approved):
Everything looks good!

sda-download/.github/integration/tests/common/70_check_download.sh
MalinAhlberg: Fixed in pair session

sda-download/api/s3/s3.go
```diff
@@ -196,15 +216,31 @@ func ListObjects(c *gin.Context) {
...
 	log.Debugf("S3 GetObject request, context: %v", c.Params)
 
-	// Get file info for the given file path (or abort)
-	fileInfo, err := database.GetDatasetFileInfo(c.Param("dataset"), c.Param("filename")+".c4gh")
+	fileInfo, err := getFileInfo(c)
 	if err != nil {
-		if err.Error() == "sql: no rows in result set" {
-			c.AbortWithStatus(http.StatusNotFound)
-		} else {
-			c.AbortWithStatus(http.StatusInternalServerError)
```
pontus, line 205: Shouldn't we flag the error here so the client gets a 5xx? (I don't see any need to look at the error, ISE should be good enough).

sda-download/api/sda/sda.go
```diff
@@ -166,6 +167,60 @@ func Download(c *gin.Context) {
 		return
 	}
 
+	// Get query params
+	qStart := c.DefaultQuery("startCoordinate", "0")
```
pontus, line 171: Maybe put detection of coordinates in a function of its own, looking at the query or range header according to the priority we want to handle them in?

sda-download/api/sda/sda.go
```diff
@@ -166,6 +167,60 @@ func Download(c *gin.Context) {
...
+		c.String(http.StatusBadRequest, "endCoordinate must be greater than startCoordinate")
+
+		return
+	}
+
+	// Calculate the content length
+	contentLength := fileDetails.DecryptedSize
+	if c.Param("type") == "encrypted" {
+		contentLength = fileDetails.ArchiveSize
+		start, end, err = calculateEncryptedCoords(start, end, c.GetHeader("Range"), fileDetails)
```
pontus, line 201: Feels odd to manage coordinates coming from the header within the `calculateEncryptedCoords` when coordinates coming in query is handled before.

MalinAhlberg: For unencrypted data, the start and end are just parsed from the query params (line 171 and below). It's only for the encrypted case that we look at the range header and adapt the coords to the data boundaries. One could of course break out the coord parsing code into a separate function, to make it easier to follow, something like (starting line 170):
```go
	start, end, err := parseCoords(c) // get the coords from the params

	// Calculate the content length
	contentLength := fileDetails.DecryptedSize
	if c.Param("type") == "encrypted" {
                // if encrypted, check the range header + adapt the previously set start and end so that it match boundaries
		start, end, err = calculateEncryptedCoords(start, end, c.GetHeader("Range"), fileDetails)
```

sda-download/api/sda/sda.go
```diff
@@ -166,6 +167,60 @@ func Download(c *gin.Context) {
...
+			// reading from an offset in encrypted file is not yet supported
+			c.Header("Content-Length", "0")
+			log.Errorf("Start coordinate for encrypted files not implemented! %v", start)
+			c.String(http.StatusBadRequest, "Start coordinate for encrypted files not implemented!")
+
+			return
+		}
+	}
+	if start == 0 && end == 0 {
+		c.Header("Content-Length", fmt.Sprint(contentLength))
```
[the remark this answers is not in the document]
MalinAhlberg: The `contentLength` is already set to match either the decrytped or encrypted file size, see line 198 and 200 above. Maybe we can put the case of re-encrypted as a TODO for that coming PR?

sda-download/api/s3/s3.go
MalinAhlberg: This is good point, I'm not sure if there's any good rule either. These changes were done [quite a while ago](https://github.com/neicnordic/sda-download/commit/170f9a024449c0a81e4107bcebe67cda0ec7fd06), I do not know exactly the reason. I guess there should be tests to verify that we do get the correct error for such cases.

sda-download/api/sda/sda.go
pontus: Yes, I see that they are managed differently and I would prefer if that wasn't the case.

For bigpicture, we have a strong use case for supporting range request as well.

So I'd prefer the "figuring out what parts the user wants" to be refactored into a separate function. I'm fine with a TODO but think it can be done now as well.

sda-download/api/s3/s3.go
pontus: Ok, created https://github.com/neicnordic/sensitive-data-archive/issues/747 for now.

sda-download/api/sda/sda.go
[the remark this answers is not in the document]
pontus: I think that the if set, Content-length should match the amount of data sent. So either we could not set it for the encrypted case, or it should include the crypt4gh header?

sda-download/api/sda/sda.go
pontus: Same  as https://github.com/neicnordic/sensitive-data-archive/pull/695#discussion_r1522705254, resolving this.

sda-download/api/sda/sda.go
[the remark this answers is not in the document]
MalinAhlberg: In 7994e9f, the content-length is not explicitly set for the encrypted case, expect for when the request is asking only for the size, using the head method, see [here](https://github.com/neicnordic/sensitive-data-archive/pull/695/commits/7994e9fa4272be2e2441a9206cd6b5c8c6fde2a3#diff-901816ca78cf2bc6802f2ab6f348e47aa5a63053ec429cef96b3831680d957a3R248-R252). That case is needed for the htsget implementation. Is this ok for now?  It works with htsget, and can be improved in #698 .

sda-download/api/sda/sda.go
MalinAhlberg: I created #749 for this.

sda-download/api/api.go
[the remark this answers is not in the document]
MalinAhlberg: I created #750 feel free to improve and refine this

pontus (approved):

aaperis (approved):
Looks good!

PR #700, opened 2024-02-28
[Merge] sda-auth

jbygdell:
Related to #14 

**Description**

This PR merges `sda-auth` into `sda` under `cmd/auth` with as few modifications as possible.
`auth` is added to the support service section of  [sda.md](https://github.com/neicnordic/sensitive-data-archive/blob/merge/sda-auth/sda/sda.md) 

Changes made to auth in this PR:
* Replace `elixir` terms with generic `oidc`
* Split up long config variables in order to be able to use yaml nesting in the config file

helm chart version is bumped to `0.24.0` for `sda-svc`
tag is bumped to `v0.3.0` (hopefully the automated tag bump will work)

What's missing is a test setup to be able to properly test the OIDC login flow, possible solution is to make use of the `ls-aai-mock` from the GDI project.


MalinAhlberg (approved):
Looks good to me

.github/integration/tests/sda/auth/integration/__init__.py
pontus, line 1: Are these intended/needed?

jbygdell: Not really, but they will be recreated when you run the tests locally. So by having them in there already makes it cleaner by them not showing up as created afterwards. optionally they can be added to the .gitignore file

pontus: Ok, I'd prefer sticking them in `.gitignore` then.

pontus (approved):

MalinAhlberg:
Remove `go.mod` and `go.sum` from `sda-auth` as well?

jbygdell:
> Remove `go.mod` and `go.sum` from `sda-auth` as well?

Interesting that they came back after doing `rm -rf sda-auth`

MalinAhlberg (dismissed):
:+1: 
Is there issue(s) created for the things noted as missing in the PR description?

.github/integration/sda-s3-integration.yml
```diff
@@ -317,13 +317,78 @@ services:
...
+    command: [ sda-auth ]
+    container_name: auth-oidc
+    depends_on:
+      oidc:
+        condition: service_healthy
+    environment:
+      - AUTH_RESIGNJWT=false
+      - OIDC_ID=XC56EL11xx
+      - OIDC_SECRET=wHPVQaYXmdDHg
+    image: ghcr.io/neicnordic/sensitive-data-archive:PR${PR_NUMBER}
```
aaperis, line 371: Why have two services? Is it just to test that they can deploy with separate logins? Seems a bit too cluttered.

.github/integration/sda-s3-integration.yml
```diff
@@ -317,13 +317,78 @@ services:
...
+    restart: always
+    volumes:
+      - ./sda/config.yaml:/config.yaml
+      - shared:/shared
+
+  auth-oidc:
+    command: [ sda-auth ]
+    container_name: auth-oidc
+    depends_on:
+      oidc:
```
aaperis, line 365: Using the python mock oidc limits using this compose as a dev environment since the python oidc lacks an authorization endpoint and auth will not work.

aaperis (commented):
This looks great! I really appreciated the verbosity of the commits :-). 

I am only concerned with the lack of a dev environment if we take away the node js mock oidc (see code comment). 
As a thought, it  seems more sustainable to me to make that serving visas for mocking with sda-download  (i.e. have a fully fledged oidc server )  than adding code to oidc.py to make it work with sda-auth's usecases. 

Edit: should have read the PR description more carefully. Adding the test mock could be a separate Issue imo and until then we leave somewhere the node oidc?

.github/integration/tests/sda/99_auth_test.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+set -e
+
+python -m pip install --upgrade pip
+pip install tox
+
+tox -e unit_tests -c /tests/sda/auth/tox.ini
```
aaperis, line 7: Should add an Issue for adding integration testing for the OIDC case? and possibly refactor the EGA test as well..

.github/integration/sda-s3-integration.yml
jbygdell: Because the auth tests are fubar as pointed out in the PR text, but this can be solved later since there are other service that rely on the NSS/OIDC endpoints.

.github/integration/tests/sda/99_auth_test.sh
jbygdell: The entire sda test suite are getting a bit to big to have all in one, so an issue for cleaning up the tests is a good idea.

.github/integration/sda-s3-integration.yml
aaperis: Maybe I was unclear, I meant why not configure one `auth:` with both cega and oidc. But in any case, testing can be revisited on a separate issue.

aaperis:
Small thing, but I noticed there is an empty folder `sda-auth/tests/` left in root.

jbygdell:
> Small thing, but I noticed there is an empty folder `sda-auth/tests/` left in root.

That's only for you locally

aaperis (dismissed):

jbygdell:
@aaperis and @MalinAhlberg can you please re-approve this one so it can be merged.

aaperis (approved):

MalinAhlberg (approved):

Issue #696, opened 2024-02-28
[sda-download] Implement random access in encrypted files


dbampalikis:
As an sda-user
I want to be able to download specific parts of an encrypted file
In order to be able to get only the region I am interested in

The service currently allows to download specific byte ranges of unencrypted files but in the case of encrypted files, that's only possible for byte ranges that start from the beginning of the file. We need to be able to support random byte ranges of encrypted files, to support the htsget case.

A/C
- A PR that allows for downloading specific ranges of encrypted files

pontus:
Assuming we want to avoid reencryption of possibly large amounts of data, this should use the intended support for this in the crypt4gh file format.

In short, each file/data stream is split into 64kbyte blocks that are encrypted/ separately. This is also the smallest unit for decryption as these blocks are what MACs are created for.

This means that to send logical byte 65535-65536 (base 0), one would need to send the reencrypted header and the first two data blocks (65536+extra bytes for crypt4gh). As the receiver only want those two bytes, there would also need to be a data edit list in the header to instruct it to throw away bytes 0-65534 and 65537-131071.

So the header reencryption service needs to be able to accept a dataeditlist to be put in the header.

Currently, I think there's only the chacha20_ietf_poly1305 cipher, so a fixed block size of 65564 can be used, but possibly it might make sense to have a function in the crypt4gh library that takes a header and responds with the block size (or similar).

pontus:
For both the unencrypted and encrypted data out case, there will also be a performance motive to not request the entire object from the archive and only return the wanted bit but rather only requesting the range actually needed.

For the encrypted case, this is fairly simple - the s3 download client could pass a `Range` with the bytes wanted.

The question would be if we would prefer having a unified handling for unencrypted and encrypted.

For the unencrypted case, it might make sense to have a reader that maps calls to `Read` to a s3 call that is essentially managed synchronously or something similar.

MalinAhlberg:
When decrypting a partial file, the resulting file size should be what was originally asked for, not more. Ie, the extra data passed on to meet the next data boundary block should be removed. Use `data-edit-list`. See https://github.com/neicnordic/sensitive-data-archive/pull/695#discussion_r1512952339

Issue #363, opened 2023-10-19
[sda-download] Implement byte range handling to support random access


pontus:
To support random-access, sda-download can handle `Range` requests.

This could include:
[ ] Send `Accept-Ranges: bytes` header on request
[ ] Change the S3 code to support getting the requested data only, provide as a `ReadSeeker`  
[ ] If serving unencrypted, make sure to provide the s3 as a `ReadSeeker` to `NewCrypt4GHReader` and serve the needed region(s) from the `Crypt4GHReader`
[ ] If serving encrypted serve the needed region(s)
[ ] Update tests

sda-download currently supports getting a segment of the (unencrypted) data by URL parameters `startCoordinate` and `endCoordinate` (https://neic-sda.readthedocs.io/en/latest/dataout/#rest-api-endpoints) these likely also needs to be supported going forwards.

pontus:
We also would need to define/document what a `Range` means for the encrypted case - should the range be interpreted as bytes in the logical datastream that would be the result of decryption or something else? Less controversial for coordinates passed in URL, but it would feel odd to respond to a request with `Range: 10-11` with a 64 kbyte response.

pontus:
This is done, although there are various issues with the implementation.

PR #703, opened 2024-02-29
[WIP] Add support for random access to backend objects, not fetching entire objects.

pontus:
This helps parts for #363 and #696.

### DIFF
--- a/sda-download/internal/storage/storage.go
+++ b/sda-download/internal/storage/storage.go
@@ -353,7 +353,6 @@ func transportConfigS3(config S3Conf) http.RoundTripper {
 }
 
 func (sb *s3SeekableBackend) NewFileReader(filePath string) (io.ReadCloser, error) {
-
 	s := sb.s3Backend
 	objectSize, err := s.GetFileSize(filePath)
 
@@ -377,7 +376,6 @@ func (sb *s3SeekableBackend) NewFileReader(filePath string) (io.ReadCloser, erro
 }
 
 func (sb *s3SeekableBackend) GetFileSize(filePath string) (int64, error) {
-
 	s := sb.s3Backend
 
 	return s.GetFileSize(filePath)
@@ -402,7 +400,6 @@ func (r *s3SeekableReader) pruneCache() {
 }
 
 func (r *s3SeekableReader) prefetchSize() int64 {
-
 	n := r.Conf.Chunksize
 
 	if n >= 5*1024*1024 {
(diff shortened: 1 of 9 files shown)


pontus:
I've done some performance tests with this code, just downloading a dataset shared with someone.

With the old s3 driver (leaving it mostly up to aws sdk), it downloads 87239339037 bytes in 4 minutes 36s (~ 316 Mbyte/second). 

After adding some performance optimisations I see the same takes 6 minutes 19 seconds (~219 Mbyte/second) with the seekable driver (a lot of that comes from needing to fail if the object doesn't exist/can't be accessed, which makes me do a `HEAD` call).

sda-download/internal/storage/storage.go
```diff
@@ -140,6 +167,15 @@ type S3Conf struct {
...
 
+func newS3SeekableBackend(config S3Conf) (*s3SeekableBackend, error) {
+	sb, err := newS3Backend(config)
+	if err != nil {
+		return nil, err
+	}
+
+	return &s3SeekableBackend{*sb}, nil
+}
+
```
jbygdell, line 178: Is a dedicated backed really needed?
It is after all just a duplication of the existing backend.

sda-download/internal/storage/storage.go
```diff
@@ -315,3 +351,439 @@ func transportConfigS3(config S3Conf) http.RoundTripper {
...
+		filePath,
+		objectSize,
+		sync.Mutex{},
+		make([]int64, 0, 32),
+		false,
+		nil,
+	}
+
+	return reader, nil
+}
```
jbygdell, line 376: Wouldn't the implementation of this as a `SeekableFileReader` in the normal backend be the same, so we can do both types of file reads depending on the incoming query.

sda-download/internal/storage/storage.go
pontus: Correct.

I made it as a separate to be able to possibly merge it but still use the previous functionality only if you do not feel confident with it (e.g. your use cases don't care about delivering parts and you prefer higher performance). Can also be imlemented with a parameter to control behaviour of the driver.

sda-download/internal/storage/storage.go
pontus: True. I made this in a way to keep the `s3` driver unchanged for now, but in an ideal future scenario, I would like to see a single driver (supporting seeking), possibly offering some more controls to adjust how it works.

MalinAhlberg:
I tested this with htsget, using [this branch](https://github.com/neicnordic/sensitive-data-archive/commit/e947be245e8a2e33b6617c92d38254d7cc566603) (wip, not cleaned-up) . Everything seems to work well!

pontus:
See #826

PR #708, opened 2024-03-04
Refactor re-encrypt

jbygdell:
**Description**
* Adds the c4gh to the server struct so we don't need any global variables
* removes all global variable declarations

**How to test**
make sda-integrationtest

### DIFF
--- a/sda/cmd/reencrypt/reencrypt.go
+++ b/sda/cmd/reencrypt/reencrypt.go
@@ -27,6 +27,7 @@ import (
 // server struct is used to implement reencrypt.ReEncryptServer.
 type server struct {
 	re.UnimplementedReencryptServer
+	c4ghPrivateKey *[32]byte
 }
 
 var (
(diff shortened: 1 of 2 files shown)


sda/cmd/reencrypt/reencrypt.go
```diff
@@ -57,9 +53,9 @@ func (s *server) ReencryptHeader(_ context.Context, in *re.ReencryptRequest) (*r
 	newReaderPublicKeyList := [][chacha20poly1305.KeySize]byte{}
 	newReaderPublicKeyList = append(newReaderPublicKeyList, newReaderPublicKey)
 
-	log.Debugf("crypt4ghkey: %v", *Conf.ReEncrypt.Crypt4GHKey)
+	log.Debugf("crypt4ghkey: %v", *s.c4ghPrivateKey)
```
blankdots, line 56: maybe base64 encoded it, not sure how nice to show in logs as byte

blankdots (approved):
small suggestion

sda/cmd/reencrypt/reencrypt.go
pontus: `%v` will format, but possibly `%x` gives a nicer output (`[104 101 106]` for `%v` vs `[68656a]` for `%x`).

pontus (approved):

aaperis (approved):

PR #761, opened 2024-03-18
Feat: connect reencrypt service to sda-download

### LINKED ISSUE #698: [sda-download] Connect re-encryption service
As an sda-dev
I want to be able to re-encrypt headers through sda-download
In order to be able to export files encrypted with the user's public key

The re-encryption service exists in the [sensitive-data-archive repo](https://github.com/neicnordic/sensitive-data-archive/tree/main/sda/cmd/reencrypt), but is currently not connected to sda-download

A/C
- Be able to send a public key and a header and get back a new header encrypted with this public key
- Add integration tests (see e.g. docker compose file posted below)

nanjiangshu:
This PR closes #698 

**Description**
With this PR, the users can reencrypt the header of the crypt4gh encrypted file by providing the client public key at the request header with the key `client-public-key`. 

**How to test**
1. Build the image for sda-download
In the sub-folder `sda-download`, run 
```
docker build . -t local-sda-download
```
 
2.  Set up the docker environment with [this branch](https://github.com/GenomicDataInfrastructure/starter-kit-storage-and-interfaces/tree/feat/reencrypt-header) of the repo `starter-kit-storage-and-interfaces`
with the command 
```
cp .env.example .env
cp config/config.yaml.example config/config.yaml
docker compose -f docker-compose-full-demo-test.yml  up -d 
```
3.  Create your client key pair by 
```
sda-cli createKey client
```
4. Obtain the token by 
```
token=$(curl -s -k https://localhost:8080/tokens | jq -r '.[0]') 
```
5. Get base64 encoded client key by 
```
clientkey=$(base64 -w0 client.pub.pem) 
```
6. Download the test BAM file with the client public key 
```
curl -v -H "Authorization: Bearer $token" -H "Client-Public-Key: $clientkey" http://localhost:8443/s3-encrypted/$dataset/NA12878.bam -o reenc.bam 
```
7. Verify by decrypt it with your client key and view the content
```
crypt4gh decrypt -s client.sec.pem -f reenc.bam
samtools view reenc.bam.dec
```
**Additional notes**
* The code for the reencrypt service has been copied (and thus duplicated) from `sda/internal/reencrypt` to `sda-download/internal/reencrypt` since it is not exported as public. This needs to be removed once `sda-download` is merged.


.github/integration/sda/config.yaml
```diff
@@ -68,6 +68,10 @@ c4gh:
   passphrase: "c4ghpass"
   syncPubKeyPath: /shared/sync.pub.pem
 
+grpc:
+  host: "reencrypt"
+  port: "50051"
+
```
jbygdell, line 74: This file down't need this addition yet.

sda-download/.github/integration/tests/common/80_check_reencrypt.sh
```diff
@@ -0,0 +1,94 @@
...
+    echo "Incorrect file size for full decrypted file"
+    exit 1
+fi
+
+# Test reencrypt the file header with the client public key 
+clientkey=$(base64 -w0 client.pub.pem)
+reencryptedFile=reencrypted.bam.c4gh
+curl --cacert certs/ca.pem -H "Authorization: Bearer $token" -H "Client-Public-Key: $clientkey" "https://localhost:8443/s3-encrypted/$dataset/$file" --output $reencryptedFile
+if [ ! -f "$reencryptedFile" ]; then
+    echo "Failed to reencrypt the header of the file from sda-download"
```
jbygdell, line 41:
```suggestion
    echo "Failed download re-encrypted file"
```

sda-download/.github/integration/tests/common/80_check_reencrypt.sh
```diff
@@ -0,0 +1,94 @@
...
+file_size=$(stat -c %s $reencryptedFile)
+if [ "$file_size" -ne "$expected_encrypted_size" ]; then
+    echo "Incorrect file size for the reencrypted file, should be $expected_encrypted_size but is $file_size"
+    exit 1
+fi
+
+# Descrypt the reencrypted file and compare it with the original unencrypted file
+export C4GH_PASSPHRASE="strongpass" # passphrase for the client crypt4gh key
+if ! crypt4gh decrypt --sk client.sec.pem < $reencryptedFile > full2.bam; then
+    echo "Failed to descrypt $reencryptedFile with the client public key"
```
jbygdell, line 55:
```suggestion
    echo "Failed to decrypt re-encrypted file with the clients private key"
```

sda-download/.github/integration/tests/common/80_check_reencrypt.sh
```diff
@@ -0,0 +1,94 @@
...
+fi
+
+# download reencrypted partial file, check file size
+partReencryptedFile=part1.bam.c4gh
+curl --cacert certs/ca.pem -H "Authorization: Bearer $token" -H "Client-Public-Key: $clientkey" "https://localhost:8443/s3-encrypted/$dataset/$file?startCoordinate=0&endCoordinate=1000" --output $partReencryptedFile 
+file_size=$(stat -c %s $partReencryptedFile)  # Get the size of the file
+part_expected_size=65688
+
+if [ "$file_size" -ne "$part_expected_size" ]; then
+    echo "Incorrect file size for partially reencrypted file, should be $part_expected_size but is $file_size"
```
jbygdell, line 71:
```suggestion
    echo "Incorrect file size for re-encrypted partial file, should be $part_expected_size but is $file_size"
```

sda-download/.github/integration/tests/common/80_check_reencrypt.sh
```diff
@@ -0,0 +1,94 @@
...
+    exit 1
+fi
+
+if ! crypt4gh decrypt --sk client.sec.pem < $partReencryptedFile > part1.bam; then
+    echo "Partially reencrypted file could not be decrypted"
```
jbygdell, line 76:
```suggestion
    echo "Re-encrypted partial file could not be decrypted"
```

sda-download/api/api.md
```diff
@@ -122,7 +122,9 @@ Headers:
 
 - `Authorization: Bearer <token>` 
 - `Range: bytes=<start>-<end>`  exact positions. Overrides parameter coordinates.
-- `User-agent` used in communication with htsget, to mark who is making the request.
+- `Client-public-key: <key>` used for re-encrypting the header of the file before sending it.
+- `Server-public-key: <key>` used in communication with htsget, for re-encrypting the header of the file.
+- `User-agent` used in communication with htsget, to mark who is making the request. For example `htsget-search/0.6.6.`
```
pontus, line 127:
```suggestion
- `User-Agent` used in communication with htsget, to mark who is making the request. For example `htsget-search/0.6.6.`
```

correct capsing

sda-download/.github/integration/tests/common/80_check_reencrypt.sh
```diff
@@ -0,0 +1,94 @@
...
+    exit 1
+fi
+
+part_decrypted_size=65536
+file_size=$(stat -c %s part1.bam)
+if [ "$file_size" -ne "$part_decrypted_size" ]; then
+    echo "Incorrect file size for partially decrypted file, should be $part_decrypted_size but is $file_size"
```
jbygdell, line 83:
```suggestion
    echo "Incorrect size for decrypted partial file, should be $part_decrypted_size but is $file_size"
```

sda-download/.github/integration/tests/common/80_check_reencrypt.sh
```diff
@@ -0,0 +1,94 @@
...
+    exit 1
+fi
+
+if ! grep -q "^THIS FILE IS JUST DUMMY DATA" part1.bam; then
+    echo "Bad content partial decrypted file"
```
jbygdell, line 88:
```suggestion
    echo "Bad content decrypted partial file"
```

sda-download/.github/integration/tests/common/80_check_reencrypt.sh
```diff
@@ -0,0 +1,94 @@
...
+    exit 1
+fi
+
+
+# Clean up
+rm full1.bam full2.bam part1.bam $reencryptedFile
```
jbygdell, line 94:
```suggestion
rm full1.bam full2.bam part1.bam $reencryptedFile

```

sda-download/api/sda/sda.go
```diff
@@ -29,6 +33,43 @@ func sanitizeString(str string) string {
...
+		return nil, fmt.Errorf("host of the reencrypt service not set")
+	} else if config.Config.Reencrypt.Port == 0 {
+		return nil, fmt.Errorf("port of the reencrypt service not set")
+	}
+	address := fmt.Sprintf("%s:%d", config.Config.Reencrypt.Host, config.Config.Reencrypt.Port)
+	log.Debugf("Address of the reencrypt service: %s", address)
+
+	opts := []grpc.DialOption{
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+	}
```
jbygdell, line 47: This needs to be expanded so it can take credentials, user-pass and/or client certificates

sda-download/api/sda/sda.go
```diff
@@ -29,6 +33,43 @@ func sanitizeString(str string) string {
...
+	conn, err := grpc.Dial(address, opts...)
+	if err != nil {
+		log.Errorf("Failed to connect to the reencrypt service, reason: %s", err)
+
+		return nil, err
+	}
+	defer conn.Close()
+	log.Debugf("Connection to the reencrypt service established, conn = %v", conn)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
```
jbygdell, line 57:
```suggestion
	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```

Connection timeout of 1 second is a bit harsh.

sda-download/dev_utils/compose.yml
```diff
@@ -105,6 +105,22 @@ services:
...
+      certfixer:
+        condition: service_completed_successfully
+    ports:
+      - "50051:50051"
+    restart: always
+    volumes:
+      - ./config.yaml:/config.yaml
+      - ./:/dev_utils/
+      - ./iss.json:/iss.json
+      - certs:/dev_utils/certs
```
jbygdell, line 122: Shouldn't this service also use certificates?

sda-download/dev_utils/config.yaml
```diff
@@ -44,6 +44,10 @@ c4gh:
   passphrase: "oaagCP1YgAZeEyl2eJAkHv9lkcWXWFgm"
   filepath: "./dev_utils/c4gh.sec.pem"
 
+grpc:
+    host: "reencrypt"
+    port: "50051"
```
jbygdell, line 49: Credentials are missing here

sda-download/internal/config/config.go
```diff
@@ -138,6 +139,14 @@ type DatabaseConfig struct {
 	ClientKey string
 }
 
+type ReencryptConfig struct {
+	Host       string
+	Port       int
+	CACert     string
+	ServerCert string
+	ServerKey  string
```
jbygdell, line 147:
```suggestion
	ClientCert string
	ClientKey  string
```

sda-download/internal/config/config.go
```diff
@@ -296,6 +308,20 @@ func (c *Map) configArchive() {
...
+	c.Reencrypt.Port = viper.GetInt("grpc.port")
+	if viper.IsSet("grpc.cacert") {
+		c.Reencrypt.CACert = viper.GetString("grpc.cacert")
+	}
+	if viper.IsSet("grpc.servercert") {
+		c.Reencrypt.ServerCert = viper.GetString("grpc.servercert")
+	}
+	if viper.IsSet("grpc.serverkey") {
+		c.Reencrypt.ServerKey = viper.GetString("grpc.serverkey")
+	}
```
jbygdell, line 322:
```suggestion
	if viper.IsSet("grpc.clientcert") {
		c.Reencrypt.ClientCert = viper.GetString("grpc.clientcert")
	}
	if viper.IsSet("grpc.serverkey") {
		c.Reencrypt.ClientKey = viper.GetString("grpc.clientkey")
	}
```

sda-download/api/sda/sda.go
```diff
@@ -29,6 +33,43 @@ func sanitizeString(str string) string {
...
+	} else if config.Config.Reencrypt.Port == 0 {
+		return nil, fmt.Errorf("port of the reencrypt service not set")
+	}
+	address := fmt.Sprintf("%s:%d", config.Config.Reencrypt.Host, config.Config.Reencrypt.Port)
+	log.Debugf("Address of the reencrypt service: %s", address)
+
+	opts := []grpc.DialOption{
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+	}
+	conn, err := grpc.Dial(address, opts...)
```
jbygdell, line 48: This should probably have a check function to recreate the connection if the re-encrypt server restarts

jbygdell (changes requested):

sda-download/api/sda/sda.go
```diff
@@ -29,6 +33,43 @@ func sanitizeString(str string) string {
 	return pattern.ReplaceAllString(str, "[identifier]: $1")
 }
 
+func reencryptHeader(oldHeader []byte, reencKey string) ([]byte, error) {
+	if config.Config.Reencrypt.Host == "" {
+		return nil, fmt.Errorf("host of the reencrypt service not set")
+	} else if config.Config.Reencrypt.Port == 0 {
+		return nil, fmt.Errorf("port of the reencrypt service not set")
+	}
```
pontus, line 41: Is this to support some kind of "optional" reencryption usage? Otherwise, check during initialisation instead?

sda-download/api/sda/sda.go
```diff
@@ -29,6 +33,43 @@ func sanitizeString(str string) string {
...
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+	}
+	conn, err := grpc.Dial(address, opts...)
+	if err != nil {
+		log.Errorf("Failed to connect to the reencrypt service, reason: %s", err)
+
+		return nil, err
+	}
+	defer conn.Close()
+	log.Debugf("Connection to the reencrypt service established, conn = %v", conn)
```
pontus, line 55: "I don't know these libraries", but https://pkg.go.dev/google.golang.org/grpc#WithBlock implies we won't know the connection succeeds here, only that it wasn't outright rejected.

sda-download/api/sda/sda.go
pontus, line 57: Configurable? Otherwise 1 second is overly aggressive.

sda-download/api/sda/sda.go
```diff
@@ -17,8 +18,11 @@ import (
 	"github.com/neicnordic/sda-download/api/middleware"
 	"github.com/neicnordic/sda-download/internal/config"
 	"github.com/neicnordic/sda-download/internal/database"
+	re "github.com/neicnordic/sda-download/internal/reencrypt"
```
pontus, line 21: Not wrong, but given that this reference seems to be used two times, maybe we don't need a shorthand? It may be a defect in me, but I also read `re` as regular expressions and get semiconfused each time.

sda-download/api/sda/sda.go
```diff
@@ -268,7 +313,24 @@ func Download(c *gin.Context) {
 
 	switch c.Param("type") {
 	case "encrypted":
-		fileStream = encryptedFileReader
+		// The key provided in the header should be base64 encoded
+		reencKey := c.GetHeader("Client-Public-Key")
+		if strings.HasPrefix(c.GetHeader("User-Agent"), "htsget") {
+			reencKey = c.GetHeader("Server-Public-Key")
+		}
```
pontus, line 320: Can't we just define the order so that `Server-Public-Key` overrides `Client-Public-Key` rather than looking at `User-Agent`?

sda-download/dev_utils/compose.yml
```diff
@@ -105,6 +105,22 @@ services:
       - "8443:8443"
     restart: always
 
+  reencrypt:
+    image: ghcr.io/neicnordic/sensitive-data-archive:v0.3.7
```
pontus, line 109: `:latest` instead (and provide the build context)?

.github/integration/sda/config.yaml
nanjiangshu: That's correct. I will remove it.

sda-download/api/sda/sda.go
nanjiangshu: No, we don't provide "optional" re-encryption usage? I will move this validation to the config. 
One question is when the Host or Port of the reencryption server is not set, but the client public key is provided, should the `sda-download` return the encrypted file or and empty file?

sda-download/api/sda/sda.go
nanjiangshu: I will check it and maybe move the debug info to a better place.

sda-download/api/sda/sda.go
nanjiangshu: I followed the usage from https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/cmd/reencrypt/reencrypt_test.go#L20. But you are right, the short name `re` might not be optimal.

sda-download/dev_utils/compose.yml
nanjiangshu: I tried with certificates but failed many times (as I mentioned in the PR description). I discussed with @dbampalikis and think we may probably skip it for this PR and to get the basic functionality ready first. And then add the TLS feature later on. But if you have some quick suggestions, I can give it another try 😄

sda-download/dev_utils/config.yaml
nanjiangshu: see https://github.com/neicnordic/sensitive-data-archive/pull/761#discussion_r1535712455

sda-download/api/sda/sda.go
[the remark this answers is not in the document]
nanjiangshu: You are right, support for mTLS has not been implement yet. see my comment at https://github.com/neicnordic/sensitive-data-archive/pull/761#discussion_r1535712455

.github/integration/sda/config.yaml
nanjiangshu: Removed in a5baaa6d

sda-download/api/sda/sda.go
pontus: Non-short term, I think also the decrypted download case should use the reencryption service, getting rid of the need to have the sensitive key that can decrypt archive contents in the public facing service (e.g. reencrypt can possibly be hosted in an "internal" cluster with only the grpc port exposed), so I believe we will want to require reencryption service to be used (these things configured)

sda-download/api/sda/sda.go
pontus: Is the connection kept for reuse somehow? I don't understand it as such when I read the code, but that would possibly be a nice feature for future development.

sda-download/internal/config/config.go
```diff
@@ -163,7 +172,7 @@ func NewConfig() (*Map, error) {
 		}
 	}
 	requiredConfVars := []string{
-		"db.host", "db.user", "db.password", "db.database", "c4gh.filepath", "oidc.configuration.url",
+		"db.host", "db.user", "db.password", "db.database", "c4gh.filepath", "c4gh.passphrase", "oidc.configuration.url",
```
pontus, line 175: Is this intentional or is not rebased and an effect of that?

dbampalikis: It was removed from sda-download due to the difference in handling password protected keys between our infra and the rust implementation. Since the re-encryption service allows for having different keys, we are putting back the password as required variable.

sda-download/api/sda/sda.go
dbampalikis: (I left the previous comment, in case anyone read it)
On a second thought, tools like samtools might do the whole procedure in one request:
- Make the request to htsget and get the results
- Make the requests to sda-download and get the relevant parts of the file
- Concatenate everything together
If that's the workflow, then the `Client-Public-Key` needs to be sent with the initial request. This is only an assumption currently, as samtools doesn't support custom header, but I think we should leave it as is for now and possibly update it when we have more information.

sda-download/api/sda/sda.go
jbygdell: We will probably fix this when we merge `download` into `sda`; it will *either* run with reencryption *or* supply decrypted data.

jbygdell, line 41: This shouldn't be here.

sda-download/internal/config/config.go
```diff
@@ -296,6 +308,20 @@ func (c *Map) configArchive() {
 	}
 }
 
+func (c *Map) configReencrypt() {
+	c.Reencrypt.Host = viper.GetString("grpc.host")
+	c.Reencrypt.Port = viper.GetInt("grpc.port")
```
jbygdell, line 313: This should either have a check that it actually is set or set the default port number.
It should also be aware that if certificates are available the port will be different.

sda-download/api/sda/sda.go
pontus: Ok, I'm not thrilled with having key being picked based on `User-Agent` but am fine with leaving as is for now.

sda-download/api/sda/sda.go
```diff
@@ -29,6 +33,43 @@ func sanitizeString(str string) string {
 	return pattern.ReplaceAllString(str, "[identifier]: $1")
 }
 
+func reencryptHeader(oldHeader []byte, reencKey string) ([]byte, error) {
```
jbygdell, line 36: This function is missing a test case

sda-download/api/sda/sda.go
nanjiangshu: I have removed the use of shortname of `re` for reencrypt in the commit 543b2118

sda-download/dev_utils/compose.yml
nanjiangshu: Hi @pontus, I checked but it seems the image `sensitive-data-archive` does not have a tag `latest`. Also, regarding the build context, I think it is better to be addressed when `sda-download` is merged into the `sda` folder.

sda-download/api/sda/sda.go
nanjiangshu: Removed in the commit 8dfc45b

nanjiangshu: Removed in the commit 8dfc45b, and this checking is done in the config.

sda-download/internal/config/config.go
nanjiangshu: Addressed in the commit 8dfc45b. Also, checking whether "grpc.host" is set is done in the parent function `NewConfig()`

sda-download/api/sda/sda.go
nanjiangshu: This has been addressed in the commit 20b3285a, and the default is set to 5 seconds.

sda-download/internal/config/config.go
```diff
@@ -296,6 +311,31 @@ func (c *Map) configArchive() {
...
+	}
+	if viper.IsSet("grpc.timeout") {
+		c.Reencrypt.Timeout = viper.GetInt("grpc.timeout")
+	}
+	if viper.IsSet("grpc.cacert") {
+		c.Reencrypt.CACert = viper.GetString("grpc.cacert")
+		// when certificate is set, the port number should be 50443
+		if c.Reencrypt.Port != 50443 {
+			log.Errorf("CACert is set, but port number is set to %v, which should be 50443", c.Reencrypt.Port)
+		}
```
jbygdell, line 329:
```suggestion
```

The port is not tied to the CA certificate but rather the existence of the client key, see comment below

sda-download/dev_utils/config.yaml
nanjiangshu: The mTLS is enabled in the commit 9f549c32

sda-download/dev_utils/compose.yml
nanjiangshu: mTLS is enabled in the commit 9f549c32

sda-download/internal/config/config.go
```diff
@@ -296,6 +311,31 @@ func (c *Map) configArchive() {
...
+	}
+	if viper.IsSet("grpc.clientcert") {
+		c.Reencrypt.ClientCert = viper.GetString("grpc.clientcert")
+	}
+	if viper.IsSet("grpc.serverkey") {
+		c.Reencrypt.ClientKey = viper.GetString("grpc.clientkey")
+	}
+}
```
jbygdell, line 337:
```suggestion
	}
	
	if c.Reencrypt.ClientCert != "" && c.Reencrypt.ClientKey != "" {
		log.Infoln("client certificates detected, setting grpc port to 50443")
		c.Reencrypt.Port = 50443
	}
}
```

sda-download/api/sda/sda.go
```diff
@@ -29,6 +37,76 @@ func sanitizeString(str string) string {
...
+			return nil, err
+		}
+		clientCreds := credentials.NewTLS(
+			&tls.Config{
+				Certificates: []tls.Certificate{certs},
+				MinVersion:   tls.VersionTLS13,
+				RootCAs:      rootCAs,
+			},
+		)
+		// Use secure gRPC connection with mutual TLS authentication
```
jbygdell, line 74: Tying this to the presence of a CA certificate is wrong, since the CA root might be public and thus doesn't need to be configured.
Also all these checks and reading/parsing of the certificate files should be done as part of the startup procedure *not* when a request comes in.

sda-download/internal/config/config.go
jbygdell, line 337: This needs to be added to the `config_test`

jbygdell (changes requested):

sda-download/api/sda/sda.go
nanjiangshu: Since this function is doing the connection of `download` to `reencrypt`, I think it is better to be tested in the integration test. As to the unit tests of the functionality of reencrypt, they are already fully covered at https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/cmd/reencrypt/reencrypt_test.go.

sda-download/api/sda/sda.go
nanjiangshu: Thank you @jbygdell for pointing out this problem. Considering it will be Easter soon, and to avoid the this PR stays for a long time, I would suggest we solve the TLS problem for `connect reencrypt to sda-download` in a new issue.
What do others think? @MalinAhlberg @aaperis @pahatz @pontus ?

MalinAhlberg: I'm not sure of the exact implications of this problem, but I tend to agree that it could be moved to a new issue. Or, if @jbygdell sits on a quick fix, maybe you could contribute with a commit to this PR? (also ping to @dbampalikis )

sda-download/internal/config/config.go
nanjiangshu: fixed in the commit 44c60ce9

sda-download/api/api.md
```diff
@@ -122,7 +122,9 @@ Headers:
 
 - `Authorization: Bearer <token>` 
 - `Range: bytes=<start>-<end>`  exact positions. Overrides parameter coordinates.
-- `User-agent` used in communication with htsget, to mark who is making the request.
+- `Client-public-key: <key>` used for re-encrypting the header of the file before sending it.
+- `Server-public-key: <key>` used in communication with htsget, for re-encrypting the header of the file.
+- `User-Agent` used in communication with htsget, to mark who is making the request. For example `htsget-search/0.6.6.`
```
jbygdell, line 127: This is utter nonsense, `download` shouldn't care if the request comes from `htsget` or not.
The only thing we actually needs is the public key in the header if all outgoing archive data is sent re-encrypted

ping @dbampalikis for notification.

jbygdell (approved):

pontus (approved):
I also think we can merge this now and iterate improvements.

sda-download/dev_utils/make_certs.sh
```diff
@@ -26,4 +30,5 @@ openssl x509 -req -in ./certs/client.csr -days 1200 -CA ./certs/ca.pem -CAkey ./
 openssl req -config "$(dirname "$0")"/ssl.cnf -new -nodes -newkey rsa:4096 -keyout ./certs/download-key.pem -out ./certs/download.csr -extensions download_cert
 openssl x509 -req -in ./certs/download.csr -days 1200 -CA ./certs/ca.pem -CAkey ./certs/ca-key.pem -set_serial 01 -out ./certs/download.pem -extensions download_cert -extfile "$(dirname "$0")"/ssl.cnf
 
+cp ./certs/ca.pem ./certs/ca-certificates.crt
```
aaperis, line 33: Is this line really needed?

aaperis (approved):
Looks good! I have a small question about `make_certs.sh` but this is not a merge blocker.
I also agree with the suggested improvements and that these can come at a later stage.

sda-download/dev_utils/make_certs.sh
jbygdell: No, most of this file can actually be removed.

sda-download/api/api.md
dbampalikis: I don't think `utter nonsense` is a very nice review comment. It is the way we have decided to implement the different headers in download for this version. The reason behind that is that we don't know how the clients will act, given that there are no clients available currently. The assumption is that the client will send all headers (including the `client-public-key` and the `authorization`) to htsget, and then htsget will move them to the sda-download, which should have a way to figure out which one is relevant. Although I understand that there might be better solutions (for example sda-download to pick the key based on whether the `server` is set), this is the one picked from the people that participated in the planning this.
As discussed in a previous comment from Pontus, that can be fixed in the future, but imho it is not a showstopper

nanjiangshu:
Thank you @jbygdell for fixing the remaining issues. It looks great now.

blankdots (approved):

PR #808, opened 2024-04-10
Use reencrypt service in download

pontus:
Make sda-download use a temporary key and reencrypt for that rather than having the archive key accessible to sda-download directly.

Also contains some followup chart changes and an added a network policy for download.

The use case for this is to support having an external deployment (that can be accessed in some way from the outside) that doesn't have access to the archive key. Since the key is quite sensitive, that is desirable for security reasons, but if we add key rotation, making sure we only need to implement it in only one place makes life easier.

While it aims to be drop-in ready, certain care should be ensured when deploying versions including this change if merged (e.g. to ensure reencrypt service is reachable from download service et.c).

### DIFF
--- a/sda-download/api/sda/sda.go
+++ b/sda-download/api/sda/sda.go
@@ -372,8 +372,18 @@ func Download(c *gin.Context) {
 		fileStream = io.MultiReader(newHr, file)
 
 	default:
-		encryptedFileReader := io.MultiReader(bytes.NewReader(fileDetails.Header), file)
-		c4ghfileStream, err := streaming.NewCrypt4GHReader(encryptedFileReader, *config.Config.App.Crypt4GHKey, nil)
+		// Reencrypt header for use with our temporary key
+		newHeader, err := reencryptHeader(fileDetails.Header, config.Config.App.Crypt4GHPublicKeyB64)
+
+		if err != nil {
+			log.Errorf("Failed to reencrypt the file header, reason: %v", err)
+			c.String(http.StatusInternalServerError, "file re-encryption error")
+
+			return
+		}
+
+		encryptedFileReader := io.MultiReader(bytes.NewReader(newHeader), file)
+		c4ghfileStream, err := streaming.NewCrypt4GHReader(encryptedFileReader, config.Config.App.Crypt4GHPrivateKey, nil)
 		defer c4ghfileStream.Close()
 		if err != nil {
 			log.Errorf("could not prepare file for streaming, %s", err)
--- a/sda-download/api/sda/sda.go
+++ b/sda-download/api/sda/sda.go
@@ -374,7 +374,6 @@ func Download(c *gin.Context) {
 	default:
 		// Reencrypt header for use with our temporary key
 		newHeader, err := reencryptHeader(fileDetails.Header, config.Config.App.Crypt4GHPublicKeyB64)
-
 		if err != nil {
 			log.Errorf("Failed to reencrypt the file header, reason: %v", err)
 			c.String(http.StatusInternalServerError, "file re-encryption error")
--- a/sda-download/internal/config/config.go
+++ b/sda-download/internal/config/config.go
@@ -1,6 +1,8 @@
 package config
 
 import (
+	"bytes"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -54,9 +56,9 @@ type AppConfig struct {
 	// Optional. Defaults to empty
 	ServerKey string
 
-	// Stores the Crypt4GH private key if the two configs above are set
-	// Unconfigurable. Depends on Crypt4GHKeyFile and Crypt4GHPassFile
-	Crypt4GHKey *[32]byte
+	// Stores the Crypt4GH private key used internally
+	Crypt4GHPrivateKey   [32]byte
+	Crypt4GHPublicKeyB64 string
 
 	// Selected middleware for authentication and authorizaton
 	// Optional. Default value is "default" for TokenMiddleware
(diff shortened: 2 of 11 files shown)


sda-download/internal/config/config.go
```diff
@@ -479,24 +482,36 @@ func constructWhitelist(obj []TrustedISS) *jwk.MapWhitelist {
...
+	b64 := bytes.Buffer{}
+
+	encoder := base64.NewEncoder(base64.StdEncoding, &b64)
+	_, err = encoder.Write(pem.Bytes())
+
+	if err != nil {
+		log.Errorf("Error when converting public key to PEM format: %v", err)
+
+		return [32]byte{}, "", err
+	}
```
blankdots, line 514: for me `base64.StdEncoding.EncodeToString(publicKey[:])` worked as well here https://github.com/neicnordic/sda-download/pull/330/files#diff-bd6683a3964980b843bd0f1e474cb2ae7e33a91430a74f5b5770daad2983f0e7R306

blankdots (commented):
maybe i am misunderstanding, however seems similar to what i did in sda-download so all ok, though i would argue:
- `GetC4GKey` can be renamed to something that reflects current status
- `GetC4GKey` can be simplified (see recommendation)

sda-download/api/sda/sda.go
```diff
@@ -372,8 +372,18 @@ func Download(c *gin.Context) {
 		fileStream = io.MultiReader(newHr, file)
 
 	default:
-		encryptedFileReader := io.MultiReader(bytes.NewReader(fileDetails.Header), file)
-		c4ghfileStream, err := streaming.NewCrypt4GHReader(encryptedFileReader, *config.Config.App.Crypt4GHKey, nil)
+		// Reencrypt header for use with our temporary key
+		newHeader, err := reencryptHeader(fileDetails.Header, config.Config.App.Crypt4GHPublicKeyB64)
+
```
jbygdell, line 377:
```suggestion
```

sda-download/internal/config/config.go
```diff
@@ -173,7 +175,7 @@ func NewConfig() (*Map, error) {
 		}
 	}
 	requiredConfVars := []string{
-		"db.host", "db.user", "db.password", "db.database", "c4gh.filepath", "c4gh.passphrase", "oidc.configuration.url",
+		"db.host", "db.user", "db.password", "db.database", "oidc.configuration.url",
```
jbygdell, line 178:
```suggestion
		"db.host", "db.user", "db.password", "db.database", "oidc.configuration.url", "grpc.host"
```

sda-download/internal/config/config.go
```diff
@@ -316,6 +315,10 @@ func (c *Map) configArchive() {
 
 func (c *Map) configReencrypt() error {
 	c.Reencrypt.Host = viper.GetString("grpc.host")
+	if c.Reencrypt.Host == "" {
+		return fmt.Errorf("grpc.host is not set")
+	}
+
```
jbygdell, line 321:
```suggestion
```

Placed as `requiredConfvar` now

sda-download/internal/config/config.go
```diff
@@ -377,7 +380,7 @@ func (c *Map) appConfig() error {
 	}
 
 	var err error
-	c.App.Crypt4GHKey, err = GetC4GHKey()
+	c.App.Crypt4GHPrivateKey, c.App.Crypt4GHPublicKeyB64, err = GetC4GHKey()
```
jbygdell, line 383:
```suggestion
	c.App.Crypt4GHPrivateKey, c.App.Crypt4GHPublicKeyB64, err = GenerateC4GHKey()
```

sda-download/internal/config/config.go
```diff
@@ -479,24 +482,36 @@ func constructWhitelist(obj []TrustedISS) *jwk.MapWhitelist {
 }
 
 // GetC4GHKey reads and decrypts and returns the c4gh key
-func GetC4GHKey() (*[32]byte, error) {
-	log.Info("reading crypt4gh private key")
-	keyPath := viper.GetString("c4gh.filepath")
-	passphrase := viper.GetString("c4gh.passphrase")
+func GetC4GHKey() ([32]byte, string, error) {
```
jbygdell, line 485:
```suggestion
func GenerateC4GHKey() ([32]byte, string, error) {
```

sda-download/internal/config/config.go
```diff
@@ -479,24 +482,36 @@ func constructWhitelist(obj []TrustedISS) *jwk.MapWhitelist {
...
-		return nil, err
+		log.Errorf("Error when generating keys: %v", err)
+
+		return [32]byte{}, "", err
 	}
 
-	key, err := keys.ReadPrivateKey(keyFile, []byte(passphrase))
+	pem := bytes.Buffer{}
+	err = keys.WriteCrypt4GHX25519PublicKey(&pem, public)
+
```
jbygdell, line 498:
```suggestion
```

Or put the entire `keys.WriteCrypt4GHX25519PublicKey(&pem, public)` in the if statement below

sda-download/internal/config/config.go
```diff
@@ -479,24 +482,36 @@ func constructWhitelist(obj []TrustedISS) *jwk.MapWhitelist {
...
+		return [32]byte{}, "", err
 	}
 
-	keyFile.Close()
-	log.Info("crypt4gh private key loaded")
+	b64 := bytes.Buffer{}
+
+	encoder := base64.NewEncoder(base64.StdEncoding, &b64)
+	_, err = encoder.Write(pem.Bytes())
+
```
jbygdell, line 509:
```suggestion
```

sda-download/internal/config/config.go
pontus: Name changed to `GenerateC4GHKey` and easier base64-ification is there now.

blankdots (approved):

jbygdell (approved):

pahatz (approved):

PR #820, opened 2024-04-16
Feat/htsget random access - PLACEHOLDER

dbampalikis:
This pull request is a **placeholder** for the version containing the htsget random access. The image of `sda-download` in the storage-and-interfaces of GDI is based on the code from this PR

**Description**
- Remove the second header for re-encryption (`server-public-key`)
- Remove the `user-agent` check for the requester
- The `content-length` is now including the header size when htsget requests the encrypted file size
- The range header returns an extra character, according to specification
- Includes the PR #703 code, to make random access possible

**How to test**

### DIFF
--- a/.github/workflows/functionality.yml
+++ b/.github/workflows/functionality.yml
@@ -33,7 +33,7 @@ jobs:
 
     strategy:
       matrix:
-        storagetype: [s3, posix, s3notls]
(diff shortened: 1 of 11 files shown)


MalinAhlberg:
Closed in favor of #831 .

PR #826, opened 2024-04-22
Seekable S3 support, second take

pontus:
This is a followup for #703. It's roughly the same code, but seekable (new) functionality is now in different files from the old things for the storage implementation.

### DIFF
--- a/sda-download/api/sda/sda.go
+++ b/sda-download/api/sda/sda.go
@@ -299,8 +299,20 @@ func Download(c *gin.Context) {
 		}
 	}
 
+	wholeFile := true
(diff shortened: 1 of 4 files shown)


MalinAhlberg (approved):
I tested this merged with #820 , which worked well! The merging does require some manual work, though, but I don't think there's anything that could be fixed in this PR.

nanjiangshu (approved):
Looks great! I also tested with the PR https://github.com/neicnordic/sensitive-data-archive/pull/820 and htsget-rust and it works well.

jbygdell (approved):

PR #830, opened 2024-04-23
Add unit test of successful downloads

pontus:
Add unit test for successful downloads as well.

### DIFF
--- a/sda-download/api/sda/sda_test.go
+++ b/sda-download/api/sda/sda_test.go
@@ -2,17 +2,31 @@ package sda
 
 import (
 	"bytes"
+	"crypto/x509"
+	"encoding/base64"
+	"encoding/binary"
 	"errors"
 	"io"
+	"net/http"
 	"net/http/httptest"
+	"net/url"
+	"os"
+	"strconv"
+	"strings"
 	"testing"
 
 	"github.com/gin-gonic/gin"
 	"github.com/stretchr/testify/assert"
+	"golang.org/x/crypto/chacha20poly1305"
+	"google.golang.org/protobuf/proto"
 
+	"github.com/neicnordic/crypt4gh/keys"
+	"github.com/neicnordic/crypt4gh/model/headers"
+	"github.com/neicnordic/crypt4gh/streaming"
 	"github.com/neicnordic/sda-download/api/middleware"
 	"github.com/neicnordic/sda-download/internal/config"
 	"github.com/neicnordic/sda-download/internal/database"
+	"github.com/neicnordic/sda-download/internal/reencrypt"
 	"github.com/neicnordic/sda-download/internal/session"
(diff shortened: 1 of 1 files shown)


sda-download/api/sda/sda_test.go
```diff
@@ -552,3 +566,217 @@ func TestEncrypted_Coords(t *testing.T) {
...
+	w = httptest.NewRecorder()
+	c, _ = gin.CreateTestContext(w)
+	c.Request = &http.Request{Method: "GET", URL: &url.URL{Path: "/s3-encrypted/somepath", RawQuery: "filename=somepath"}}
+	c.Request.Header = http.Header{"Client-Public-Key": []string{config.Config.App.Crypt4GHPublicKeyB64},
+		"Range": []string{"bytes=0-10"}}
+
+	c.Params = make(gin.Params, 1)
+	c.Params[0] = gin.Param{Key: "type", Value: "encrypted"}
+
+	// Test the outcomes of the handler
```
nanjiangshu, line 753: Maybe this comment can be more informative?
So as the rest of comments `// Test the outcomes of the handler`

sda-download/api/sda/sda_test.go
```diff
@@ -552,3 +566,217 @@ func TestEncrypted_Coords(t *testing.T) {
...
+
+	err = binary.Write(w, binary.BigEndian, int32(len(response)))
+	assert.NoError(f.t, err, "Could not write response length")
+
+	_, err = w.Write(response)
+	assert.NoError(f.t, err, "Could not write response")
+	w.Header().Add("Grpc-status", "0")
+}
+
+func TestDownload_Success_Whole(t *testing.T) {
```
nanjiangshu, line 614: It seems this test function contains many scenarios. Not sure if it is better to divide them into separate functions, but if we keep them in one function, maybe change the function name to e.g. `Test_Download_Whole_Or_Range` to better reflect the testing scenarios?

nanjiangshu (commented):
The tests worked well when I tested! I just left a few minor comments.

MalinAhlberg (dismissed):
Looks good! I agree with @nanjiangshu's comments, but it's not stopping.

sda-download/api/sda/sda_test.go
pontus: Yes, I've updated/added comments a bit now, good?

sda-download/api/sda/sda_test.go
pontus: Yes, it turned out to be a bit more setup than I intended from the start so it felt like it made sense to test more cases. I guess one could factor most of that out but there'll still be a fair bit of setup so I think it's best to keep it as is for now, updated the name.

MalinAhlberg (approved):
:+1:

Issue #776, opened 2024-03-28
[download] simplify the header logic for dealing with re-encrypted data


jbygdell:
**Describe the bug**
Download shouldn't care if the request comes from a specific middle ware application like `htsget` or `curl` or `sda-scli`.
Only a single header entry `c4gh-public-key` should exists, whomever is sending the request is irrelevant.

**Expected behavior**
- [ ] No special cases for specific middle ware applications should exists
- [ ] Tests verifying the fix are added

**Additional context**

**Estimation of size**: small

**Estimation of priority**: medium

Issue #701, opened 2024-02-28
[sda-download] Add user-agent header


dbampalikis:
As a developer
I want sda-download to read the User-Agent header
In order to decide which key header should be used for re-encryption

Description
The htsget is currently forwarding all the header from any user request, including the ones containing the public keys of the user and the htsget server. We have defined two different header, namely `client-public-key` and `server-public-key`. Since these headers are passed in every request, we need a way to figure out if the request comes from the user or the htsget server, and re-encrypt accordingly. The idea is to use a `user-agent` header, with a specific value for the htsget server

viklund:
I think htsget-rs is the client of download and should therefore just replace the `client-public-key` with it's own key when making the request to the `sda-download` server. In case htsget need to do some more shenanigans to the client of htsget they should still have access to that key while processing the request.

PR #831, opened 2024-04-23
Feat/htsget seekable

### LINKED ISSUE #696: [sda-download] Implement random access in encrypted files

MalinAhlberg:
This PR closes #696 and neicnordic/sda-download#372.
It includes the changes of #820, but is based on #826 instead of #703.

**Description**

- Remove the second header for re-encryption (server-public-key)
- Remove the user-agent check for the requester
- The content-length is now including the header size when htsget requests the encrypted file size
- The range header returns an extra character, according to specification
- Includes the PR #826 code, to make random access possible

**How to test**
- make sure full and partial files can be downloaded from the `s3` endpoint
- make sure  full and partial files can be downloaded from the `s3-encrypted` endpoint
- make sure it works with htsget (instructions: https://github.com/GenomicDataInfrastructure/starter-kit-htsget?tab=readme-ov-file#accessing-data-with-htsget)

Also see #701 and #776.

### DIFF
--- a/sda-download/api/sda/sda.go
+++ b/sda-download/api/sda/sda.go
@@ -331,13 +331,20 @@ func Download(c *gin.Context) {
 
 	if c.Request.Method == http.MethodHead {
 
+		// Create headers for htsget, containing size of the crypt4gh header
+		reencKey := c.GetHeader("Client-Public-Key")
+		headerSize := bytes.NewReader(fileDetails.Header).Size()
+		// Size of the header in the archive
+		c.Header("Server-Additional-Bytes", fmt.Sprint(headerSize))
+		if reencKey != "" {
+			newHeader, _ := reencryptHeader(fileDetails.Header, reencKey)
+			headerSize = bytes.NewReader(newHeader).Size()
+			// Size of the header if the file is re-encrypted before downloading
+			c.Header("Client-Additional-Bytes", fmt.Sprint(headerSize))
+		}
 		if c.Param("type") == "encrypted" {
-			c.Header("Content-Length", fmt.Sprint(fileDetails.ArchiveSize))
-
-			// set the length of the crypt4gh header for htsget
-			c.Header("Server-Additional-Bytes", fmt.Sprint(bytes.NewReader(fileDetails.Header).Size()))
-			// TODO figure out if client crypt4gh header will have other size
-			// c.Header("Client-Additional-Bytes", ...)
+			// Update the content length to match the encrypted file size
+			c.Header("Content-Length", fmt.Sprint(int(headerSize)+fileDetails.ArchiveSize))
 		}
 
 		return
--- a/sda-download/api/sda/sda.go
+++ b/sda-download/api/sda/sda.go
@@ -272,23 +272,18 @@ func Download(c *gin.Context) {
 		return
 	}
 
-	switch c.Param("type") {
-	case "encrypted":
-		// calculate coordinates
-		start, end, err = calculateEncryptedCoords(start, end, c.GetHeader("Range"), fileDetails)
-		if err != nil {
-			log.Errorf("Byte range coordinates invalid! %v", err)
+	wholeFile := true
+	if start != 0 || end != 0 {
+		wholeFile = false
+	}
 
-			return
-		}
-		if start > 0 {
-			// reading from an offset in encrypted file is not yet supported
-			log.Errorf("Start coordinate for encrypted files not implemented! %v", start)
-			c.String(http.StatusBadRequest, "Start coordinate for encrypted files not implemented!")
+	start, end, err = calculateCoords(start, end, c.GetHeader("Range"), fileDetails, c.Param("type"))
+	if err != nil {
+		log.Errorf("Byte range coordinates invalid! %v", err)
 
-			return
-		}
-	default:
+		return
+	}
+	if c.Param("type") != "encrypted" {
 		// set the content-length for unencrypted files
 		if start == 0 && end == 0 {
 			c.Header("Content-Length", fmt.Sprint(fileDetails.DecryptedSize))
@@ -297,11 +292,6 @@ func Download(c *gin.Context) {
 			togo := end - start
 			c.Header("Content-Length", fmt.Sprint(togo))
 		}
-	}
-
-	wholeFile := true
-	if start != 0 || end != 0 {
-		wholeFile = false
 	}
 
 	// Get archive file handle
@@ -355,6 +345,7 @@ func Download(c *gin.Context) {
 	}
 
 	// Prepare the file for streaming, encrypted or decrypted
+
 	var fileStream io.Reader
 
 	switch c.Param("type") {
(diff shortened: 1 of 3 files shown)


sda-download/api/sda/sda.go
```diff
@@ -272,23 +272,18 @@ func Download(c *gin.Context) {
...
+		wholeFile = false
+	}
 
-			return
-		}
-		if start > 0 {
-			// reading from an offset in encrypted file is not yet supported
-			log.Errorf("Start coordinate for encrypted files not implemented! %v", start)
-			c.String(http.StatusBadRequest, "Start coordinate for encrypted files not implemented!")
+	start, end, err = calculateCoords(start, end, c.GetHeader("Range"), fileDetails, c.Param("type"))
```
pontus, line 280: If you are refactoring coordinate/range handling already, I think a cleaner interface would be passing the context itself, so ideally for me this would be something like

```
start, end, rangeSpecified, err := calculateCoords(c)
```

to pick up either range or startCoordinate and endCoordinate. I strongly think we want this harmonized for all use cases (I definitely want Range supported for both download without s3 and unencrypted s3).

I think it might make somewhat more sense to have a separate function to transform the coordinates from the underlying data stream coordinate system to encrypted coordinates and having header adjustments at the time of need (`Content-length` I assume).

aaperis:
We should open this PR which also covers #820

sda-download/api/sda/sda.go
MalinAhlberg: I do agree, it would be very nice to refactor this. But in order to get this merged before the deadline, I wonder if it would not make sense to save all the refactoring for later. Considering the risk of introducing new errors and the time we need to spend testing after refactoring...
Maybe create a new issue for this and list all ideas there?

sda-download/api/sda/sda.go
```diff
@@ -272,23 +272,18 @@ func Download(c *gin.Context) {
...
-	switch c.Param("type") {
-	case "encrypted":
-		// calculate coordinates
-		start, end, err = calculateEncryptedCoords(start, end, c.GetHeader("Range"), fileDetails)
-		if err != nil {
-			log.Errorf("Byte range coordinates invalid! %v", err)
+	wholeFile := true
+	if start != 0 || end != 0 {
+		wholeFile = false
+	}
```
nanjiangshu, line 278: is the condition 
```golang
start == 0 && end >= fileDetails.DecryptedSize
```
also satisfy the condition `wholeFile`? 
so maybe we can change the code to 
```suggestion
        wholeFile := false
        if start == 0 && (end == 0 || end >= fileDetails.DecryptedSize) {
              wholeFile = true
        }
```

sda-download/api/sda/sda.go
```diff
@@ -450,6 +449,24 @@ func Download(c *gin.Context) {
 	}
 }
 
+var seekStart = func(fileStream io.ReadSeeker, start, end int64) (int64, int64, error) {
```
nanjiangshu, line 452: the function `seekStart` does not reflect what is doing very well, maybe change to e.g. `adjustSeekPosition`?

MalinAhlberg: good point, fixed in 468848378f87ce5ff92692864ef622a6f7113184

sda-download/api/sda/sda.go
nanjiangshu: make sense to me.

nanjiangshu (approved):

sda-download/api/sda/sda.go
jbygdell, line 278: This entire wholeFile thingy could be removed if the check inside the switch case looks something like this:
```
		if start == 0 && end == 0 {
			file, err := Backend.NewFileReader(fileDetails.ArchivePath)
			if err != nil {
				log.Errorf("could not find archive file %s, %s", fileDetails.ArchivePath, err)
				c.String(http.StatusInternalServerError, "archive error")
		
				return
			}

			fileStream = io.MultiReader(newHr, file)
		} else {
			file, err := Backend.NewFileReadSeeker(fileDetails.ArchivePath)
```

sda-download/api/sda/sda.go
```diff
@@ -337,35 +327,39 @@ func Download(c *gin.Context) {
...
 		log.Debugf("Got to setting the headers: %s", c.GetHeader("client-public-key"))
 		c.Header("Client-Public-Key", c.GetHeader("Client-Public-Key"))
-		c.Header("Server-Public-Key", c.GetHeader("Server-Public-Key"))
 	}
 
 	if c.Request.Method == http.MethodHead {
 
+		// Create headers for htsget, containing size of the crypt4gh header
+		reencKey := c.GetHeader("Client-Public-Key")
+		headerSize := bytes.NewReader(fileDetails.Header).Size()
```
pontus, line 336:
```suggestion
		headerSize := len(fileDetails.Header)
```

sda-download/api/sda/sda.go
```diff
@@ -337,35 +327,39 @@ func Download(c *gin.Context) {
...
+		// Size of the header in the archive
+		c.Header("Server-Additional-Bytes", fmt.Sprint(headerSize))
+		if reencKey != "" {
```
pontus, line 339: Do we really care about the client key here? For now I think we can assume the same kinds of public keys, meaning the same header size, so "fixed" number.

pontus (approved):

sda-download/api/sda/sda.go
MalinAhlberg: Yes, but the end position might have been altered in `calculateCoords`. So one needs to make sure to check the original positions for the `wholeFile` check, if doing this kind of refactoring.

sda-download/api/sda/sda.go
MalinAhlberg: Both `server-additional-bytes` and `client-additional-bytes` are part of the spec from crypt4gh, although optional. I thought it would be good to have them set, so that we don't run into problems later, if someone would have different types of keys. But I might be overdoing things... @dbampalikis ?

pontus: Sorry if this was unclear, I didn't mean we shouldn't send those but rather that the numbers we send won't depend on the public key used in the request, so there's no need to check for key here (they will likely depend on whatever we stick a data-edit-list in there, once we support that, so they will depend on whatever coordinates is requested but that can be dealt with later).

Possibly, one might move up the check for a valid key and bail out earlier if none is supplied if encrypted transfers are required, but I've already approved and am fine with merging this as is.

sda-download/api/sda/sda.go
MalinAhlberg: to be picked up in #834 .

sda-download/api/sda/sda.go
MalinAhlberg: to be picked up in #834.

sda-download/api/sda/sda.go
MalinAhlberg: to be picked up in #834 .

sda-download/api/sda/sda.go
MalinAhlberg: to be picked up in #834 .

PR #842, opened 2024-05-06
Update GO version to 1.22

jbygdell:
Updates the go version used, this was brought up due to a pending package update #841

### DIFF
--- a/sda-download/go.mod
+++ b/sda-download/go.mod
@@ -1,6 +1,6 @@
 module github.com/neicnordic/sda-download
 
-go 1.21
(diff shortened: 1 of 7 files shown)


sda-download/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/neicnordic/sda-download
 
-go 1.21
+go 1.22.2
```
blankdots, line 3: i think we don\t need the patch version ...
```suggestion
go 1.22
```

aaperis (approved):

PR #856, opened 2024-05-08
Do some cleanup of obsolete TLS configuration since moving to go 1.22

pontus:
Go with recent version maintains good defaults, don't override those without good reasons (and if we to my surprise have such, we should document them better). Remove some odd things about CA pools.

1.22 will default to sensible minimum TLS 1.2 unless explicity told otherwise, but since the gosec linter complains if we remove the same configuration from` tls.Config`, they are left in here.

### DIFF
--- a/sda/cmd/s3inbox/healthchecks.go
+++ b/sda/cmd/s3inbox/healthchecks.go
@@ -66,7 +66,7 @@ func (h *HealthCheck) RunHealthChecks() {
 }
 
 func (h *HealthCheck) httpsGetCheck(url string, timeout time.Duration) healthcheck.Check {
-	cfg := &tls.Config{MinVersion: tls.VersionTLS12}
+	cfg := &tls.Config{}
 	cfg.RootCAs = h.tlsConfig.RootCAs
 	tr := &http.Transport{TLSClientConfig: cfg}
 	client := http.Client{
--- a/sda/cmd/s3inbox/healthchecks.go
+++ b/sda/cmd/s3inbox/healthchecks.go
@@ -66,7 +66,9 @@ func (h *HealthCheck) RunHealthChecks() {
 }
 
 func (h *HealthCheck) httpsGetCheck(url string, timeout time.Duration) healthcheck.Check {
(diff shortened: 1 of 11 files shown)


sda/cmd/s3inbox/healthchecks.go
```diff
@@ -66,7 +66,9 @@ func (h *HealthCheck) RunHealthChecks() {
 }
 
 func (h *HealthCheck) httpsGetCheck(url string, timeout time.Duration) healthcheck.Check {
-	cfg := &tls.Config{MinVersion: tls.VersionTLS12}
+	cfg := &tls.Config{
+		MinVersion: tls.VersionTLS12,
+	}
```
jbygdell, line 71:
```suggestion
	cfg := &tls.Config{MinVersion: tls.VersionTLS12}
```

We should be consistent across all files, either a one liner or verbose writing.

sda/internal/config/config.go
```diff
@@ -1087,15 +1086,15 @@ func TLSConfigBroker(c *Config) (*tls.Config, error) {
...
-	// Enforce TLS1.2 or higher
-	cfg.MinVersion = 2
-
 	// Read system CAs
-	var systemCAs, _ = x509.SystemCertPool()
-	if reflect.DeepEqual(systemCAs, x509.NewCertPool()) {
-		log.Debug("creating new CApool")
-		systemCAs = x509.NewCertPool()
+	systemCAs, err := x509.SystemCertPool()
+
```
jbygdell, line 1091:
```suggestion
```

jbygdell (dismissed):
Some minor inconsistencies that's all

sda/cmd/s3inbox/healthchecks.go
pontus: Fixed.

sda/internal/config/config.go
pontus: Fixed.

jbygdell (approved):

aaperis (approved):

PR #882, opened 2024-05-28
Fix/ssl config in tests

aaperis:
This should fix errors in  #877 #878 #879,

**Description**
This PR fixes an issue with ssl.conf that seems to appear after [a bug was fixed](https://github.com/openssl/openssl/issues/22966#issuecomment-1858396738) for openssl. The issue leads to an error in `certfixer` which makes integration tests fail.

The `authorityKeyIdentifier` resource is not needed to be specified here and removing it solves the issue.

**How to test**


jbygdell:
Remove the dependabot commit form this PR

jbygdell (approved):

pontus (approved):

PR #887, opened 2024-05-30
feat: configurable allow uncrypted download

### LINKED ISSUE #868: [download] Make serving unencrypted files configurable
Description:
sda-download at the moment is always serving unencrypted files through its dedicated endpoints.

A/C:
- Have a configuration option to disable serving unencrypted files
- Asking for unencrypted data should return a 400 (or some more appropriate 40x error)
- Update Helm chart for the necessary change

nanjiangshu:
This PR fixes #868 

**Description**
By default, downloading unencrypted files is disabled and will return a 400 (BadRequest) error. This is however configurable. When `serveUnencryptedData: true` is set in the config file under the item `app`, download unencrypted files will be  allowed.  

**How to test**
Use [this](https://github.com/GenomicDataInfrastructure/starter-kit-storage-and-interfaces/tree/fix/no-s3seekable) branch from the `starter-kit-storage-and-interfaces` to start the stack. Then obtain the token and key by 
```
token=$(curl -s -k https://localhost:8080/tokens | jq -r '.[0]')
sda-cli createKey c4gh
pubkey=$(base64 -w0 c4gh.pub.pem)
```

After that, 

```
# this should return 400
curl -o /dev/null -s -w "%{http_code}\n" -H "Authorization: Bearer $token" -H -k http://localhost:8443/files/FILE0000001 

# this should return 400
curl -o /dev/null -s -w "%{http_code}\n" -H "Authorization: Bearer $token" -H -k http://localhost:8443/s3/DATASET0001//htsnexus_test_NA12878.bam  # unenrypted

# this should return 200
curl -o /dev/null -s -w "%{http_code}\n" -H "Client-Public-Key: $pubkey" -H "Authorization: Bearer $token" http://localhost:8443/s3-encrypted/DATASET0001/htsnexus_test_NA12878.bam 
```

Try to set `allowUnencryptedDownload: true` in the config file and all of the above three commands should return `200`


sda-download/api/sda/sda.go
```diff
@@ -207,6 +207,11 @@ func Files(c *gin.Context) {
 
 // Download serves file contents as bytes
 func Download(c *gin.Context) {
+	if c.Param("type") != "encrypted" && !config.Config.App.AllowUnencryptedDownload {
+		c.String(http.StatusForbidden, "request to unencrypted file not allowed")
+
+		return
+	}
```
jbygdell, line 214:
```suggestion
	if c.GetHeader("Client-Public-Key") == "" && !config.Config.App.AllowUnencryptedDownload {
		c.String(http.StatusBadRequest, "no public key in request")

		return
	}
```

Target the header instead, it will make this more future proof.

sda-download/internal/config/config.go
```diff
@@ -63,6 +63,8 @@ type AppConfig struct {
 	// Selected middleware for authentication and authorizaton
 	// Optional. Default value is "default" for TokenMiddleware
 	Middleware string
+
+	AllowUnencryptedDownload bool
```
jbygdell, line 67: This should have better name so we don't have to change it in the near future.
Something like `ServerUnencryptedData`, then it will make sense even when we get to the point where the app only can serve encrypted *or* unencrypted data.

jbygdell (commented):
Looks good overall, some comments about future proofing the changes going in.

sda-download/api/sda/sda.go
nanjiangshu: I don't know exactly how we will refactor in the future. Will we remove `c.Param("type")` ?
For the moment, I think the condition `c.Param("type") != "encrypted"` is not suitable since the user can easily bypass the restriction of `config.Config.App.AllowUnencryptedDownload` by providing `Client-Public-Key` in the header when download unencrypted file, e.g. 

```
curl -H "Client-Public-Key: $pubkey" -H "Authorization: Bearer $token" -k http://localhost:8443/s3/DATASET0001//htsnexus_test_NA12878.bam --output t2.bam 
```

MalinAhlberg: This would require some more refactoring, though, right? The public-key header is only checked for the `s3-encrypted` endpoint.
```sh
$ curl -o /dev/null -s -w "%{http_code}\n" -H "Authorization: Bearer $token" -H -k http://localhost:8443/s3/DATASET0001//htsnexus_test_NA12878.bam 
403
$ curl -o test.bam -s -w "%{http_code}\n" -H "Authorization: Bearer $token" -H -k http://localhost:8443/s3/DATASET0001//htsnexus_test_NA12878.bam  -H 'Client-Public-Key: abc' # test.bam is downloaded, unencrypted
200 
```

sda-download/internal/config/config.go
nanjiangshu: Should I use the name `ServeUnencryptedData` instead of `ServerUnencryptedData`?

pontus:
What are the required use cases here? With bigpicture perspective, I don't see that we need have the same server serve both encrypted.

If that goes for all your use cases as well,  I would feel better about just looking at this setting rather than what's now set as `c.Param("type") == "encrypted"`.

sda-download/internal/config/config.go
jbygdell: Yes, the `r` is a typo, of course it should say `ServeUnencryptedData`

nanjiangshu: addressed in the commit 749fb894

sda-download/api/sda/sda.go
nanjiangshu: As Malin @MalinAhlberg pointed out as well, I think for the moment it is better to use the condition `c.Param("type") != "encrypted"` instead of checking `Client-Public-Key` in the header, and leave that for the future refactoring. I will change the return code from `http.StatusForbidden` to `http.StatusBadRequest`. Does it makes sense @jbygdell ?

charts/sda-svc/values.yaml
```diff
@@ -232,6 +232,7 @@ global:
       iss:
         - iss: "https://login.elixir-czech.org/oidc"
           jku: "https://login.elixir-czech.org/oidc/jwk"
+    serveUnencryptedData: false
```
jbygdell, line 235: This option needs to be documented in the readme file as well.

nanjiangshu: Thanks for pointing this out. Addressed in the commit c6bf828e

aaperis (approved):
Looks good! This should do for now, until we refactor the endpoint as others discussed.

sda-download/api/sda/sda_test.go
```diff
@@ -777,6 +847,26 @@ func TestDownload_Whole_Range_Encrypted(t *testing.T) {
 	assert.Equal(t, 200, response.StatusCode, "Unexpected status code from download")
 	assert.Equal(t, []byte("crypt4gh"), body[:8], "Unexpected body from download")
 
+	// Test encrypted download, should work even when AllowedUnencryptedDownload is false
```
MalinAhlberg, line 850: I don't really see why this is part of the `TestDownload_Whole_Range_Encrypted` function?

MalinAhlberg (approved):
Looks good!
I think it would be nice to make a new test instead of making `TestDownload_Whole_Range_Encrypted` even longer, but it won't stop my approval.

sda-download/api/sda/sda_test.go
nanjiangshu: @MalinAhlberg I agree it's not the best place to put it here. But for the first, I wanted to avoid duplicating the code that prepare the data, and for the second this config setting will anyway be refactored later on. Do you think it will be better that I just remove this test?

MalinAhlberg: I see. I would vote for keeping it as is, and plan for a general refactoring of the tests later. So feel free to resolve this conversation.

kostas-kou (commented):
Looks good, just fix the failure in the one of the tests and then I will approve

aaperis:
> Looks good, just fix the failure in the one of the tests and then I will approve

This should hopefully be resolved once the PR is rebased on main.

nanjiangshu:
> > Looks good, just fix the failure in the one of the tests and then I will approve
> 
> This should hopefully be resolved once the PR is rebased on main.

Very good guess @aaperis ! Tests are passed after rebasing to the main branch @kostas-kou

jbygdell (approved):

PR #904, opened 2024-06-17
Add nginx-affinity label to ingress templates in the SDA-SVC chart.

### LINKED ISSUE #431: [sda-auth] Download configuration file fails the first time
The button for downloading the configuration file fails the first time the user tries to click it. Going through the login process again, solves the problem, however, it's not optimal.

jbygdell:
This PR closes #431 

**Description**
This PR resolves an issue when the number of replicas for the Auth pod is greater than 1.
By using nginx specific labels a user can be directed to the same pod for each request.

The label is added to all ingress templates, so we don't get the same problem with DOA, Download or the Inbox.

* Chart version is updated.

### DIFF
--- a/charts/sda-svc/Chart.yaml
+++ b/charts/sda-svc/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: sda-svc
-version: 0.26.4
+version: 0.26.5
 appVersion: v0.3.77
 kubeVersion: '>= 1.26.0'
 description: Components for Sensitive Data Archive (SDA) installation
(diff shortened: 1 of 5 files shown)


pontus (approved):

aaperis (approved):
Works well!

aaperis (approved):

pontus (approved):

aaperis (approved):

nanjiangshu (approved):
Works well when I tried with concurrent login. Nice fix!

PR #939, opened 2024-07-15
[Acrions] Retry when downloading external components

### LINKED ISSUE #645: Dependencies for integration tests
**Describe the bug**
Some times the integration tests fails because downloading/installation of required apps fails (s3cmd, crypt4gh). This is about the tests running as github actions.

**Additional context**
- Ensure all apps get installed/downloaded before continuing with the test

**Estimation of size**: small

**Estimation of priority**: low

jbygdell:
This PR closes #645 .

**Description**
Add a retry to all curl commands when we download external components to our tests so that they don't fail because there where a network hiccup with the runner.

> --retry
> 
> If a transient error is returned when curl tries to perform a transfer, it will retry this number of times before giving up. Setting the number to 0 makes curl do no retries (which is the default). Transient error means either: a timeout, an FTP 4xx response code or an HTTP 5xx response code.
> 
> When curl is about to retry a transfer, it will first wait one second and then for all forthcoming retries it will double the waiting time until it reaches 10 minutes which then will be the delay between the rest of the retries.

**How to test**


kostas-kou (approved):
Looks good

pahatz (approved):

PR #949, opened 2024-07-19
[download] ignore trailing slashes in ISS and JKU URLs

### LINKED ISSUE #915: [download] Iss and jku matching should be done as URL matching
**Describe the bug**
Currently the matching is done via string matching:
```
if value.ISS == issuerValue && value.JKU == jkuValue
```
 which causes problems if for example a trailing `/` is neglected somewhere. 

**Steps to reproduce**

Remove or add  a `/` in a iss or jku URL in the iss.json file of a local download instance and try again to access a dataset.

**Expected behavior**
- [ ] Matching is done on the URL level 

**Estimation of size**: small

jbygdell:
This PR closes #915 .

**Description**
This PR adds a small function that cleans URLs before comparison so that `http://example.com/` and `http://example.com` is identical. The current string matching will evaluate these as false.

**How to test**

### DIFF
--- a/sda-download/pkg/auth/auth.go
+++ b/sda-download/pkg/auth/auth.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"errors"
 	"net/http"
+	"net/url"
+	"path"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pahatz (approved):
Looks good!

kostas-kou (approved):
I had the impression that this bug was fixed in another PR

PR #955, opened 2024-07-24
First stab at creating admin endpoints in the API

jbygdell:
**Background**  
For standalone setups, like the one used in GDI, there are a need to be able to take care of the administrative tasks without the need to have direct access to the MQ or the PostgreSQL database.

This PR begins the process of simplifying these administrative tasks.
The first goal is to be able to handle the ingestion process i.e. trigger the ingestion of a file, assign a stable ID to the file and in the end create a dataset from a set of stable IDs.

**Description**
First take on the admin API
* Restricts endpoints to a set of whitelisted users
* Triggers ingestion of a file
* Assign accession ID to a file
* Create a dataset from a set of accession IDs
* Release the dataset

**How to test**

### DIFF
--- a/sda/cmd/api/api.go
+++ b/sda/cmd/api/api.go
@@ -75,6 +75,7 @@ func setup(config *config.Config) *http.Server {
 	r.GET("/files", getFiles)
 	// admin endpoints below here
 	r.POST("/file/ingest", isAdmin(), ingestFile) // start ingestion of a file
+	r.POST("/file/accession", isAdmin(), setAccession) // assign accession ID to a file
 
 	cfg := &tls.Config{MinVersion: tls.VersionTLS12}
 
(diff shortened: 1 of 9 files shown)


sda/cmd/api/api.md
```diff
@@ -1,19 +1,66 @@
...
+    ```bash
+    $ curl 'https://server/files' -H "Authorization: Bearer $token"
+  [{"inboxPath":"requester_demo.org/data/file1.c4gh","fileStatus":"uploaded","createAt":"2023-11-13T10:12:43.144242Z"}] 
+  ```
+
+  If the `token` is invalid, 401 is returned.
+
+### Admin endpoints
+
+Admin endpoints are only available to a set of whitelisted users specified in the application config.
```
MalinAhlberg, line 26: For all of the below endpoints, it would be nice if it was mentioned what error messages/codes you can get.

sda/cmd/api/api_test.go
```diff
@@ -513,3 +535,561 @@ func (suite *TestSuite) TestAPIGetFiles() {
...
+	}
+	ingestMsg, _ := json.Marshal(ingest{User: "", FilePath: filePath})
+	// Mock request and response holders
+	w := httptest.NewRecorder()
+	r := httptest.NewRequest(http.MethodPost, "/file/ingest", bytes.NewBuffer(ingestMsg))
+
+	_, router := gin.CreateTestContext(w)
+	router.POST("/file/ingest", ingestFile)
+
+	// admin user should be allowed
```
MalinAhlberg, line 679: This comment appears in multiple places but seems a bit off

sda/cmd/api/api_test.go
```diff
@@ -374,6 +360,42 @@ func (suite *TestSuite) SetupTest() {
...
+		Exchange: "sda",
+		Vhost:    "/sda",
+	}
+	Conf.API.MQ, err = broker.NewMQ(Conf.Broker)
+	assert.NoError(suite.T(), err)
+
+}
+
+func (suite *TestSuite) SetupTest() {
+	log.Print("Setup DB for my test")
```
MalinAhlberg, line 377: Do we really want this `Print`?

MalinAhlberg (commented):
Looks good! Also very nice tests added :ok_hand: .
I left some minor comments, I'm ready to approve when the comments in `api_test.go` have been fixed.

sda/cmd/api/api_test.go
jbygdell: That block already existed in another place, but yes the print should not be there.

sda/cmd/api/api_test.go
jbygdell: fixed

sda/cmd/api/api.md
jbygdell: fixed

sda/cmd/api/api.go
```diff
@@ -174,3 +183,178 @@ func getFiles(c *gin.Context) {
...
+func isAdmin() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		token, err := auth.Authenticate(c.Request)
+		if err != nil {
+			log.Debugln("bad token")
+			c.AbortWithStatusJSON(http.StatusUnauthorized, gin.H{"error": err.Error()})
+
+			return
+		}
+		if !slices.Contains(Conf.API.Admins, token.Subject()) {
```
pontus, line 196: I think and identity is the pair of Subject, Issuer that both need to be checked here? (Issuer to key id makes sense to check when getting the token.)

jbygdell: The token validation, (issuer and public key) is done in the `auth.Authenticate` call.

pontus: I don't recall now, but it makes sense to me to support multiple sources (issuers), in which case we'd should check that as part of the identity here as well.

jbygdell: Put it in the wish list for the future, this is by no means the end of the work needed to get the API to a functional state.

aaperis (approved):
Looks really good :-)

PR #956, opened 2024-07-24
API admin part 2

jbygdell:
**Background**  
As a repository admin I need a simple way of seeing which users have ongoing submissions and a way to list the uploaded files for each user.

**Description**
Continuation on PR #955, this PR adds the following endpoints:
* listing all users with ongoing submissions
* listing all files for belonging to a user that are not part of a dataset

**How to test**

### DIFF
--- a/sda/cmd/api/api.md
+++ b/sda/cmd/api/api.md
@@ -111,7 +111,7 @@ Admin endpoints are only available to a set of whitelisted users specified in th
     Example:
 
     ```bash
-    curl -H "Authorization: Bearer $token" -X GET  https://HOSTNAME/users
+    curl -H "Authorization: Bearer $token" -X GET  https://HOSTNAME/users/submitter@example.org/files
     ```
 
 - Error codes
--- a/sda/cmd/api/api.md
+++ b/sda/cmd/api/api.md
@@ -103,3 +103,18 @@ Admin endpoints are only available to a set of whitelisted users specified in th
   - `200` Query execute ok.
   - `401` User is not in the list of admins.
   - `500` Internal error due to DB failure.
+
+- `/users/:username/files`
+  - accepts `GET` requests`
+  - Returns all files for a user with active uploads as a JSON array
+
+    Example:
+
+    ```bash
+    curl -H "Authorization: Bearer $token" -X GET  https://HOSTNAME/users
+    ```
+
+- Error codes
+  - `200` Query execute ok.
+  - `401` User is not in the list of admins.
+  - `500` Internal error due to DB failure.
--- a/sda/cmd/api/api.md
+++ b/sda/cmd/api/api.md
@@ -88,3 +88,18 @@ Admin endpoints are only available to a set of whitelisted users specified in th
     ```bash
     curl -H "Authorization: Bearer $token" -X POST  https://HOSTNAME/dataset/release/my-dataset-01
     ```
+
+- `/users`
+  - accepts `GET` requests`
+  - Returns all users with active uploads as a JSON array
+
+    Example:
+
+    ```bash
+    curl -H "Authorization: Bearer $token" -X GET  https://HOSTNAME/users
+    ```
+
+- Error codes
+  - `200` Query execute ok.
+  - `401` User is not in the list of admins.
+  - `500` Internal error due to DB failure.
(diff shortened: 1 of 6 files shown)


sda/cmd/api/api.md
```diff
@@ -88,3 +88,33 @@ Admin endpoints are only available to a set of whitelisted users specified in th
...
+
+    Example:
+
+    ```bash
+    curl -H "Authorization: Bearer $token" -X GET  https://HOSTNAME/users
+    ```
+
+- Error codes
+  - `200` Query execute ok.
+  - `401` User is not in the list of admins.
```
MalinAhlberg, line 119: There's a potential for confusion here; that `User` could be interpreted as the user in `/user/:username/file` (instead of the requester)

sda/internal/database/db_functions.go
```diff
@@ -635,14 +635,9 @@ func (dbs *SDAdb) getUserFiles(userID string) ([]*SubmissionFileInfo, error) {
...
 	// select all files of the user, each one annotated with its latest event
-	const query = "SELECT f.submission_file_path, e.event, f.created_at " +
-		"FROM sda.files f " +
-		"LEFT JOIN ( " +
-		"SELECT DISTINCT ON (file_id) file_id, started_at, event " +
-		"FROM sda.file_event_log " +
-		"ORDER BY file_id, started_at DESC" +
-		") e ON f.id = e.file_id " +
-		"WHERE f.submission_user = $1; "
+	const query = "SELECT f.submission_file_path, e.event, f.created_at FROM sda.files f " +
```
MalinAhlberg, line 638: This function (`getUserFiles`) is also used in the `/files` enpoint, right? Do we really want these changes to that request too?

sda/internal/database/db_test.go
```diff
@@ -116,6 +116,10 @@ func (suite *DatabaseTests) SetupTest() {
 		ClientKey:  "",
 	}
 
+	db, err := NewSDAdb(suite.dbConf)
+	assert.Nil(suite.T(), err, "got %v when creating new connection", err)
+	_, err = db.DB.Exec("TRUNCATE sda.files CASCADE")
+	assert.NoError(suite.T(), err)
```
MalinAhlberg, line 122: why are we adding this now..?

sda/cmd/api/api.md
jbygdell: so `:username` should be `:submitter` to make things more clear.

sda/internal/database/db_functions.go
jbygdell: Yes, because files that are already part of a dataset should be of no interest.

sda/internal/database/db_test.go
jbygdell: This is to prevent false errors when all tests are run together vs when a single test is run.

sda/cmd/api/api.md
jbygdell: Reworded the readme

sda/internal/database/db_functions.go
```diff
@@ -635,14 +635,9 @@ func (dbs *SDAdb) getUserFiles(userID string) ([]*SubmissionFileInfo, error) {
...
+		"LEFT JOIN (SELECT DISTINCT ON (file_id) file_id, started_at, event FROM sda.file_event_log ORDER BY file_id, started_at DESC) e ON f.id = e.file_id WHERE f.submission_user = $1 " +
+		"AND f.id NOT IN (SELECT f.id FROM sda.files f RIGHT JOIN sda.file_dataset d ON f.id = d.file_id); "
```
aaperis, line 640: the exclusion of files already existing in a dataset should be mentioned in the function description of the comments above and in the `api.md`

sda/internal/database/db_functions.go
```diff
@@ -696,3 +691,30 @@ func (dbs *SDAdb) getCorrID(user, path string) (string, error) {
 
 	return corrID, nil
 }
+
+func (dbs *SDAdb) ListActiveUsers() ([]string, error) {
```
aaperis, line 695:
```suggestion
// ListActiveUsers returns a list with all users who have submitted files that are not mapped to a dataset
func (dbs *SDAdb) ListActiveUsers() ([]string, error) {
```

It would be nice to add a comment description for `getCorrID` and other functions if they're missing it.

aaperis (approved):
Looks good! minor suggestions

sda/internal/database/db_functions.go
jbygdell: There's a third PR in the making (not the helm chart) that will sort out all of that.

PR #967, opened 2024-07-31
update readme for the download api

### LINKED ISSUE #928: Fix the api README
The documentation of https://github.com/neicnordic/sensitive-data-archive/blob/main/sda-download/api/api.md#s3-requests-for-htsget seems to be outdated.
- [x] Check the whole README and update everything that is outdated.
- [x] Update section:
```
htsget:

    /s3/<datasetid>/<fileid>
    /s3-encrypted/<datasetid>/<fileid>
```
- [x] Update the headers under https://github.com/neicnordic/sensitive-data-archive/blob/main/sda-download/api/api.md#s3-requests-for-htsget

MalinAhlberg:
This PR closes #928 

**Description**
- removed info about headers that are no longer used
- rephrased info about the `/s3` endpoints
- clarifications
- added info on that the unencrypted file download may be disallowed since #887.

**How to test**
Read :).

### DIFF
--- a/sda-download/api/api.md
+++ b/sda-download/api/api.md
@@ -27,8 +27,8 @@ The client can establish a session to bypass time-costly visa validations for fu
 
 **[File download requests, for htsget](#file-download-requests)**
 
-- `/s3/<datasetid>/<fileid>`
-- `/s3-encrypted/<datasetid>/<fileid>`
+- `/s3/*datasetid/*filepath`
+- `/s3-encrypted/*datasetid/*filepath`
 
 ### Data out API
 #### Datasets
@@ -54,18 +54,10 @@ GET /metadata/datasets/{datasetName}/files
 The `?scheme=` query parameter is optional. When a dataset name contains a scheme, such as `https://`, it may sometimes encounter issues with reverse proxies.
 This can be solved by separating the scheme from the dataset name and suppling it as a query parameter.
 ```
-dataset := strings.Split("https://doi.org/abc/123", "://")
-len(dataset) // 2 -> scheme can be used
-dataset[0] // "https"
-dataset[1] // "doi.org/abc/123
-
-dataset := strings.Split("EGAD1000", "://")
-len(dataset) // 1 -> no scheme
-dataset[0] // "EGAD1000"
-```
-```
 GET /metadata/datasets/{datasetName}/files?scheme=https
 ```
+For example, given a dataset name `https://doi.org/abc/123`, one can do `GET /metadata/datasets/doi.org/abc/123/files?scheme=https`.
+ 
 ##### Response
 ```
 [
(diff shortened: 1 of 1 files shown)


sda-download/api/api.md
```diff
@@ -19,13 +19,13 @@ The client can establish a session to bypass time-costly visa validations for fu
...
+**[Data out API](#data-out-api)**:
 
 - `/metadata/datasets`
 - `/metadata/datasets/*dataset`
 - `/files/:fileid`
 
-**htsget**:
+**[File download requests, for htsget](#file-download-requests)**
 
 - `/s3/<datasetid>/<fileid>`
```
nanjiangshu, line 30: To not confused with the `fileid` in the `/files` endpoint, I think it is better to use `/s3/<datasetid>/<filepath>`. Similar to the `/s3-encrypted` endpoint.

sda-download/api/api.md
```diff
@@ -51,8 +51,8 @@ The files contained in a dataset are listed using the `datasetName` obtained fro
 GET /metadata/datasets/{datasetName}/files
 ```
 **Scheme Parameter**
-The `?scheme=` query parameter is optional. When a dataset contains a scheme, it may sometimes encounter issues with reverse proxies.
-The scheme can be separated from the dataset name and supplied in a query parameter.
+The `?scheme=` query parameter is optional. When a dataset name contains a scheme, such as `https://`, it may sometimes encounter issues with reverse proxies.
+This can be solved by separating the scheme from the dataset name and suppling it as a query parameter.
 ```
```
nanjiangshu, line 56: It's a bit confusing with the inclusion of the source code here. I think it is better either remove it from the README, i.e. only including the instruction for how to use the `?scheme` API.
If we really want to include how it works, maybe add it as annotating text by 
> Below is a demonstration of how the dataset name with a schema is parsed in the code.
> ```dataset := ...```

sda-download/api/api.md
MalinAhlberg: Good point. How about this (using the same syntax as for the data out API endpoints):
```
- `/s3/*datasetid/*filepath`
- `/s3-encrypted/*datasetid/*filepath`
```

sda-download/api/api.md
MalinAhlberg: I agree. I'm tempted to replace it with an example, as below. What do you think?

> This can be solved by separating the scheme from the dataset name and suppling it as a query parameter.
> ```
> GET /metadata/datasets/{datasetName}/files?scheme=https
> ```
> For example, given a dataset name `https://doi.org/abc/123`, one can do `GET /metadata/datasets/doi.org/abc/123/files?scheme=https`.
>

MalinAhlberg:
> Looks great @MalinAhlberg! The description of headers are now up-to-date. I left a few comments for the places where I think need to be more clarified.

Thanks @nanjiangshu , very good comments! I added some new suggestions in the comments, but have not committed anything yet.

aaperis (dismissed):
Looks good! I agree with replacing code with an example regarding `?scheme`.

sda-download/api/api.md
nanjiangshu: You solution looks great!

MalinAhlberg:
The suggested changes are implemented in b303dd8, then rebased on main.

aaperis (approved):

pahatz (approved):
Looks good!

PR #983, opened 2024-08-08
Admin API part 3

jbygdell:
None

**Description**
This PR mainly adds the ways to define which users should have admin access.

**How to test**

### DIFF
--- a/sda/cmd/api/api.go
+++ b/sda/cmd/api/api.go
@@ -379,7 +379,7 @@ func listUserFiles(c *gin.Context) {
 	username = strings.TrimPrefix(username, "/")
 	username = strings.TrimSuffix(username, "/files")
 	log.Debugln(username)
-	files, err := Conf.API.DB.GetUserFiles(strings.ReplaceAll(username, "@", "_"))
+	files, err := Conf.API.DB.GetUserFiles(username)
 	if err != nil {
 		c.AbortWithStatusJSON(http.StatusInternalServerError, err.Error())
 
--- a/sda/cmd/api/api.go
+++ b/sda/cmd/api/api.go
@@ -75,12 +75,14 @@ func setup(config *config.Config) *http.Server {
 	r.GET("/ready", readinessResponse)
 	r.GET("/files", getFiles)
 	// admin endpoints below here
-	r.POST("/file/ingest", isAdmin(), ingestFile)                  // start ingestion of a file
-	r.POST("/file/accession", isAdmin(), setAccession)             // assign accession ID to a file
-	r.POST("/dataset/create", isAdmin(), createDataset)            // maps a set of files to a dataset
-	r.POST("/dataset/release/*dataset", isAdmin(), releaseDataset) // Releases a dataset to be accessible
-	r.GET("/users", isAdmin(), listActiveUsers)                    // Lists all users
-	r.GET("/users/:username/files", isAdmin(), listUserFiles)      // Lists all unmapped files for a user
+	if len(config.API.Admins) > 0 {
+		r.POST("/file/ingest", isAdmin(), ingestFile)                  // start ingestion of a file
+		r.POST("/file/accession", isAdmin(), setAccession)             // assign accession ID to a file
+		r.POST("/dataset/create", isAdmin(), createDataset)            // maps a set of files to a dataset
+		r.POST("/dataset/release/*dataset", isAdmin(), releaseDataset) // Releases a dataset to be accessible
+		r.GET("/users", isAdmin(), listActiveUsers)                    // Lists all users
+		r.GET("/users/:username/files", isAdmin(), listUserFiles)      // Lists all unmapped files for a user
+	}
 
 	cfg := &tls.Config{MinVersion: tls.VersionTLS12}
 
--- a/sda/cmd/api/api.md
+++ b/sda/cmd/api/api.md
@@ -118,3 +118,17 @@ Admin endpoints are only available to a set of whitelisted users specified in th
   - `200` Query execute ok.
   - `401` Token user is not in the list of admins.
   - `500` Internal error due to DB failure.
+
+#### Configure Admin users
+
+The users that should have administrative access can be set in two ways:
+
+- As a comma separated list of user identifiers asigned to: `admin.users`.
+- As a JSON file containg a list of the user identities, the path to the file is assigned to: `admin.configFile`. This is the recommended way.
+
+```json
+[
+"foo-user@example.com"
+"bar-user@example.com"
+]
+```
--- a/sda/cmd/api/api.md
+++ b/sda/cmd/api/api.md
@@ -29,11 +29,11 @@ Admin endpoints are only available to a set of whitelisted users specified in th
   - accepts `POST` requests with JSON data with the format: `{"filepath": "</PATH/TO/FILE/IN/INBOX>", "user": "<USERNAME>"}`
   - triggers the ingestion of the file.
 
-- Error codes
-  - `200` Query execute ok.
-  - `400` Error due to bad payload i.e. wrong `user` + `filepath` combination.
-  - `401` Token user is not in the list of admins.
-  - `500` Internal error due to DB or MQ failures.
+  - Error codes
+    - `200` Query execute ok.
+    - `400` Error due to bad payload i.e. wrong `user` + `filepath` combination.
+    - `401` Token user is not in the list of admins.
+    - `500` Internal error due to DB or MQ failures.
 
     Example:
 
@@ -45,11 +45,11 @@ Admin endpoints are only available to a set of whitelisted users specified in th
   - accepts `POST` requests with JSON data with the format: `{"accession_id": "<FILE_ACCESSION>", "filepath": "</PATH/TO/FILE/IN/INBOX>", "user": "<USERNAME>"}`
   - assigns accession ID to the file.
 
-- Error codes
-  - `200` Query execute ok.
-  - `400` Error due to bad payload i.e. wrong `user` + `filepath` combination.
-  - `401` Token user is not in the list of admins.
-  - `500` Internal error due to DB or MQ failures.
+  - Error codes
+    - `200` Query execute ok.
+    - `400` Error due to bad payload i.e. wrong `user` + `filepath` combination.
+    - `401` Token user is not in the list of admins.
+    - `500` Internal error due to DB or MQ failures.
 
     Example:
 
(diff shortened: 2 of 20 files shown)


sda/cmd/api/api.md
```diff
@@ -106,7 +106,7 @@ Admin endpoints are only available to a set of whitelisted users specified in th
 
 - `/users/:username/files`
   - accepts `GET` requests`
-  - Returns all files for a user with active uploads as a JSON array
+  - Returns all files (that is not part of a dataset) for a user with active uploads as a JSON array
```
aaperis, line 109:
```suggestion
  - Returns all files (that are not part of a dataset) for a user with active uploads as a JSON array
```

sda/internal/database/db_functions.go
```diff
@@ -634,7 +634,7 @@ func (dbs *SDAdb) getUserFiles(userID string) ([]*SubmissionFileInfo, error) {
 	files := []*SubmissionFileInfo{}
 	db := dbs.DB
 
-	// select all files of the user, each one annotated with its latest event
+	// select all files (that is not part of a dataset) of the user, each one annotated with its latest event
```
aaperis, line 637:
```suggestion
	// select all files (that are not part of a dataset) of the user, each one annotated with its latest event
```

aaperis (approved):
Nice!

sda/cmd/api/api.md
```diff
@@ -118,3 +118,17 @@ Admin endpoints are only available to a set of whitelisted users specified in th
   - `200` Query execute ok.
   - `401` Token user is not in the list of admins.
   - `500` Internal error due to DB failure.
+
+#### Configure Admin users
+
+The users that should have administrative access can be set in two ways:
+
+- As a comma separated list of user identifiers asigned to: `admin.users`.
```
MalinAhlberg, line 126:
```suggestion
- As a comma separated list of user identifiers assigned to: `admin.users`.
```

sda/cmd/api/api.go
```diff
@@ -75,12 +75,14 @@ func setup(config *config.Config) *http.Server {
 	r.GET("/ready", readinessResponse)
 	r.GET("/files", getFiles)
 	// admin endpoints below here
-	r.POST("/file/ingest", isAdmin(), ingestFile)                  // start ingestion of a file
-	r.POST("/file/accession", isAdmin(), setAccession)             // assign accession ID to a file
-	r.POST("/dataset/create", isAdmin(), createDataset)            // maps a set of files to a dataset
-	r.POST("/dataset/release/*dataset", isAdmin(), releaseDataset) // Releases a dataset to be accessible
-	r.GET("/users", isAdmin(), listActiveUsers)                    // Lists all users
-	r.GET("/users/:username/files", isAdmin(), listUserFiles)      // Lists all unmapped files for a user
+	if len(config.API.Admins) > 0 {
```
MalinAhlberg, line 78: So if there are no admins, all requests to these endpoints would get 404 instead of 403? What's the main reasoning behind this change (curious)?

sda/cmd/api/api.md
```diff
@@ -118,3 +118,17 @@ Admin endpoints are only available to a set of whitelisted users specified in th
...
+- As a JSON file containg a list of the user identities, the path to the file is assigned to: `admin.configFile`. This is the recommended way.
+
+```json
+[
+"foo-user@example.com"
```
nanjiangshu, line 131:
```suggestion
"foo-user@example.com",
```

sda/cmd/api/api.md
```diff
@@ -118,3 +118,17 @@ Admin endpoints are only available to a set of whitelisted users specified in th
...
+"bar-user@example.com"
```
nanjiangshu, line 132:
```suggestion
"foo-user@example.com",
"bar-user@example.com"
```

sda/cmd/api/api.go
nanjiangshu: I'm also curious.

nanjiangshu (commented):
Great work!

Apart from the comments in the code, I have a few other comments:
1. When I tested with the following command 
```
$ curl -s -w "%{http_code}\n" -H "Authorization: Bearer $token" -H "Content-Type: application/json" -X POST -d '{"accession_id": "my-id-01", "filepath": "/uploads/file.c4gh", "user": "testuser"}' http://localhost:8090/file/accession
```
It's a bit weird that the return code is `000`.

2. I got an empty list for the following command 
```
curl -H "Authorization: Bearer $token"  "http://localhost:8090/users/requester@demo.org/files" 
```
However,  I got 
```
$ curl -H "Authorization: Bearer $token"  "http://localhost:8090/files"
[{"inboxPath":"requester_demo.org/data/file1.c4gh","fileStatus":"uploaded","createAt":"2024-08-09T14:28:14.43718Z"}]
```
and 
```
$ curl -H "Authorization: Bearer $token"  "http://localhost:8090/users"
["requester@demo.org","test@dummy.org"] 
```
What did I do wrong?

3. Formatting in the README: `Error code` for each API endpoint section should be under the section, for example  
```
* /file/ingest
    - accepts POST requests with JSON data with the format: {"filepath": "</PATH/TO/FILE/IN/INBOX>", "user": "<USERNAME>"}
    - triggers the ingestion of the file.
* Error codes
    - 200 Query execute ok.
    - 400 Error due to bad payload i.e. wrong user + filepath combination.
```   
should be changed to 
```
- /file/ingest
    - accepts POST requests with JSON data with the format: {"filepath": "</PATH/TO/FILE/IN/INBOX>", "user": "<USERNAME>"}
    - triggers the ingestion of the file.

    * Error codes
        - 200 Query execute ok.
        - 400 Error due to bad payload i.e. wrong user + filepath combination.
```

sda/cmd/api/api.go
jbygdell: 404 is the standard "not found" response which you get for any query that the app is not designed to handle.

pontus (dismissed):

sda/cmd/api/api.go
MalinAhlberg: yes, I am aware of that, but it's unclear to me why there would be a distinction between the cases
"the user is not an admin but there are other admins" (403) vs 
"the user is not an admin and there are no other admins" (404)

pontus: I see this as "administrative users are specified" being an alternate way of configuring "administrative API should be enabled" and that way it makes sense to me (even if there was a separate setting for it, I'd still prefer admin APIs to be unreachable if no administrative users are configured).

jbygdell: 404 == admin endpoints not enabled.
403 == endpoints enabled but user is not an admin.

pontus (dismissed):

sda/cmd/api/api.go
MalinAhlberg: Ok, I think I get your point. Thanks @pontus and @jbygdell !

MalinAhlberg (approved):

nanjiangshu (dismissed):

sda/cmd/api/api_test.go
```diff
@@ -1138,7 +1139,7 @@ func (suite *TestSuite) TestListUserFiles() {
 	testUsers := []string{"user_example.org", "User-B", "User-C"}
```
aaperis, line 1139:
```suggestion
	testUsers := []string{"user@example.org", "User-B", "User-C"}
```

sda/cmd/api/api_test.go
```diff
@@ -1138,7 +1139,7 @@ func (suite *TestSuite) TestListUserFiles() {
...
 	for _, user := range testUsers {
 		for i := 0; i < 5; i++ {
-			fileID, err := Conf.API.DB.RegisterFile(fmt.Sprintf("/%v/TestGetUserFiles-00%d.c4gh", user, i), user)
+			fileID, err := Conf.API.DB.RegisterFile(fmt.Sprintf("/%v/TestGetUserFiles-00%d.c4gh", user, i), strings.ReplaceAll(user, "_", "@"))
```
aaperis, line 1142:
```suggestion
			fileID, err := Conf.API.DB.RegisterFile(fmt.Sprintf("/%v/TestGetUserFiles-00%d.c4gh", user, i), user)
```
Since now we don't substitute `requester@demo.org` for requester_demo.org, if I understand correctly.

aaperis (approved):
Minor suggestions :-)

sda/cmd/api/api_test.go
jbygdell: Nope, look at what's really happening here and which functions being called.

sda/cmd/api/api_test.go
jbygdell: Nope, look at what's really happening here and which functions being called.

MalinAhlberg (approved):

sda/cmd/api/api_test.go
aaperis: hmm from a look at our databases, I am missing why `UpdateFileEventLog` admits a username with `_` instead of `@` for `upload` events whereas any other event has username with `@`. Not a big deal but I don't get why :sweat_smile:

jbygdell: Filepaths needs `_` but the `submission_user` has an `@`. The tests is done backwards since we then only needs to do substitution in one place.

sda/internal/broker/broker_test.go
```diff
@@ -32,15 +32,9 @@ var tMqconf = MQConf{}
 
 func TestMain(m *testing.M) {
 	certPath, _ = os.MkdirTemp("", "gocerts")
-	defer os.RemoveAll(certPath)
```
MalinAhlberg, line 35: Although I assume you have good reason, I don't really see why this change is done..? What if there is panic somewhere before `m.Run()`, would line 106 really be reached?

jbygdell: The linter complained about not having an `os.Exit` and then it started to complain about the `defer` not being called due to the `os.Exit`.

The other option is of course to just ignore the linter complaints about not having an `os.Exit`

MalinAhlberg: Hehe, interesting. I think ignore the linter for that line sounds like a nice(r) option then, so that we don't end up with unwanted files/dirs hanging around. What do you think?

aaperis: I think that since the cert files are created in the system's temp dir, they will be cleaned up at some point either way so this shouldn't be a problem.

MalinAhlberg: Ok, good point. I resolve this comment and approve. If the `defer` is re-added and linter-ignored, I will approve that as well :smile:

MalinAhlberg (approved):

PR #988, opened 2024-08-14
add grpc healthcheck

### LINKED ISSUE #743: [reencrypt] implement health check endpoint
The reencrypt service needs to have a proper health check endpoint so we know that the service is working properly.
The TCP socket health check tells if the service responds but does not add info on whether it actually works as intended.

**Acceptance criteria**
- [ ] gRPC health check is implemented in the go code of reencrypt
- [ ] helm chart is update to use `gRPC` instead of `TCPsocket`

**Additional context**
[GRPC Health Checking Protocol](https://github.com/grpc/grpc/blob/master/doc/health-checking.md)
[kubernetes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-grpc-liveness-probe) specific info

aaperis:
This PR closes #743.

**Description**
The native k8s grpc probe cannot do rpc's to services running behind TLS. For implementing k8s grpc health checks in such a case,  the recommended way is to follow this [paradigm](https://github.com/grpc/grpc-java/tree/master/examples/example-xds) and implement a grpc health server  that is separate from the server where the probed services run.

In this PR we have the following scheme: 
1. the grpc server that exposes the reencrypt service also exposes a health service that reports the health status of the reencrypt service and that of the server overall. This can be TLS secured if TLS is enabled.
2. A separate grpc server which only exposes a health service is also spun up. This proxy grpc health server listens to a separate port with no TLS enabled, and implements a healthcheck function that probes a given service of the TLS enabled server (e.g. the reencrypt service) and returns back the status of that service.
3. A rpc probe to the proxy health server with an empty service value, `service: ""`,  returns the overall status of the TLS enabled server and with  `service: reencrypt.Reencrypt` returns the status of the reencrypt service of the TLS enabled server, etc.
 

**How to test**
- See that integration tests for Helm charts pass.

### DIFF
--- a/charts/sda-svc/templates/re-encrypt-certificate.yaml
+++ b/charts/sda-svc/templates/re-encrypt-certificate.yaml
@@ -18,6 +18,7 @@ spec:
     algorithm: ECDSA
     size: 256
   usages:
+    - client auth
     - server auth
   # At least one of a DNS Name, URI, or IP address is required.
   dnsNames:
--- a/charts/sda-svc/templates/re-encrypt-deploy.yaml
+++ b/charts/sda-svc/templates/re-encrypt-deploy.yaml
@@ -74,10 +74,14 @@ spec:
         ports:
           - name: grpc
             containerPort: {{ ternary 50443 50051 ( .Values.global.tls.enabled ) }}
+          - name: grpchealth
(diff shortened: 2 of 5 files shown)


MalinAhlberg (approved):
Nice job, and thanks for the explanation in the PR description (very needed for my understanding :sweat_smile: )!

MalinAhlberg (approved):

PR #993, opened 2024-08-16
inbox: tls for the health endpoint

### LINKED ISSUE #459: [s3inbox] add TLS listener to the healthcheck endpoint
In order to provide meaningful responses for external health check services like [uptimerrobot](https://uptimerobot.com/) we should make the health endpoint available externally.

A/C
* Healtcheck uses TLS if configured
* s3inbox deplpyment is modified so that the liveness/radiness probes use TLS if configured
* Service and ingress in the helm chart is updates so the endpoint is available externally

MalinAhlberg:
This PR closes #459 .

**Description**
- healthcheck server uses tls, if configured to do so
- charts are updated to expose `/healthz` endpoint for inbox
- timeout for the `broker-tcp` health check of inbox is increased (from very short 5 ms)

**How to test**
Make sure all checks pass, specifically `Build PR container / chart (1.28, true, s3)`. The healthendpoint will be exposed at `GET inbox.../healthz`

### DIFF
--- a/charts/sda-svc/templates/s3-inbox-ingress.yaml
+++ b/charts/sda-svc/templates/s3-inbox-ingress.yaml
@@ -44,7 +44,7 @@ spec:
             name: {{ template "sda.fullname" . }}-inbox
             port:
               number: {{ ternary 443 80 .Values.global.tls.enabled }}
-      - pathType: Prefix
+      - pathType: Exact
         path: "/healthz"
         backend:
           service:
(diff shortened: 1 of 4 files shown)


charts/sda-svc/templates/s3-inbox-ingress.yaml
```diff
@@ -44,6 +44,13 @@ spec:
             name: {{ template "sda.fullname" . }}-inbox
             port:
               number: {{ ternary 443 80 .Values.global.tls.enabled }}
+      - pathType: Prefix
```
jbygdell, line 47:
```suggestion
      - pathType: Exact
```

Since we don't have implemented any sub paths here we can set it to exact.

jbygdell (approved):

nanjiangshu (approved):
Great work!

PR #994, opened 2024-08-19
Add key hash to the files table

### LINKED ISSUE #788: Extend the "files" table with the key hash
As a sda-dev,
I want to know what key was used to encrypt a specific file
So we can handle multiple keys.

In order to accomplish this we should extend the `files` table. PO thinks it would be good to have a separate table for the keys where we can apart from just the key hash can also store when the key was created and possibly also at what date it was deprecated. The primary key can be the key hash.

**Acceptance criteria**
- [ ] `files` table includes key hash used for encryption
- [ ] another table where info for the key is stored (creation and deprecation date)
- [ ] migration to new schema version (update all entries with key hash and key info as well)
- [ ] tests verifying that key can be retrieved correctly from the database

This is demoable.

**Estimation of size**: small

nanjiangshu:
This PR closes #788 

**Description**
**Changes made**
1. Added a new column `key_hash` into the `files` table
2. Added a new table `encryption_keys` for storing the detailed information of the encryption keys
3. Added the migration script for these changes and updated the dbschema_version

**How to test**
1. Build the images locally by running the following command in the root folder of the repo
```
make build-sda build-postgresql 
```
2. Start the docker setup by 
```
docker compose -f .github/integration/sda-s3-integration.yml up -d
```
3. Check in the database if `key_hash` exists in the `files` table and the new table `encryption_keys` exists.
4. Insert a record to the table `encryption_keys` by e.g. 

```
INSERT INTO sda.encryption_keys (key_hash, description) VALUES ('example_key_hash', 'Example encryption key description'); 
```
and then try to insert a record to the `files` table with the `key_hash` matching `example_key_hash` and that does not match `example_key_hash`.
The former should work and the latter should fail.

**Additional note**
To test if the migration scripts works on the existing database, one can first start the docker setup with the code from the main branch and then turn down the docker setup without removing the database volume. After that, build the images with this branch and start the docker setup again.

### DIFF
--- a/postgresql/initdb.d/01_main.sql
+++ b/postgresql/initdb.d/01_main.sql
@@ -25,7 +25,8 @@ VALUES (0, now(), 'Created with version'),
        (8, now(), 'Add ingestion functions'),
        (9, now(), 'Add dataset event log'),
        (10, now(), 'Create Inbox user'),
-       (11, now(), 'Grant select permission to download on dataset_event_log');
+       (11, now(), 'Grant select permission to download on dataset_event_log'),
+       (12, now(), 'Add key hash');
 
 -- Datasets are used to group files, and permissions are set on the dataset
 -- level
@@ -35,6 +36,14 @@ CREATE TABLE datasets (
     title               TEXT,
     description         TEXT,
     created_at          TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT clock_timestamp()
+);
+
+-- the keys table is used to store the information of the encryption keys
+CREATE TABLE encryption_keys (
+    key_hash          TEXT PRIMARY KEY,
+    created_at        TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT clock_timestamp(),
+    deprecated_at     TIMESTAMP WITH TIME ZONE,
+    description       TEXT
 );
 
 -- `files` is the main table of the schema, holding the file paths, encryption
@@ -53,6 +62,7 @@ CREATE TABLE files (
 
     header               TEXT,
     encryption_method    TEXT,
+    key_hash             TEXT REFERENCES encryption_keys(key_hash),
 
     -- Table Audit / Logs
     created_by           NAME DEFAULT CURRENT_USER, -- Postgres users
--- a/postgresql/migratedb.d/12.sql
+++ b/postgresql/migratedb.d/12.sql
@@ -1,0 +1,28 @@
+DO
+$$
+DECLARE
+-- The version we know how to do migration from, at the end of a successful migration
+-- we will no longer be at this version.
+  sourcever INTEGER := 11;
+  changes VARCHAR := 'Add key hash';
+BEGIN
+  IF (select max(version) from sda.dbschema_version) = sourcever then
+    RAISE NOTICE 'Doing migration from schema version % to %', sourcever, sourcever+1;
+    RAISE NOTICE 'Changes: %', changes;
+    INSERT INTO sda.dbschema_version VALUES(sourcever+1, now(), changes);
+
+    CREATE TABLE IF NOT EXISTS sda.encryption_keys (
+        key_hash          TEXT PRIMARY KEY,
+        created_at        TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT clock_timestamp(),
+        deprecated_at     TIMESTAMP WITH TIME ZONE,
+        description       TEXT
+    );
+
+    ALTER TABLE sda.files ADD COLUMN IF NOT EXISTS key_hash TEXT,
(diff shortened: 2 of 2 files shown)


postgresql/initdb.d/01_main.sql
```diff
@@ -53,6 +62,8 @@ CREATE TABLE files (
 
     header               TEXT,
     encryption_method    TEXT,
+    key_hash             TEXT REFERENCES encryption_keys(key_hash) ON DELETE SET NULL,
```
viklund, line 65: I'm not sure the "ON DELETE SET NULL" should be there. If we delete a key in the encryption_keys table there should not be any key_hashes in the files table with that key. So if there are, an error should occur.

viklund (changes requested):
Looks good, but I think the "ON DELETE SET NULL" is incorrect. It's fine if the key_hash is null initially. But if a key is removed and there are still files using that key, we should throw an error.

postgresql/migratedb.d/12.sql
```diff
@@ -0,0 +1,28 @@
...
+
+    CREATE TABLE IF NOT EXISTS sda.encryption_keys (
+        key_hash          TEXT PRIMARY KEY,
+        created_at        TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT clock_timestamp(),
+        deprecated_at     TIMESTAMP WITH TIME ZONE,
+        description       TEXT
+    );
+
+    ALTER TABLE sda.files ADD COLUMN IF NOT EXISTS key_hash TEXT,
+    ADD CONSTRAINT fk_files_key_hash FOREIGN KEY (key_hash) REFERENCES sda.encryption_keys(key_hash) ON DELETE SET NULL;
```
viklund, line 22: Here too, not "ON DELETE SET NULL"

postgresql/initdb.d/01_main.sql
nanjiangshu: Thanks for the comment @viklund. Do you want to ensure that a key from the `encryption_keys` table cannot be deleted if there are still files in the `files` table that reference it? If so, we should use 

```
key_hash             TEXT REFERENCES encryption_keys(key_hash) ON DELETE RESTRICT,
```
In this case, psql will throw an error if we try to delete a key from the `encryption_keys` table that is still referenced in the `files` table.

MalinAhlberg (commented):
Nice work, @nanjiangshu !
Should the be any privileges granted for the `encryption_keys` table (in `04_grants.sql`)?

jbygdell:
> Nice work, @nanjiangshu ! Should the be any privileges granted for the `encryption_keys` table (in `04_grants.sql`)?

That will be handled when we implement the actual usage of this new stuff

MalinAhlberg:
> > That will be handled when we implement the actual usage of this new stuff
> 
> It's a valid point @MalinAhlberg, I mentioned briefly this morning during the standup and the decision is as @jbygdell answered 😄

Ah, ok. And oops, sorry for not remembering.

MalinAhlberg (dismissed):
Apart from Johan's comment, I think everything looks good and works as expected :+1:

postgresql/initdb.d/01_main.sql
jbygdell: There shouldn't be any need to delete anything from the `encryption_keys` table, although a nice sanity feature would be to have the restriction in place.

viklund: Yes, explicitly specifying RESTRICT might not be a bad idea. Not sure what difference it makes compared to not having a "ON DELETE" option.

nanjiangshu: I removed the "ON DELETE" option completely and the behavior of the database operation meets all requirement. 
1. Can insert a row to the `files` table without `key_hash`, i.e. set `key_hash` to NULL
2. Insert a row in the `files` table with the key_hash not matching any record in the `encryption_keys` table will raise an error
3. Delete a key from the `encryption_keys` table that is still referenced in the `files` table will raise an error.

The changes have been implemented in the commit d7565f2a

aaperis (dismissed):
It looks great :-)

MalinAhlberg (approved):

jbygdell (changes requested):
Squash and rebase

nanjiangshu:
> Squash and rebase

I rebase and squashed commits upon @jbygdell 's request. Now it needs your re-approval @aaperis , @MalinAhlberg 😃

jbygdell:
> > Squash and rebase
> 
> I rebase and squashed commits upon @jbygdell 's request. Now it needs your re-approval @aaperis , @MalinAhlberg 😃

It's still 7 commmits, two of which are `merge commits` that should not be part of the PR.
These changes could easily be confined to two commits, one for each file being changed.

jbygdell (approved):

aaperis (approved):

viklund (approved):

PR #1001, opened 2024-08-21
add checks for integer overflows

aaperis:
**Description**
This PR addresses errors of the latest version of the go linter, see e.g.
https://github.com/neicnordic/sensitive-data-archive/actions/runs/10488098518/job/29049848758#step:4:27
https://github.com/neicnordic/sensitive-data-archive/actions/runs/10488098518/job/29049848067#step:4:27

that have to do with unsafe integer conversions. Manual safe checks are added and the linter is silenced.

**How to test**
Linter tests pass.

### DIFF
--- a/sda-download/api/sda/sda_test.go
+++ b/sda-download/api/sda/sda_test.go
@@ -7,6 +7,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"io"
+	"math"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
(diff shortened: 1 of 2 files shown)


sda-download/api/sda/sda_test.go
```diff
@@ -677,7 +677,8 @@ func (f *fakeGRPC) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	_, err = w.Write([]byte{0})
 	assert.NoError(f.t, err, "Could not write response flag")
 
-	err = binary.Write(w, binary.BigEndian, int32(len(response)))
+	assert.Less(f.t, len(response), int(^uint32(0)), "Response too long")
```
MalinAhlberg, line 680: I might be confusing things here, but why not use `math.MaxInt32` instead? ` int(^uint32(0)=4294967295` and `int32(4294967294)`  does overflow..?

aaperis: 1. I didn't think to use a built in function for this!
2. I didn't notice that here we deal with `int32` and NOT `unint32` (see e.g. the fix for reencrypt) :blush: Great catch!

My old self would be inclined to change `int(^uint32(0))` to `int(^uint32(0)/2)` and marvel at the fact that no new module was imported for this bit. 
But my new self values readability so this is now changed, thanks for pointing out :smile:

MalinAhlberg (approved):
Looks good!

jbygdell (approved):

Issue #992, opened 2024-08-16
[s3inbox] rework healthcheck


jbygdell:
**Please describe the feature**
As a repository admin I want the healthcheck to be connected to the main process so that any failures will be fixed directly.

**Acceptance criteria**
- [ ] Rework the health endpoint to run on the main http thread
- [ ] utilize the concept in the other services:
  - [ ] check if the MQ channel is open
  - [ ] check that the DB responds to ping
  - [ ] check that the s3 backend responds
- [ ] Tests verifying the changes are added

**Additional context**

**Estimation of size**: small

**Estimation of priority**: medium

PR #1025, opened 2024-09-03
[s3inbox] run healthchecks in main process

MalinAhlberg:
This PR addresses #992.

**Description**
- The `/health` endpoint is put on the same port as s3inbox (ie. is run on the server started in `s3inbox.go`).
- Also a `HEAD` request to s3inbox will return the health status.
- If there are problems with the connection to MQ or DB while checking the health, `CheckHealth` attempts to restore it.
- Go tests updated
- Integration tests added
- The `/ready` and `/live` endpoints are removed
- The `/healthz` enpoint (provided by the charts) is removed 

If this is to be continue: create issues for updating charts etc reflect this change.

**How to test**
```curl 'localhost:18000' -H 'Authorization: Bearer $token' -I```
```curl 'localhost:18000/health' -H 'Authorization: Bearer $token'```

### DIFF
--- a/.github/integration/tests/sda/09_healthchecks.sh
+++ b/.github/integration/tests/sda/09_healthchecks.sh
@@ -22,7 +22,7 @@ if [ "$response" != "200" ]; then
 	exit 1
 fi
 
-response="$(curl -s -k -LI "http://s3inbox:8000/health" -o /dev/null -w "%{http_code}\n")"
+response="$(curl -s -k -L "http://s3inbox:8000/health" -o /dev/null -w "%{http_code}\n")"
 if [ "$response" != "200" ]; then
 	echo "Bad health response from /health, expected 200 got: $response"
 	exit 1
--- a/.github/integration/tests/sda/09_healthchecks.sh
+++ b/.github/integration/tests/sda/09_healthchecks.sh
@@ -1,0 +1,33 @@
+#!/bin/sh
+set -e
+
+# install tools if missing
+for t in curl  jq ; do
+    if [ ! "$(command -v $t)" ]; then
+        if [ "$(id -u)" != 0 ]; then
+            echo "$t is missing, unable to install it"
+            exit 1
+        fi
+
+        apt-get -o DPkg::Lock::Timeout=60 update >/dev/null
+        apt-get -o DPkg::Lock::Timeout=60 install -y "$t" >/dev/null
+    fi
+done
+
+
+# Test the s3inbox's healthchecks, GET /health and HEAD /
+token="$(curl -s http://oidc:8080/tokens | jq -r '.[0]')"
+
+response="$(curl -s -k -LI "http://s3inbox:8000" -o /dev/null -w "%{http_code}\n" -H "Authorization: Bearer $token")"
+if [ "$response" != "200" ]; then
+	echo "Bad health response from HEAD /, expected 200 got: $response"
+	exit 1
+fi
+
+response="$(curl -s -k -LI "http://s3inbox:8000/health" -o /dev/null -w "%{http_code}\n" -H "Authorization: Bearer $token")"
+if [ "$response" != "200" ]; then
+	echo "Bad health response from /health, expected 200 got: $response"
+	exit 1
+fi
+
+echo "Healthcheck tests completed successfully"
--- a/.github/integration/tests/sda/09_healthchecks.sh
+++ b/.github/integration/tests/sda/09_healthchecks.sh
@@ -16,15 +16,13 @@ done
 
 
 # Test the s3inbox's healthchecks, GET /health and HEAD /
-token="$(curl -s http://oidc:8080/tokens | jq -r '.[0]')"
-
-response="$(curl -s -k -LI "http://s3inbox:8000" -o /dev/null -w "%{http_code}\n" -H "Authorization: Bearer $token")"
+response="$(curl -s -k -LI "http://s3inbox:8000" -o /dev/null -w "%{http_code}\n")"
 if [ "$response" != "200" ]; then
 	echo "Bad health response from HEAD /, expected 200 got: $response"
 	exit 1
 fi
 
-response="$(curl -s -k -LI "http://s3inbox:8000/health" -o /dev/null -w "%{http_code}\n" -H "Authorization: Bearer $token")"
+response="$(curl -s -k -LI "http://s3inbox:8000/health" -o /dev/null -w "%{http_code}\n")"
 if [ "$response" != "200" ]; then
 	echo "Bad health response from /health, expected 200 got: $response"
 	exit 1
--- a/charts/sda-svc/templates/s3-inbox-ingress.yaml
+++ b/charts/sda-svc/templates/s3-inbox-ingress.yaml
@@ -44,13 +44,6 @@ spec:
             name: {{ template "sda.fullname" . }}-inbox
             port:
               number: {{ ternary 443 80 .Values.global.tls.enabled }}
-      - pathType: Exact
-        path: "/healthz"
-        backend:
-          service:
-            name: {{ template "sda.fullname" . }}-inbox
-            port:
-              number: {{ ternary 443 80 .Values.global.tls.enabled }}
 {{- if .Values.global.tls.enabled }}
   tls:
   - hosts:
--- a/charts/sda-svc/templates/s3-inbox-ingress.yaml
+++ b/charts/sda-svc/templates/s3-inbox-ingress.yaml
@@ -50,7 +50,7 @@ spec:
           service:
             name: {{ template "sda.fullname" . }}-inbox
             port:
-              number: 8001
+              number: {{ ternary 443 80 .Values.global.tls.enabled }}
 {{- if .Values.global.tls.enabled }}
   tls:
   - hosts:
--- a/sda/cmd/s3inbox/healthchecks.go
+++ b/sda/cmd/s3inbox/healthchecks.go
@@ -57,6 +57,8 @@ func (p *Proxy) CheckHealth(w http.ResponseWriter, _ *http.Request) {
 	if err != nil {
 		log.Error(err)
 		w.WriteHeader(http.StatusServiceUnavailable)
+
+		return
 	}
 	w.WriteHeader(http.StatusOK)
 }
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

**Comment by jbygdell** [comment]
You're on the right path.
But the healcheck functions needs to be inside the proxy struct so it can issue reconnect/recreate if the MQ or DB connections go down. Look in the `checkAndSendMessage` function and see an example.

File: sda/cmd/s3inbox/healthchecks.go [hunk 1]
```diff
@@ -1,114 +1,91 @@
...
-func (h *HealthCheck) httpsGetCheck(url string, timeout time.Duration) healthcheck.Check {
-	cfg := &tls.Config{MinVersion: tls.VersionTLS12}
-	cfg.RootCAs = h.tlsConfig.RootCAs
-	tr := &http.Transport{TLSClientConfig: cfg}
-	client := http.Client{
-		Transport: tr,
-		Timeout:   timeout,
-		// never follow redirects
-		CheckRedirect: func(*http.Request, []*http.Request) error {
-			return http.ErrUseLastResponse
```
MalinAhlberg, line 98: Using the client of proxy instead, hence not setting tls stuff or timeout.

.github/integration/tests/sda/09_healthchecks.sh
```diff
@@ -0,0 +1,33 @@
...
+        apt-get -o DPkg::Lock::Timeout=60 update >/dev/null
+        apt-get -o DPkg::Lock::Timeout=60 install -y "$t" >/dev/null
+    fi
+done
+
+
+# Test the s3inbox's healthchecks, GET /health and HEAD /
+token="$(curl -s http://oidc:8080/tokens | jq -r '.[0]')"
+
+response="$(curl -s -k -LI "http://s3inbox:8000" -o /dev/null -w "%{http_code}\n" -H "Authorization: Bearer $token")"
```
jbygdell, line 21:
```suggestion
response="$(curl -s -k -LI "http://s3inbox:8000" -o /dev/null -w "%{http_code}\n")"
```

The health endpoint should not require any authentication.

.github/integration/tests/sda/09_healthchecks.sh
```diff
@@ -0,0 +1,33 @@
...
+if [ "$response" != "200" ]; then
+	echo "Bad health response from HEAD /, expected 200 got: $response"
+	exit 1
+fi
+
+response="$(curl -s -k -LI "http://s3inbox:8000/health" -o /dev/null -w "%{http_code}\n" -H "Authorization: Bearer $token")"
```
jbygdell, line 27:
```suggestion
response="$(curl -s -k -LI "http://s3inbox:8000/health" -o /dev/null -w "%{http_code}\n")"
```

sda/cmd/s3inbox/healthchecks.go
```diff
@@ -1,114 +1,91 @@
...
 
-		return nil
+func (p *Proxy) getS3ReadyPath() string {
+	s3URL := p.s3.URL
+	if p.s3.Port != 0 {
+		s3URL = fmt.Sprintf("%s:%d", s3URL, p.s3.Port)
 	}
+	if p.s3.Readypath != "" {
+		s3URL += p.s3.Readypath
+	}
```
jbygdell, line 88: This should probably be sanitized using `net/url` and `path.join` so we don't risk getting a malformed URL

sda/cmd/s3inbox/s3inbox.go
```diff
@@ -7,6 +7,8 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/gorilla/mux"
```
jbygdell, line 10:
```suggestion
```

MUX is part of the standard [HTTP library](https://pkg.go.dev/net/http@go1.22.2#ServeMux) since GO 1.22

sda/cmd/s3inbox/healthchecks.go
```diff
@@ -1,114 +1,91 @@
...
-	client := http.Client{
-		Transport: tr,
-		Timeout:   timeout,
-		// never follow redirects
-		CheckRedirect: func(*http.Request, []*http.Request) error {
-			return http.ErrUseLastResponse
-		},
+// httpsGetCheck sends a request to the S3 backend and makes sure it is healthy
+func (p *Proxy) httpsGetCheck(url string) error {
+
```
jbygdell, line 68:
```suggestion
```

sda/cmd/s3inbox/healthchecks.go
```diff
@@ -1,114 +1,91 @@
...
+		if err != nil {
+			w.WriteHeader(http.StatusServiceUnavailable)
+
+			return
 		}
 	}
+
+	// Check that s3 backend responds
+	s3URL := p.getS3ReadyPath()
+	err = p.httpsGetCheck(s3URL)
```
jbygdell, line 56:
```suggestion
	err = p.httpsGetCheck(p.getS3ReadyPath())
```

jbygdell (changes requested):

sda/cmd/s3inbox/s3inbox.go
MalinAhlberg: Yes, but to do route [HEAD  requests to /](https://github.com/neicnordic/sensitive-data-archive/pull/1025/files#diff-30d249c8dab4f2de2d776b274b7ea1cf0ae9bf858f21b88bfa8aa7f31099817dR107) I needed gorilla mux.

MalinAhlberg:
Thanks for the quick review @jbygdell! Fixes are pushed.

sda/cmd/s3inbox/healthchecks.go
```diff
@@ -1,114 +1,91 @@
...
-		return nil
+func (p *Proxy) getS3ReadyPath() string {
+	s3URL := p.s3.URL
+	if p.s3.Port != 0 {
+		s3URL = path.Join(s3URL, strconv.Itoa(p.s3.Port))
 	}
+	if p.s3.Readypath != "" {
+		s3URL += path.Join(s3URL, p.s3.Readypath)
+	}
+
```
jbygdell, line 89: This will not work since `path.Join(s3URL, strconv.Itoa(p.s3.Port)` generates a string like this: `example.com/555` instead of `example.com:555`


```suggestion
	s3URL := url.Parse(p.s3.URL)
	if p.s3.Port != 0 {
		s3URL.Host = net.JoinHostPort(s3URL.Hostname(), strconv.Itoa(p.s3.Port))
	}
	if p.s3.Readypath != "" {
		s3URL.Path = path.Join(s3URL.Path, p.s3.Readypath)
	}
```
And the you have to return `s3URL.String()`

sda/cmd/s3inbox/s3inbox.go
jbygdell: with the standard library you should be able to do something like this:

```
mux.HandleFunc(GET "/health", proxy.CheckHealth ) // this will only accept GET and HEAD requests.
mux.HandleFunc("/", proxy)  // this will accept any request and then at the start of the proxy function where we detect which request it is we handle the HEAD request to "/" as one case.
```

But we can leave that for later since there are other things in this app that should be refactored.

sda/cmd/s3inbox/healthchecks.go
MalinAhlberg: Oops. Rewritten and added some tests.

charts/sda-svc/templates/s3-inbox-ingress.yaml
```diff
@@ -50,7 +50,7 @@ spec:
           service:
             name: {{ template "sda.fullname" . }}-inbox
             port:
-              number: 8001
+              number: {{ ternary 443 80 .Values.global.tls.enabled }}
```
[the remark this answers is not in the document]
jbygdell: The healthz endpoint here will not work since that path doesn't exists in the s3inbox

.github/integration/tests/sda/09_healthchecks.sh
```diff
@@ -0,0 +1,31 @@
...
+
+
+# Test the s3inbox's healthchecks, GET /health and HEAD /
+response="$(curl -s -k -LI "http://s3inbox:8000" -o /dev/null -w "%{http_code}\n")"
+if [ "$response" != "200" ]; then
+	echo "Bad health response from HEAD /, expected 200 got: $response"
+	exit 1
+fi
+
+response="$(curl -s -k -LI "http://s3inbox:8000/health" -o /dev/null -w "%{http_code}\n")"
```
jbygdell, line 25: Shouldn't  this be a `GET` instead of a `HEAD` request (-I)

MalinAhlberg: It should! Fixed in 3446820.

charts/sda-svc/templates/s3-inbox-ingress.yaml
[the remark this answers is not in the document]
MalinAhlberg: Ok, then I remove it, in 65e71af

jbygdell (approved):

.github/integration/tests/sda/09_healthchecks.sh
```diff
@@ -22,7 +22,7 @@ if [ "$response" != "200" ]; then
 	exit 1
 fi
 
-response="$(curl -s -k -LI "http://s3inbox:8000/health" -o /dev/null -w "%{http_code}\n")"
+response="$(curl -s -k -L "http://s3inbox:8000/health" -o /dev/null -w "%{http_code}\n")"
```
nanjiangshu, line 25: I'm curious why use GET instead of HEAD request given that what we want here is only the HTTP status code?
Also the HEAD request is used in the previous command.

nanjiangshu (approved):
Great work! I left a minor comment regarding the use of GET request and HEAD request for obtaining the HTTP code. But I approve it anyway 😄

.github/integration/tests/sda/09_healthchecks.sh
MalinAhlberg: Good question, I see your point..! It's just to make sure that a GET request does work, in contrast to `curl -LI "http://s3inbox:8000" ` where you need to do a HEAD to get the health. Maybe both should be tested. I would argue that this is good enough for now, anyway :).

PR #1026, opened 2024-09-03
sda-admin cli 

### LINKED ISSUE #1007: Migrate the sda-admin script to a go cli tool
As a sda-dev I need a cli tool, written in golang, to interact with the admin-api 
so that 
I can operate the archive and maintain the tool easily at the same time.

Description
- this should be created in the sensitive data archive repo

A/C:
- the cli should be able to call all admin-api endpoints
- make this project agnostic
- add testsuite

Demo:
functionality

Dublicate: https://github.com/neicnordic/sensitive-data-archive/issues/1004

nanjiangshu:
This PR closes #1007 

**Description**
The command line tool that can talk to SDA API endpoints. It provides functionalities to list users and files, ingest and set accession IDs for files, and create or release datasets.

**How to test**
1. At the root of the repo, start the stack by running 
```
make build-all
PR_NUMBER=$(/usr/bin/date '+%F')  docker compose -f .github/integration/sda-s3-integration.yml run integration_test
```
2. Set the ENVs by 
```
export ACCESS_TOKEN=$(curl -s -k http://localhost:8080/tokens | jq -r '.[0]')
```
3. Build the executable in the folder `sda-admin` by 
```
go build 
```

3. Test various functionalities 
```
# list all users
./sda-admin list users  

# list all files for the user requester@demo.org
./sda-admin list files  

# list all files for the user test@dummy.org
./sda-admin list files -user test@dummy.org 

# trigger the ingestion of a given file 
./sda-admin file ingest --filepath requester_demo.org/data/file1.c4gh --user requester@demo.org  

# assign an accession ID to a file
./sda-admin file accession --filepath requester_demo.org/data/file1.c4gh  --user requester@demo.org   --accession-id my-id1 

# create a dataset from a list of accession IDs and the dataset ID
./sda-admin dataset create --dataset-id dataset001 my-id1 my-id2 

# # release a dataset for downloading
./sda-admin dataset release --dataset-id dataset001 
```


sda-admin/file/file.go
```diff
@@ -0,0 +1,66 @@
...
+		return fmt.Errorf("failed to marshal JSON, reason: %v", err)
+	}
+
+	_, err = helpers.PostRequest(url, token, jsonBody)
+
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
jbygdell, line 42:
```suggestion
func Ingest(apiURL, token, username, filepath string) error {
	url, err := url.Parse(apiURL)
	if err != nil {
		return err
	}

	url.Path = path.Join(url.Path, "/file/ingest")

	requestBody := RequestBodyFileIngest{
		Filepath: filepath,
		User:     username,
	}

	jsonBody, err := json.Marshal(requestBody)
	if err != nil {
		return fmt.Errorf("failed to marshal JSON, reason: %v", err)
	}

	_, err = helpers.PostRequest(url.String(), token, jsonBody)
	if err != nil {
		return err
	}

	return nil
}
```
This function is part of the `file` package, as such any calls to this function will be prefixed with `file.` so having `File` as the start of the name makes it stutter `file.FileIngest`

When joining URIs and paths use the library helpers to ensure you get a proper string in the end, without bad characters or double slashes.

Always have a blank line before any return unless it is the only statement in an if block.

sda-admin/list/list.go
```diff
@@ -0,0 +1,41 @@
...
+func ListFiles(api_uri, token, username string) error {
+	var url string
+
+	if username == "" {
+		url = api_uri + "/files"
+	} else {
+		url = fmt.Sprintf("%s/users/%s/files", api_uri, username)
+	}
+
+	response, err := helpers.GetResponseBody(url, token)
```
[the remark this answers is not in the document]
nanjiangshu: I see. modified in the commit f087a02

sda-admin/file/file.go
nanjiangshu: using path.Join() to concatenate url may have problems on Windows. I have modified the function names and also using `url.Parse()` as you suggested in the commit 6952f7d

jbygdell: If a windows environment will mess things up make use of the standard libraries to overcome that.
`filepath.Toslash` is very convenient to ensure that the final path is UNIX styled.

You current solution does not fix the issue of getting double slashes in the final path that `path.Join` does.

jbygdell: But the proper way of figuring out what needs to be done is to create a test case where you check that the final URL is correctly formatted

Since we will have to set up an GH action to test this app in a windows environment anyway.

sda-admin/main.go
```diff
@@ -0,0 +1,460 @@
...
+		listUsage()
+		os.Exit(1)
+	}
+	switch flag.Arg(1) {
+	case "users":
+		checkToken(token)
+		err := list.ListUsers(api_uri, token)
+		if err != nil {
+			fmt.Printf("Error: failed to get users, reason: %v\n", err)
+		}
```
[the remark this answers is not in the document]
nanjiangshu: In the current setup, different errors will be followed by different help messages so that the help message is more specific relate to the errors. If we return all errors to the main function, the whole logic need to be rewritten. 
I can return just this error message without disturbing the logic but feels meaningless if we just return this error message but not others to the main function. 
I'm inclined to keep as is.

sda-admin/file/file.go
nanjiangshu: It makes sense to me. I will create a commit to fix the issues of double slash.
However, if we will be the only user for `sda-admin`, I guess we don't need to spend too much effort for over engineering.

jbygdell: `We` are not the main target user of this app, the helpdesk staff and all repository managers in GDI are.

sda-admin/main.go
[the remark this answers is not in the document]
jbygdell: You need to return an error to the main thread so that `os.Exit(1)` can be called from there and not the subfunction.

sda-admin/file/file.go
nanjiangshu: I think I was wrong in the previous message. It seem `path.Join` always uses '/' as path separators, even on Windows. I have implemented changes in the commit 34e82e5

sda-admin/main.go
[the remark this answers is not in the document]
nanjiangshu: I want specific error message and usage help for each command, e.g. for the following logic 
```golang
	if filepath == "" || username == "" {
		fmt.Println("Error: both -filepath and -user are required.")
		fileIngestUsage()
		os.Exit(1)
	}
```
If I return the error message, in the main function, it is not easy to track it is `fileIngestUsage` that should be printed and in the order after the error message. It's doable but will be a bit more complicated. 
If you think the problem is that os.Exit() should only happen in the main function. I can instead return the exit status code and then make sure os.Exit() happens only in the main function.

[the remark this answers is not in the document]
jbygdell: Do you really need both the error message and the usage message?

Anyway this is a fully viable solution:
```
if filepath == "" || username == "" {
	fmt.Println("Error: both -filepath and -user are required.")
	fileIngestUsage()

	return fmt.Errorln("exit")
}
```

MalinAhlberg (commented):
Really well done, @nanjiangshu!
I added some minor comments. Also, shouldn't `.github/workflows/test.yml` be updated so that the tests are run?

viklund (changes requested):
I think commands should either be `<verb> <noun>` or `<noun> <verb>`. Currently there's a mix of both.

I suggest not having `list` as a toplevel command. So refactor `list files` and `list users` into `files list` and `users list`.

sda-admin/helpers/helpers.go
```diff
@@ -0,0 +1,166 @@
...
+	if !ok {
+		return fmt.Errorf("broken token (claims are empty): %v\nerror: %s", claims, err)
+	}
+
+	// Check if the token has exp claim
+	if claims["exp"] == nil {
+		return fmt.Errorf("could not parse token, reason: no expiration date")
+	}
+
+	// Parse the expiration date from token, handle cases where
```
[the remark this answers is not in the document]
nanjiangshu: Yes. In case of "bad token", the message is shown in the log but the user will not see it. It's better to check it on the client side as well.

sda-admin/helpers/helpers.go
```diff
@@ -0,0 +1,166 @@
...
+	"time"
+
+	"github.com/golang-jwt/jwt"
+)
+
+// Helper functions used by more than one module
+
+// CheckTokenExpiration is used to determine whether the token is expiring in less than a day
+func CheckTokenExpiration(accessToken string) error {
+	// Parse jwt token with unverifies, since we don't need to check the signatures here
```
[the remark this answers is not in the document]
nanjiangshu: Added in f090467

sda-admin/helpers/helpers.go
```diff
@@ -0,0 +1,166 @@
...
+		)
+	}
+
+	return nil
+}
+
+// necessary for mocking in unit tests
+var GetResponseBody = GetBody
+
+// GetBody gets the body of the response from the URL
```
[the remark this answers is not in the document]
nanjiangshu: Add in f090467

nanjiangshu:
> Really well done, @nanjiangshu! I added some minor comments. Also, shouldn't `.github/workflows/test.yml` be updated so that the tests are run?

Good catch @MalinAhlberg , I added in the commit ab22a90

nanjiangshu:
> I think commands should either be `<verb> <noun>` or `<noun> <verb>`. Currently there's a mix of both.
> 
> I suggest not having `list` as a toplevel command. So refactor `list files` and `list users` into `files list` and `users list`.

Good suggestion @viklund . The syntax of the commands is now consistent with the format <none> <verb>. I use `file` instead of `files` and `set-accession` instead of `session`. 
```
sda-admin user list
sda-admin file list
sda-admin file ingest
sda-admin file set-accession
sda-admin dataset create
sda-admin dataset release  
```

sda-admin/main.go
```diff
@@ -0,0 +1,524 @@
...
+	}
+
+	return nil
+}
+
+func handleUserCommand() error {
+	if flag.NArg() < 2 {
+		fmt.Fprint(os.Stderr, "Error: 'user' requires a subcommand (list).\n")
+		fmt.Fprint(os.Stderr, userUsage)
+		return fmt.Errorf("")
```
jbygdell, line 275:
```suggestion
		fmt.Fprint(os.Stderr, userUsage)
		return errors.New("error: 'user' requires a subcommand")
```

sda-admin/main.go
```diff
@@ -0,0 +1,524 @@
...
+			return err
+		}
+		err := user.List(apiURI, token)
+		if err != nil {
+			return fmt.Errorf("Error: failed to get users, reason: %v\n", err)
+		}
+	default:
+		fmt.Fprintf(os.Stderr, "Unknown subcommand '%s' for '%s'.\n", flag.Arg(1), flag.Arg(0))
+		fmt.Fprint(os.Stderr, userUsage)
+		return fmt.Errorf("")
```
jbygdell, line 289:
```suggestion
		fmt.Fprint(os.Stderr, userUsage)
		return fmt.Errorf("unknown subcommand '%s' for '%s'", flag.Arg(1), flag.Arg(0))
```

jbygdell (changes requested):
* Add this folder to the `code-linter` action
* Add an option in the `Makefile` to build the app, both separate and as part of the `build-all` step
* Add an option in the `Makefile` to run the linter for this folder

sda-admin/helpers/helpers.go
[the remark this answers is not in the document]
jbygdell: This is a CLI, it doesn't have logs. Either it is shown to the user or it is not relevant.

MalinAhlberg:
I tested this a bit, and I think it's a really nice way to work with the pipeline! Also, it seems to be a really good way to run into bugs, by being a maybe-not-so-organized user. I quickly ran into bug #1027, and also managed to run into something similar for verify/postgres. They keep logging even though I did not get any error from `sda-admin`:
```
postgres  | 2024-09-12 13:36:22.248 UTC [80] ERROR:  duplicate key value violates unique constraint "unique_checksum"
...
verify    | {"level":"error","msg":"SetVerified failed, reason: (pq: duplicate key value violates unique constraint \"unique_checksum\")","time":"2024-09-12T13:36:22Z"}
```
I'm not sure what I did wrong.
When trying to create a dataset with incorrect file IDs (which could happen if you for example mistype), no error is reported but I was not able to continue after that, all commands returned errors such as
```
Error: failed to [...], reason: failed to send request, reason: Post "http://localhost:8090/file/accession": EOF
```

sda-admin/helpers/helpers.go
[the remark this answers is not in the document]
nanjiangshu: I mean the logs of the container. The reason about the problem of bad token is only shown in the logs of the container and not returned from the API call. That's why `CheckTokenExpiration` needs to be called in the CLI.

sda-admin/main.go
```diff
@@ -0,0 +1,524 @@
...
+		case "dataset":
+			if err := handleHelpDataset(); err != nil {
+				return err
+			}
+		case "version":
+			fmt.Fprint(os.Stderr, versionUsage)
+		default:
+			fmt.Fprintf(os.Stderr, "Unknown command '%s'.\n", flag.Arg(1))
+			fmt.Fprint(os.Stderr, usage)
+			return fmt.Errorf("")
```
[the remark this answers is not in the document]
nanjiangshu: I think it is more natural to print `"Unknown command '%s'.\n"` first and then the `usage`. I will return the combined string with `"Unknown command '%s'.\n"` and `usage` and then print it in the main function. Does it make sense @jbygdell ?

sda-admin/helpers/helpers.go
```diff
@@ -0,0 +1,166 @@
...
+		fmt.Fprintln(
+			os.Stderr,
+			"The provided access token expires in",
+			time.Until(expiration).Truncate(time.Second),
+		)
+	}
+
+	return nil
+}
+
```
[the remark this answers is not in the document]
nanjiangshu: I think it is needed for more user friendly feedback to the user as I explained at https://github.com/neicnordic/sensitive-data-archive/pull/1026#discussion_r1756939763. Does it make sense?

sda-admin/helpers/helpers.go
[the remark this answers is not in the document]
MalinAhlberg: Aha, I see your point @nanjiangshu. It would be nice if the API call would return the status of the token, so that we don't have to duplicate that code. But I agree that the most important thing is that the users understand what's happening and gets an informative message if the token is bad.

jbygdell:
> I tested this a bit, and I think it's a really nice way to work with the pipeline! Also, it seems to be a really good way to run into bugs, by being a maybe-not-so-organized user. I quickly ran into bug #1027, and also managed to run into something similar for verify/postgres. They keep logging even though I did not get any error from `sda-admin`:
> 
> ```
> postgres  | 2024-09-12 13:36:22.248 UTC [80] ERROR:  duplicate key value violates unique constraint "unique_checksum"
> ...
> verify    | {"level":"error","msg":"SetVerified failed, reason: (pq: duplicate key value violates unique constraint \"unique_checksum\")","time":"2024-09-12T13:36:22Z"}
> ```
> 

All of that needs to be solved in the API. There are a few steps we need to double check.

sda-admin/helpers/helpers.go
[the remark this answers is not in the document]
jbygdell: Yes, the API has its short comings, but we shouldn't implement band aids here for things that needs to be fixed in the API.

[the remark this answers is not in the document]
jbygdell: in order for the user to know what went wrong you need to take action depending on the response you get back from the API.

No token, bad token and expired token will all generate a response status code of `401`.
The actual response body will differ depending on the case:
* No token: "no access token supplied"
* Bad token and expired token: message starts with "signed token not valid"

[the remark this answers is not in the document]
nanjiangshu: make sense. I have remove token checking from the CLI and show the response when server returned error. 
Now the error message looks informative 
```
error: failed to ingest file, reason: server returned status 401: {"error":"signed token not valid: invalid JWT, (token was 11)"}
```
implemented in the commits 575ddea and 7bfac1e

sda-admin/dataset/dataset.go
```diff
@@ -0,0 +1,59 @@
...
+	return nil
+}
+
+// Release releases a dataset for downloading
+func Release(apiURI, token, datasetID string) error {
+	parsedURL, err := url.Parse(apiURI)
+	if err != nil {
+		return err
+	}
+	parsedURL.Path = path.Join(parsedURL.Path, "dataset/release", datasetID)
```
jbygdell, line 49: This will only work if the datasetID is not an URL, which it will be.

jbygdell (changes requested):

MalinAhlberg:
> All of that needs to be solved in the API. There are a few steps we need to double check.

I created #1037 and #1038.

sda-admin/dataset/dataset.go
nanjiangshu: Good point! Fixed in e9e9ac9
I find also at `sda-cli` sometimes url concatenation is not handled properly with path.Join, e.g. https://github.com/NBISweden/sda-cli/blob/main/htsget/htsget.go#L102. Maybe needs some refactoring in the future.

sda-admin/main.go
```diff
@@ -0,0 +1,486 @@
...
+	if datasetID == "" || len(accessionIDs) == 0 {
+		return fmt.Errorf("Error: -dataset-id and at least one accession ID are required.\n%s", datasetCreateUsage)
+	}
+
+	err := dataset.Create(apiURI, token, datasetID, accessionIDs)
+	if err != nil {
+		return fmt.Errorf("error: failed to create dataset, reason: %v", err)
+	}
+
+	fmt.Println("Dataset created successfully.")
```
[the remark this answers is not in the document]
nanjiangshu: Although I committed the change, as a user, I feel a bit weird that the command does not return anything after running.

viklund (approved):

sda-admin/main.go
```diff
@@ -0,0 +1,479 @@
...
+
+Additional Commands:
+  help             Show this help message.
+  -h, -help        Show this help message.`
+
+var userUsage = `
+List Users:
+  Usage: sda-admin user list 
+    List all users in the system.
+`
```
jbygdell, line 50:
```suggestion
var userUsage = `List Users:
  Usage: sda-admin user list 
    List all users in the system with ongoing submissions.`
```

sda-admin/main.go
```diff
@@ -0,0 +1,479 @@
...
+
+var userListUsage = `
+Usage: sda-admin user list 
+  List all users in the system.
+`
```
jbygdell, line 55:
```suggestion
var userListUsage = `Usage: sda-admin user list 
  List all users in the system with ongoing submissions.`
```

sda-admin/main.go
```diff
@@ -0,0 +1,479 @@
...
+  Usage: sda-admin file set-accession -filepath FILEPATH -user USERNAME -accession-id ACCESSION_ID
+    Assign an accession ID to a file for a given user.
+
+Options:
+  -user USERNAME       Specify the username associated with the file.
+  -filepath FILEPATH   Specify the path of the file to ingest.
+  -accession-id ID     Specify the accession ID to assign to the file.
+
+Use 'sda-admin help file <command>' for information on a specific command.
+`
```
jbygdell, line 76:
```suggestion
var fileUsage = `List all files for a user:
  Usage: sda-admin file list -user USERNAME
	List all files for a specified user.

Ingest a file:
  Usage: sda-admin file ingest -filepath FILEPATH -user USERNAME
    Trigger the ingestion of a given file for a specific user.

Set accession ID to a file:
  Usage: sda-admin file set-accession -filepath FILEPATH -user USERNAME -accession-id ACCESSION_ID
    Assign an accession ID to a file for a given user.

Options:
  -user USERNAME       Specify the username associated with the file.
  -filepath FILEPATH   Specify the path of the file to ingest.
  -accession-id ID     Specify the accession ID to assign to the file.

Use 'sda-admin help file <command>' for information on a specific command.`
```

sda-admin/main.go
```diff
@@ -0,0 +1,479 @@
...
+
+var fileListUsage = `
+Usage: sda-admin file list -user USERNAME
+  List all files for a specified user.
+
+Options:
+  -user USERNAME 	Specify the username associated with the files.
+`
```
jbygdell, line 84:
```suggestion
var fileListUsage = `Usage: sda-admin file list -user USERNAME
  List all files for a specified user.

Options:
  -user USERNAME 	Specify the username associated with the files.`
```

sda-admin/main.go
```diff
@@ -0,0 +1,479 @@
...
+
+var fileIngestUsage = `
+Usage: sda-admin file ingest -filepath FILEPATH -user USERNAME
+  Trigger the ingestion of a given file for a specific user.
+
+Options:
+  -filepath FILEPATH   Specify the path of the file to ingest.
+  -user USERNAME       Specify the username associated with the file.
+`
```
jbygdell, line 93:
```suggestion
var fileIngestUsage = `Usage: sda-admin file ingest -filepath FILEPATH -user USERNAME
  Trigger the ingestion of a given file for a specific user.

Options:
  -filepath FILEPATH   Specify the path of the file to ingest.
  -user USERNAME       Specify the username associated with the file.`
```

sda-admin/main.go
```diff
@@ -0,0 +1,479 @@
...
+
+var fileAccessionUsage = `
+Usage: sda-admin file set-accession -filepath FILEPATH -user USERNAME -accession-id ACCESSION_ID
+  Assign accession ID to a file and associate it with a user.
+
+Options:
+  -filepath FILEPATH   Specify the path of the file to assign the accession ID.
+  -user USERNAME       Specify the username associated with the file.
+  -accession-id ID     Specify the accession ID to assign to the file.
+`
```
jbygdell, line 103:
```suggestion
var fileAccessionUsage = `Usage: sda-admin file set-accession -filepath FILEPATH -user USERNAME -accession-id ACCESSION_ID
  Assign accession ID to a file and associate it with a user.

Options:
  -filepath FILEPATH   Specify the path of the file to assign the accession ID.
  -user USERNAME       Specify the username associated with the file.
  -accession-id ID     Specify the accession ID to assign to the file.`
```

sda-admin/main.go
```diff
@@ -0,0 +1,479 @@
...
+Release a dataset:
+  Usage: sda-admin dataset release -dataset-id DATASET_ID
+    Release a dataset for downloading based on its dataset ID.
+
+Options:
+  -dataset-id DATASET_ID   Specify the unique identifier for the dataset.
+  [ACCESSION_ID ...]       (For dataset create) Specify one or more accession IDs to include in the dataset.
+
+Use 'sda-admin help dataset <command>' for information on a specific command.
+`
```
jbygdell, line 119:
```suggestion
var datasetUsage = `Create a dataset:
  Usage: sda-admin dataset create -dataset-id DATASET_ID [ACCESSION_ID ...]
    Create a dataset from a list of accession IDs and a dataset ID.
    
Release a dataset:
  Usage: sda-admin dataset release -dataset-id DATASET_ID
    Release a dataset for downloading based on its dataset ID.

Options:
  -dataset-id DATASET_ID   Specify the unique identifier for the dataset.
  [ACCESSION_ID ...]       (For dataset create) Specify one or more accession IDs to include in the dataset.

Use 'sda-admin help dataset <command>' for information on a specific command.`
```

sda-admin/main.go
```diff
@@ -0,0 +1,479 @@
...
+
+var datasetCreateUsage = `
+Usage: sda-admin dataset create -dataset-id DATASET_ID [ACCESSION_ID ...]
+  Create a dataset from a list of accession IDs and a dataset ID.
+
+Options:
+  -dataset-id DATASET_ID    Specify the unique identifier for the dataset.
+  [ACCESSION_ID ...]         (For dataset create) Specify one or more accession IDs to include in the dataset.
+`
```
jbygdell, line 128:
```suggestion
var datasetCreateUsage = `Usage: sda-admin dataset create -dataset-id DATASET_ID [ACCESSION_ID ...]
  Create a dataset from a list of accession IDs and a dataset ID.

Options:
  -dataset-id DATASET_ID    Specify the unique identifier for the dataset.
  -acession-ids [ACCESSION_ID ...]         (For dataset create) Specify one or more accession IDs to include in the dataset.`
```

sda-admin/main.go
```diff
@@ -0,0 +1,479 @@
...
+
+var datasetReleaseUsage = `
+Usage: sda-admin dataset release -dataset-id DATASET_ID
+  Release a dataset for downloading based on its dataset ID.
+
+Options:
+  -dataset-id DATASET_ID    Specify the unique identifier for the dataset.
+`
```
jbygdell, line 136:
```suggestion
var datasetReleaseUsage = `Usage: sda-admin dataset release -dataset-id DATASET_ID
  Release a dataset for downloading based on its dataset ID.

Options:
  -dataset-id DATASET_ID    Specify the unique identifier for the dataset.`
```

sda-admin/main.go
```diff
@@ -0,0 +1,479 @@
...
+
+var versionUsage = `
+Usage: sda-admin version
+  Show the version information for sda-admin.
+`
```
jbygdell, line 141:
```suggestion
var versionUsage = `Usage: sda-admin version
  Show the version information for sda-admin.`
```

sda-admin/main.go
```diff
@@ -0,0 +1,479 @@
...
+		return fmt.Errorf("unknown command '%s'.\n%s", flag.Arg(1), usage)
+	}
+
+	return nil
+}
+
+func handleHelpUser() error {
+	switch {
+	case flag.NArg() == 2:
+		fmt.Fprint(os.Stderr, userUsage)
```
jbygdell, line 219:
```suggestion
		fmt.Println(userUsage)
```

sda-admin/main.go
```diff
@@ -0,0 +1,479 @@
...
+	case flag.Arg(2) == "list":
+		fmt.Fprint(os.Stderr, userListUsage)
```
jbygdell, line 221:
```suggestion
		fmt.Println(userListUsage)
```

sda-admin/main.go
```diff
@@ -0,0 +1,479 @@
...
+
+	return nil
+}
+
+func handleHelpFile() error {
+	switch {
+	case flag.NArg() == 2:
+		fmt.Fprint(os.Stderr, fileUsage)
+	case flag.Arg(2) == "list":
+		fmt.Fprint(os.Stderr, fileListUsage)
```
jbygdell, line 234:
```suggestion
		fmt.Println(fileListUsage)
```

jbygdell (changes requested):
In order to test that the messages arrive in the correct stream (stderr or stdout) you can use this: `/sda-admin -uri example.com -token fooBar help dataset 2>stderr.log 1>stdout.log`

sda-admin/main.go
```diff
@@ -0,0 +1,479 @@
...
+	case flag.Arg(2) == "ingest":
+		fmt.Fprint(os.Stderr, fileIngestUsage)
+	case flag.Arg(2) == "set-accession":
+		fmt.Fprint(os.Stderr, fileAccessionUsage)
```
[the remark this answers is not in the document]
nanjiangshu: Hi @jbygdell , do you mean usage should be output to stdout and only error message should be output to stderr? Then I probably misunderstood you previously. I actually also think usage info should be output to STDOUT, especially when the exit code is 0. 

By the way, in `sda-cli`, it seems usage info is output to stderr.

sda-admin/dataset/dataset.go
```diff
@@ -0,0 +1,59 @@
...
+// Release releases a dataset for downloading
+func Release(apiURI, token, datasetID string) error {
+	parsedURL, err := url.Parse(apiURI)
+	if err != nil {
+		return err
+	}
+	parsedURL.Path = path.Join(parsedURL.Path, "dataset/release") + "/" + datasetID
+
+	jsonBody := []byte("{}")
+	_, err = helpers.PostRequest(parsedURL.String(), token, jsonBody)
```
jbygdell, line 52:
```suggestion
	_, err = helpers.PostRequest(parsedURL.String(), token, nil)
```

sda-admin/main.go
```diff
@@ -0,0 +1,455 @@
...
+
+	// Parse global flags first
+	flag.Parse()
+
+	// If no command is provided, show usage
+	if flag.NArg() == 0 {
+		return errors.New(usage)
+	}
+
+	if flag.Arg(0) == "help" {
```
jbygdell, line 144:
```suggestion
	if flag.Arg(0) == "help" || flag.Arg(0) == "version" {
```

Version is also a special case.

sda-admin/main.go
```diff
@@ -0,0 +1,455 @@
...
+  Release a dataset for downloading based on its dataset ID.
+
+Options:
+  -dataset-id DATASET_ID    Specify the unique identifier for the dataset.`
+
+var versionUsage = `Usage: sda-admin version
+  Show the version information for sda-admin.`
+
+func printVersion() {
+	fmt.Printf("sda-admin version %s\n", version)
```
jbygdell, line 123:
```suggestion
	fmt.Printf("sda-admin version %s\n", version)
```
```suggestion
	fmt.Printf("sda-admin %s\n", version)
```

sda-admin/main.go
```diff
@@ -0,0 +1,455 @@
...
+
+var datasetReleaseUsage = `Usage: sda-admin dataset release -dataset-id DATASET_ID
+  Release a dataset for downloading based on its dataset ID.
+
+Options:
+  -dataset-id DATASET_ID    Specify the unique identifier for the dataset.`
+
+var versionUsage = `Usage: sda-admin version
+  Show the version information for sda-admin.`
+
```
jbygdell, line 121:
```suggestion

```

Do you really need a help entry for displaying the version??

sda-admin/main.go
```diff
@@ -0,0 +1,455 @@
...
+	case "file":
+		if err := handleHelpFile(); err != nil {
+			return err
+		}
+	case "dataset":
+		if err := handleHelpDataset(); err != nil {
+			return err
+		}
+	case "version":
+		fmt.Println(versionUsage)
```
jbygdell, line 187:
```suggestion
```

jbygdell (changes requested):

sda-admin/dataset/dataset.go
nanjiangshu: Since PostRequest is defined as 
```golang
func PostReq(url, token string, jsonBody []byte) ([]byte, error) {
```
isn't it better to specify the third parameter as `[]byte("{}")` than `nil`?

jbygdell: That's because how you defined the function.
An empty byte slice is the same as `nil` and since the dataset release doesn't require any payload you can just as well omit it by sending`nil` here.

nanjiangshu: OK. fixed in the commit 95c6bf9

jbygdell (approved):

viklund (approved):
⭐

PR #1028, opened 2024-09-09
[linter] Exclude `G115 - integer overflow`

jbygdell:
When the language has no solution to check the overflow in time of convert, so the linter shouldn't check the overflow.

**Description**

**How to test**


aaperis:
It would be useful to add manual checks for this as in #1001 before telling the linter to ignore these checks.

aaperis (approved):

MalinAhlberg:
One could argue that the linter checking this is important just because the language itself does not...? I have no strong opinion, though.

jbygdell:
> One could argue that the linter checking this is important just because the language itself does not...? I have no strong opinion, though.

The problem is that there is no way to satisfy the linter at this time.
Even if you add a check for integer overflow the linter will still complain.

MalinAhlberg (approved):

PR #1073, opened 2024-10-04
Ingest adds keyhash to files db table

MalinAhlberg:
**Related issue(s) and PR(s)**  
This PR closes #893.
It will be undrafted when #1036 is closed.

Commit [tmp: temporary fixes to api/hash](ceecb05)  will not be included in the final version of this PR.

**Description**
- ingest hex encodes the user's public key and adds it to the files info in the `sda.files` table
- if the keyhash is not in the `sda.encryption_keys`, an error is thrown and ingest stops
- code test and integration test are added

**How to test**
Add the hashed public key to `sda.encryption_keys` and then try to ingest a file. See the [integration test](28be1c) for details.

PR #1080, opened 2024-10-12
[API] Add sanity checks when releasing a dataset. 

jbygdell:
**Related issue(s) and PR(s)**  
This PR closes #1017.

**Description**
This adds some sanity checking when releasing a dataset
* Ensure dataset is registered before releasing
* Ensure dataset is not deprecated when releasing
* Ensure dataset is not already released when releasing

**How to test**

PR #1084, opened 2024-10-14
Feature/api keys

### LINKED ISSUE #1036: Extend admin API for adding key hash
As a sda-dev I want to use the admin API for adding the key hash in the keys table
so that to keep track of keys for the key rotation.

A/C:
- Store the hex encoded public key with the additional info in the encryption keys table
- Description should be supplied
- Optional created date

Out of scope:
Update admin cli

kostas-kou:
This PR closes #1036.

**Description**
With this PR an endpoint for adding key hash in added in the api service.
An extra step of adding an api role in the database was necessary in order to make it work.

**Comment**
There is no integration test for this feature because Malin has already created one and it will be [in this PR](https://github.com/neicnordic/sensitive-data-archive/pull/1073) which is on hold because of the current.

**How to test**
From the root of the folder run `make build-all`
and then for linux run:
```bash
PR_NUMBER=$(/usr/bin/date '+%F')  docker compose -f .github/integration/sda-s3-integration.yml run integration_test
```
while for mac run:
```bash
PR_NUMBER=$(/bin/date '+%F')  docker compose -f .github/integration/sda-s3-integration.yml run integration_test
```

Then set the token as envvar: `export ACCESS_TOKEN=$(curl -s -k http://localhost:8080/tokens | jq -r '.[0]')`
and run:
```bash
curl -H "Authorization: Bearer $ACCESS_TOKEN" -H "Content-Type: application/json" -X POST -d '{"hash": "this-is-just-a-hash", "description": "this is keys description"}' http://localhost:8090/key/hashed
```

Confirm that the key hash is in the database by running: `docker exec -it postgres psql -U postgres -d sda` and then the query `select * from sda.encryption_keys;`


postgresql/initdb.d/04_grants.sql
```diff
@@ -158,10 +158,29 @@ GRANT SELECT ON local_ega_ebi.file_dataset TO download;
 
 --------------------------------------------------------------------------------
 
+CREATE ROLE api;
+
+GRANT USAGE ON SCHEMA sda TO api;
+GRANT SELECT ON sda.files TO api;
+GRANT SELECT ON sda.file_dataset TO api;
+GRANT SELECT ON sda.checksums TO api;
+GRANT SELECT ON sda.file_event_log TO api;
```
jbygdell, line 167: API also needs to be able to read the datasets and the statuses of the datasets see #1080

postgresql/migratedb.d/13.sql
```diff
@@ -0,0 +1,56 @@
...
+        END IF;
+    END;
+    $created$ LANGUAGE plpgsql;
+
+    PERFORM create_role_if_not_exists('api');
+    GRANT USAGE ON SCHEMA sda TO api;
+    GRANT SELECT ON sda.files TO api;
+    GRANT SELECT ON sda.file_event_log TO api;
+    GRANT SELECT ON sda.file_dataset TO api;
+    GRANT SELECT  ON sda.checksums TO api;
```
jbygdell, line 39: See comment above

sda/cmd/api/api.go
```diff
@@ -80,6 +82,7 @@ func setup(config *config.Config) *http.Server {
 		r.POST("/file/accession", isAdmin(), setAccession)             // assign accession ID to a file
 		r.POST("/dataset/create", isAdmin(), createDataset)            // maps a set of files to a dataset
 		r.POST("/dataset/release/*dataset", isAdmin(), releaseDataset) // Releases a dataset to be accessible
+		r.POST("/key/hashed", isAdmin(), addHashedKey)                 // Adds a hashed key to the database
```
jbygdell, line 85: This should follow the existing semantics.
```suggestion
		r.POST("/c4gh-keys/add", isAdmin(), addKeyHash)                 // Adds a key hash to the database
```

sda/cmd/api/api.go
```diff
@@ -389,3 +392,53 @@ func listUserFiles(c *gin.Context) {
...
 	c.JSON(200, files)
 }
+
+// addHashedKey handles the addition of a hashed key to the database.
+// It expects a JSON payload containing the key hash and its description.
+// If the JSON payload is invalid, it responds with a 400 Bad Request status.
+// If there is no row update in the database, it responds with a 409 Conflict status
+// If the database insertion fails, it responds with a 500 Internal Server Error status.
+// On success, it responds with a 200 OK status.
+func addHashedKey(c *gin.Context) {
```
jbygdell, line 402:
```suggestion
func addC4ghKey(c *gin.Context) {
```

We don't want to calculate the keyhash manually as it will be error prone.
We should send the entire public key and have the API do the hashing.

sda/internal/schema/schema.go
```diff
@@ -179,3 +179,8 @@ type SyncMetadata struct {
 type Metadata struct {
 	Metadata interface{}
 }
+
+type KeyhashInsertion struct {
+	Hash        string `json:"hash"`
```
jbygdell, line 184:
```suggestion
	C4ghPubKey        string `json:"c4ghPubKey"`
```
This should not be the key hash but the full public key as a base64 encoded string, i.e `cat c4gh.pub ¦ base64 -w0`
We should also do validation so that we don't  accept things like empty strings.

sda/cmd/api/api.md
```diff
@@ -119,6 +119,24 @@ Admin endpoints are only available to a set of whitelisted users specified in th
...
+    - `200` Query execute ok.
+    - `400` Error due to bad payload.
+    - `401` Token user is not in the list of admins.
+    - `409` Key hash already exists in the database.
+    - `500` Internal error due to DB failures.
+
+    Example:
+
+    ```bash
+    curl -H "Authorization: Bearer $token" -H "Content-Type: application/json" -X POST -d '{"hash": "key-hex-hash", "description": "this is the key description"}' https://HOSTNAME/kes/hashed
```
jbygdell, line 136:
```suggestion
    curl -H "Authorization: Bearer $token" -H "Content-Type: application/json" -X POST -d '{"c4ghPubKey": "'"$( base64 /PATH/TO/c4gh.pub)"'", "description": "this is the key description"}' https://HOSTNAME/c4gh-keys/add
```

sda/cmd/api/api_test.go
```diff
@@ -1186,3 +1187,62 @@ func (suite *TestSuite) TestListUserFiles() {
...
+	req, err := http.NewRequest("POST", ts.URL+"/key/hashed", bytes.NewBuffer(body))
+	assert.NoError(suite.T(), err)
+	req.Header.Add("Authorization", "Bearer "+suite.Token)
+	req.Header.Add("Content-Type", "application/json")
+
+	resp, err := client.Do(req)
+	assert.NoError(suite.T(), err)
+	assert.Equal(suite.T(), http.StatusBadRequest, resp.StatusCode)
+	defer resp.Body.Close()
+}
```
jbygdell, line 1248: You probably want at least 2 negative test, one for an empty string and one where the data is not a public key.

sda/cmd/api/api.go
```diff
@@ -389,3 +392,53 @@ func listUserFiles(c *gin.Context) {
...
+				"error":  "json decoding : " + err.Error(),
+				"status": http.StatusBadRequest,
+			},
+		)
+
+		log.Errorf("Invalid JSON payload: %v", err)
+
+		return
+	}
+
```
jbygdell, line 417: At this point we should do validation of the payload data i.e ensure that it is a c4gh public key before we hash it and store it in the database.

jbygdell (changes requested):

sda/cmd/api/api.go
```diff
@@ -414,3 +417,53 @@ func listUserFiles(c *gin.Context) {
 	c.Writer.Header().Set("Content-Type", "application/json")
 	c.JSON(200, files)
 }
+
+// addHashedKey handles the addition of a hashed key to the database.
+// It expects a JSON payload containing the key hash and its description.
+// If the JSON payload is invalid, it responds with a 400 Bad Request status.
+// If there is no row update in the database, it responds with a 409 Conflict status
```
MalinAhlberg, line 424:
```suggestion
// If the key is already in the database, it responds with a 409 Conflict status
```

MalinAhlberg (commented):
Good work! I agree with @jbygdell that it would be very nice, and much more user friendly, if the user could just send the public key and then the api would calculate the hex.
Apart from that it looks good and works with the mentioned tests coming in #1073 .

.github/integration/sda-s3-integration.yml
```diff
@@ -292,8 +292,8 @@ services:
       - BROKER_PASSWORD=ingest
       - BROKER_USER=ingest
```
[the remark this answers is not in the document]
nanjiangshu: I think we should change the value for the BROKER as well for consistency and debugging purpose.

sda/cmd/api/api.go
```diff
@@ -414,3 +417,53 @@ func listUserFiles(c *gin.Context) {
 	c.Writer.Header().Set("Content-Type", "application/json")
 	c.JSON(200, files)
 }
+
+// addHashedKey handles the addition of a hashed key to the database.
```
nanjiangshu, line 421:
```suggestion
// addHashedKey handles the addition of a key hash to the database.
```

sda/cmd/api/api.go
nanjiangshu: But this function will receive the public key and do the hash calculation and then insert the hash to the database, right? In this case, I think we can still call it `addHashedKey` or `addKeyHash`.

nanjiangshu (commented):
Great work! Works well when I tested and I just left a few minor comments.

Apart from that, I find that when trying to insert an already existing key, it takes more than 60 seconds to return the response. I guess this is caused by repeated retrying to insert the key? Would be great if the check of the existence a key can be done at an earlier stage so that the user don't need to wait 60 seconds to get the error message.

MalinAhlberg:
> it takes more than 60 seconds to return the response.

Great catch @nanjiangshu! I was wondering why the integrations tests were even slower than usual :sweat_smile:

jbygdell (dismissed):

MalinAhlberg (approved):
Great!

nanjiangshu (approved):
Looks good!

PR #1108, opened 2024-10-29
[SDA-ADMIN] Add key hash handling

jbygdell:
**Description**
This PR builds on #1107 and implements the key hash handling for the admin cli

**How to test**
```
make build-all
PR_NUMBER=$(date +%F) docker compose -f .github/integration/sda-s3-integration.yml run integration_test

token="$(curl http://localhost:8080/tokens | jq -r '.[0]')"
export ACCESS_TOKEN=$token

# read the instructions
sda-admin/sda-admin help c4gh-hash

# list all key hashes in the setup
sda-admin/sda-admin c4gh-hash list
```


MalinAhlberg (commented):
Good work, and :star: for good testing description.
I think this branch should be based on `feature/api-keyhash-actions`, since it won't work without those changes.
Also, as a user, it would be nice to have the possibility to re-add/de-deprecate a key or at least modify the description of a key. (Use case: you add a key but forget the description. You try to deprecate the key, so that you can re-add it, but even if a key is deprecated, you can't add it again). But I'm fine with that not being part of this PR.

I added some minor suggestions, will approve when they have been resolved.

MalinAhlberg (approved):

MalinAhlberg (approved):

PR #1115, opened 2024-11-04
[API] Add a RBAC solution

jbygdell:
**Description**
A RBAC solution for the admin API.

RBAC is configured through a JSON structure as shown below:

Each endpoint needs an entry or access will be disallowed, a role entry of "*" in the `policy` section will allow access to any user with a valid token. The `action` can be a single entry GET/POST/PUT or several separate by `|` as per example.

Specific roles are mapped to the `roles` section, where the role is either a role name or a username. the rolebinding allows  a role to function as a group where multiple roles/users can be bound to the same role. This simplifies the policy section as only the top level role needs to be specified for an endpoint

```json
{
   "policy": [
      {
         "role": "admin",
         "path": "/c4gh-keys/*",
         "action": "(GET)|(POST)|(PUT)"
      },
      {
         "role": "submission",
         "path": "/file/*",
         "action": "POST"
      },
      {
         "role": "*",
         "path": "/files",
         "action": "GET"
      }
   ],
   "roles": [
      {
         "role": "admin",
         "rolebinding": "submission"
      },
      {
         "role": "requester@demo.org",
         "rolebinding": "admin"
      },
      {
         "role": "dummy@example.com",
         "rolebinding": "submission"
      }
   ]
}
```


viklund (changes requested):
Role names should not be hardcoded.

It should be possible to define new roles without changing the go source code, and what endpoints they have access to.

jbygdell:
> Role names should not be hardcoded.
> 
> It should be possible to define new roles without changing the go source code, and what endpoints they have access to.

That would make this extremely complex to almost no gain.
You better come up with some reasonable scenarios where this will be beneficial before we spend the time to implement it.

viklund:
> > Role names should not be hardcoded.
> > It should be possible to define new roles without changing the go source code, and what endpoints they have access to.
> 
> That would make this extremely complex to almost no gain. You better come up with some reasonable scenarios where this will be beneficial before we spend the time to implement it.

The code in this file should not become more complicated (probably less). It's basically one annotation per endpoint. RBAC is a solved problem.

sda/cmd/api/api.md
```diff
@@ -136,16 +136,69 @@ Admin endpoints are only available to a set of whitelisted users specified in th
     curl -H "Authorization: Bearer $token" -H "Content-Type: application/json" -X POST -d '{"pubkey": "'"$( base64 -w0 /PATH/TO/c4gh.pub)"'", "description": "this is the key description"}' https://HOSTNAME/c4gh-keys/add
     ```
 
-#### Configure Admin users
+#### Configure RBAC
 
-The users that should have administrative access can be set in two ways:
+RBAC is configure according to the JSON schema below.
```
MalinAhlberg, line 141:
```suggestion
RBAC is configured according to the JSON schema below.
```

sda/cmd/api/api.md
```diff
@@ -136,16 +136,69 @@ Admin endpoints are only available to a set of whitelisted users specified in th
...
+The path to the JSON file containing the RBAC policies needs to be passed through the `api.rbacFile` config definition.
 
-- As a comma separated list of user identifiers assigned to: `admin.users`.
-- As a JSON file containg a list of the user identities, the path to the file is assigned to: `admin.usersFile`. This is the recommended way.
+The `policy` section will configure access to the defined endpoints, unless specific rules are set, an endpoint will not be accessible
```
MalinAhlberg, line 144:
```suggestion
The `policy` section will configure access to the defined endpoints. Unless specific rules are set, an endpoint will not be accessible
```

sda/cmd/api/api.md
```diff
@@ -136,16 +136,69 @@ Admin endpoints are only available to a set of whitelisted users specified in th
...
+
+- `action`: can be single string value i,e `GET` or a regex string with `|` as separator i.e. `(GET)|(POST)|(PUT)`. In the later case all actions in the list are allowed.
+- `paht`: have two different wildcard notations `*`, matches any value and `:` that matches a specifc named value
```
MalinAhlberg, line 147:
```suggestion
- `path`: the endpoint. Should be a string value with two different wildcard notations: `*` matches any value and `:` that matches a specific named value
```

sda/cmd/api/api.md
```diff
@@ -136,16 +136,69 @@ Admin endpoints are only available to a set of whitelisted users specified in th
...
+- `role`: is the role that will be able to access the path, `"*"` will match any role or user.
```
MalinAhlberg, line 148:
```suggestion
- `role`: the role that will be able to access the path, `"*"` will match any role or user.
```

MalinAhlberg (commented):
Looks good! I added some minor suggestions for the documentaiton. Will do some testing before I approve.

MalinAhlberg (approved):
Good work!

sda/cmd/api/api.md
```diff
@@ -136,16 +136,69 @@ Admin endpoints are only available to a set of whitelisted users specified in th
...
-#### Configure Admin users
+#### Configure RBAC
 
-The users that should have administrative access can be set in two ways:
+RBAC is configured according to the JSON schema below.
+The path to the JSON file containing the RBAC policies needs to be passed through the `api.rbacFile` config definition.
 
-- As a comma separated list of user identifiers assigned to: `admin.users`.
-- As a JSON file containg a list of the user identities, the path to the file is assigned to: `admin.usersFile`. This is the recommended way.
+The `policy` section will configure access to the defined endpoints. Unless specific rules are set, an endpoint will not be accessible
```
nanjiangshu, line 144:
```suggestion
The `policy` section will configure access to the defined endpoints. Unless specific rules are set, an endpoint will not be accessible.
```

nanjiangshu (approved):
Looks good!

MalinAhlberg (approved):

PR #1129, opened 2024-11-15
Update release_sda-admin.yaml

jbygdell:
Single and double quotes are handled differently on path parsing.

This action should *only* be run when the version file is updated.

### DIFF
--- a/.github/workflows/release_sda-admin.yaml
+++ b/.github/workflows/release_sda-admin.yaml
@@ -4,7 +4,7 @@ on:
     merge_group:
     pull_request:
       paths:
-        - 'sda-admin/.version'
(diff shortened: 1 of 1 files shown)


.github/workflows/release_sda-admin.yaml
```diff
@@ -4,7 +4,7 @@ on:
     merge_group:
     pull_request:
       paths:
-        - 'sda-admin/.version'
+        - "sda-admin/.version"
```
nanjiangshu, line 7: But for the string `sda-admin/.version`, there shouldn't be any difference with the single quote and double quote regarding path parsing, right?

nanjiangshu (approved):

PR #1150, opened 2024-11-27
Rework integration-test setup

jbygdell:
**Description**
Given the ongoing work with the admin API and the fact that the sync service needs a full rework some changes to the tests are required.

* Split out sync test to a separate docker compose file
* Only run the sync test if the code in the `cmd/sync*` subfolders change
* Expand the `Makefile` so that it will be easier to run the tests without having to rebuild the containers or run tests not related to the ongoing work (posix, s3, sync)

**Notice**
As part of this the release of version tagged artifacts, containers and charts are disabled, so we don't release anything that is not working properly.
This will be resolved when we move over to using release branches.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,11 +76,30 @@ integrationtest-postgres: build-postgresql
 integrationtest-rabbitmq: build-rabbitmq build-sda
 	@PR_NUMBER=$$(date +%F) docker compose -f .github/integration/rabbitmq-federation.yml run federation_test
 	@PR_NUMBER=$$(date +%F) docker compose -f .github/integration/rabbitmq-federation.yml down -v --remove-orphans
-integrationtest-sda: build-all
-	@PR_NUMBER=$$(date +%F) docker compose -f .github/integration/sda-s3-integration.yml run integration_test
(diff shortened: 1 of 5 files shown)


pahatz (approved):
This looks fine.

Makefile
```diff
@@ -76,11 +76,30 @@ integrationtest-postgres: build-postgresql
 integrationtest-rabbitmq: build-rabbitmq build-sda
 	@PR_NUMBER=$$(date +%F) docker compose -f .github/integration/rabbitmq-federation.yml run federation_test
 	@PR_NUMBER=$$(date +%F) docker compose -f .github/integration/rabbitmq-federation.yml down -v --remove-orphans
-integrationtest-sda: build-all
```
MalinAhlberg, line 79: could it make sense to keep the target `integrationtest-sda` as an way to run all integration tests? s3, posix and sync.

MalinAhlberg (approved):
Looks very good! I just have one question.

MalinAhlberg (approved):
:+1:

pahatz (approved):

Issue #1123, opened 2024-11-11
Create and test SQL-statements for updating the database to remove user-ids from the filepaths


viklund:
Once the services has been updated, legacy files has to be updated to reflect how the new stack works.

This needs to be thourougly tested. For example, use the integration tests and test before and after upgrade with old and new containers. Also make sure the download API works by fetching a file and listing in the end.

A/C:

- [x] A document with a SQL statement or similar that one can run on the production database to update the data.
- [ ] we will do some more checks to catch any more potential edge cases for the query

kjellp:
Sounds like this will be a possible breaking change. Is the plan to have the services (ingest pipeline and downliad/doa) supporting old and new style for some time? Or do "a hard shift" at an agreed point in time?

jbygdell:
@kjellp this is related to #1097.

Ingest uses the filepath in the message not what's in the database, and it is your inbox that sends that message.
So the services primarily affected by this will be the API + data out components, since they reads the file paths from the DB.

That being said, we should probably start using separate tracks (v1, v2 ...) for this kind of stuff since updating a component prematurely will cause big issues.

pahatz:
It seems there are some other places in sda that we hadn't thought about and making the switch with the SQL query will warrant some parts not working. This needs to be put on hold until we have all the changes in place.
 Next step is to create an issue for those.

pahatz:
We will move this in a readme somewhere

Issue #1095, opened 2024-10-23
Create a list of download API calls that leak the user id


kostas-kou:
As a developer
I want to investigate when the user id is exposed (filepath) when the download APIs are used
so that we know exactly where this is happening and fix it.

A/C:
- Create a list with the findings
  - list of the API calls that give user ids to the user.

nanjiangshu:
List of the API calls that shows the user ids 
# list of endpoints
1. sda-download
    1.1 `GET /metadata/datasets/*dataset/files`
2. admin-api
    2.1 `GET /files`
    2.2 `GET /users/:username/files`

PR #1161, opened 2024-12-03
Remove user id from filepath

### LINKED ISSUE #1097: Modify filepath not storing the user id part (to be continued)
### Note that changes to the first step makes all other things break, so this has to be kept on a separate branch until it can be merged into the master branch.

As a developer
I want to change the way the filepath is stored in database (remove the user id part)
so that the user is not part of the filepath

### Note

Changes required in these places:
- s3inbox should not store the userid as part of the filename in the database
  - the message to cega should not include the userid in the filepath (no message should include the userid in the filepath).
- ingest needs to be able to find the file, so it has to concatenate the userid
- mapper needs to be able to remove the file from the inbox
- download - no change if it's fixed above
  - s3
  - encrypted s3
  - decrypted REST

MalinAhlberg:
This PR closes #1097 .

**Description**
The user's id is no longer stored in the database, so that it will never be shown to a user of download.
- filepaths in integration tests are updated to reflect this change
- inbox does not store the user id
- ingest and mapper do not expect the user id to be there

**How to test**
- all tests look ok and pass
- the endpoints in #1095 do not leak user information
- download still works

**Note**
The db schema version from #1123 should be set, but potentially in another PR.
The base of this PR is `feature/do-not-leak-user-ids`, which is branched off main. I do not feel strongly about the naming of this branch, so feel totally free to have opinions.


jbygdell (approved):

.github/integration/tests/sda/60_api_admin_test.sh
```diff
@@ -104,27 +101,18 @@ fi
...
 s3cmd -c s3cfg put NA12878.bam.c4gh s3://test_dummy.org/NE12878.bam.c4gh
+echo "waiting for upload to complete"
+sleep 10
 
-URI=http://rabbitmq:15672
-if [ -n "$PGSSLCERT" ]; then
-   URI=https://rabbitmq:15671
+if [ $((stream_size++)) -eq "$(curl -s -u guest:guest $URI/api/queues/sda/inbox | jq '.messages_ready')" ]; then
+   echo "Upload did not complete successfully"
+   exit 1
```
aaperis, line 109: If it randomly takes more than `sleep 10`, will this fail randomly as well?

sda/cmd/mapper/mapper.go
```diff
@@ -102,14 +103,14 @@ func main() {
 				}
 
 				for _, aID := range mappings.AccessionIDs {
-					log.Debugf("Mapped file to dataset (corr-id: %s, datasetid: %s, accessionid: %s)", delivered.CorrelationId, mappings.DatasetID, aID)
```
aaperis, line 105: why remove these logs?

aaperis (commented):
Looks good! Some minor comments/questions :-)

.github/integration/tests/sda/60_api_admin_test.sh
```diff
@@ -41,25 +41,25 @@ fi
...
+if [ $((stream_size++)) -eq "$(curl -s -u guest:guest http://rabbitmq:15672/api/queues/sda/inbox | jq '.messages_ready')" ]; then
    echo "waiting for upload to complete"
-   RETRY_TIMES=$((RETRY_TIMES + 1))
-   if [ "$RETRY_TIMES" -eq 30 ]; then
-      echo "::error::Time out while waiting for upload to complete"
-      exit 1
-   fi
-   sleep 2
-done
+   echo "but too bad will eexit"
```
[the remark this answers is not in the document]
aaperis: resolved :-)

.github/integration/tests/sda/60_api_admin_test.sh
MalinAhlberg: Reworked in 3353a1f

sda/cmd/mapper/mapper.go
MalinAhlberg: By accident :face_with_peeking_eye: . Good that you noticed. Readded in e263eb3.

aaperis (approved):
Nice work!

PR #1163, opened 2024-12-04
Configurable s3 endpoint for download

### LINKED ISSUE #750: [download] Configurable option for whether to support decrypted xor encrypted downloads.
As a downloader,
I want a clear and simple interface for downloading files,
so that it's easier to use.

### Description

The API for getting files from the repository using the `download` service is currently through `/s3` or `/s3-encrypted` depending on whether one is eligable to getting plain or re-encrypted files. There should be only one endpoint `/s3` and then depending on if a crypt4gh key pair is provided or not, should serve decrypted or encrypted files respectively.

Errors should be given to the user:
- [x] In case a user supplies a public key and the endpoint is configured to return plain files an error should be generated.
```sh
$ curl -s --cacert certs/ca.pem -H "Authorization: Bearer $token" -H "Client-Public-Key: $clientkey" "https://localhost:9443/s3/$dataset/$file"
downloading encrypted data is not supported
```
- [x] In case a user forgets to supply a public key and the endpoint is configure to return re-encrypted files.
```sh
$ curl -s --cacert certs/ca.pem -H "Authorization: Bearer $token" "https://localhost:8443/s3/$dataset/$file" 
c4gh public key is missing from the header
```
- [x] In case the public key is not readable.
```sh
$ clientkey2=$(base64 -w0 iss.json)
$ curl -s --cacert certs/ca.pem -H "Authorization: Bearer $token" -H "Client-Public-Key: $clientkey2" "https://localhost:8443/s3/$dataset/$file"
file re-encryption error
```

```sh
$ curl -s --cacert certs/ca.pem -H "Authorization: Bearer $token" -H "Client-Public-Key: $nonexistent" "https://localhost:8443/s3/$dataset/$file" 
c4gh public key is missing from the header
```
- [x] In case the user doesn't have access to the file.
```sh
$ curl -s --cacert certs/ca.pem -H "Authorization: Bearer $WrongToken" -H "Client-Public-Key: $clientkey" "https://localhost:8443/s3/$dataset/$file"
get visas failed
```
Trying with any text after `/s3` lists the datasets that the user has access to:
```sh
$ curl -s --cacert certs/ca.pem -H "Authorization: Bearer $token" -H "Client-Public-Key: $clientkey" "https://localhost:8443/s3/whatever/whatsoever"
<?xml version="1.0" encoding="UTF-8"?>
<ListAllMyBucketsResult><Buckets><Bucket><CreationDate>2024-12-10T14:56:16.277694Z</CreationDate><Name>https://doi.example/ty009.sfrrss/600.45asasga</Name></Bucket></Buckets><Owner></Owner></ListAllMyBucketsResult>
```

- [x] In case a file does not exist. This is the last test (it should not be possible to enumerate files)
```sh
$ curl -s --cacert certs/ca.pem -H "Authorization: Bearer $token" -H "Client-Public-Key: $clientkey" "https://localhost:9443/s3/$dataset/file" -v
```
returns a `< HTTP/1.1 404 Not Found`

### Demo
Suggestion: Two deployments of the download service, one that works with encrypted, one with decrypted and show what's possible or not on the command line.

### Acceptance criteria
- [x] Tests verifying the changes are added

### Additional context
See discussion in previous PR: https://github.com/neicnordic/sensitive-data-archive/pull/695#discussion_r1519911855

aaperis:
This PR closes #750.

**Description**
With this PR `download` can either be deployed as serving either unencrypted or encrypted files. The `serveUnencryptedData` boolean is removed as well as the logic for `download` to always generate an internal c4gh key-pair (that would allow the service to serve unencrypted files in case `serveUnencryptedData=true`).

Main changes:
- `download` can now be configured to only serve unencrypted files if a filepath to a c4gh private key file is provided. Internally, the code checks the validity of the provided key by using the supplied passphrase to retrieve the corresponding c4gh public key. The service will not start if this procedure fails and the service will serve unencrypted data only if the retrieved public key is non-empty. This ensures that there can be no misconfiguration by e.g. providing malfunctioning keys etc. For example, the crypt4gh library and therefore `reencrypt` will use any string provided to it and encrypt a file but with this mechanism we avoid such a scary scenario.
- if no filepath and passphrase are provided, then `download` will serve only encrypted files. This is the default behavior.

Other changes include:
-  go testsuite updates and some refactoring along the way
- update of dev environment: in order to be able to test both setups at once, the docker compose files have been modified to include two instances of  `download` listening at different ports, one serving encrypted and the other serving unencrypted files
- update of integration tests, including a separate bash script for error messages returned to the user (according to the issue description)
- update of helm chart

**How to test**
Integration tests pass.


sda-download/internal/config/config.go
```diff
@@ -481,25 +485,27 @@ func constructWhitelist(obj []TrustedISS) *jwk.MapWhitelist {
...
-// GeneerateC4GHKey generates a keypair and returns the private key as a byte
-// array and the public key as a base64 encoded string
-func GenerateC4GHKey() ([32]byte, string, error) {
-	log.Info("creating temporary crypt4gh key")
-
-	public, private, err := keys.GenerateKeyPair()
+// GetC4GHKey reads and decrypts and returns the c4gh key
+func GetC4GHKeys() ([32]byte, string, error) {
+	keyPath := viper.GetString("app.c4ghPrivateKeyPath")
+	passphrase := viper.GetString("app.c4ghPassphrase")
```
jbygdell, line 491:
```suggestion
	keyPath := viper.GetString("app.c4gh.privateKeyPath")
	passphrase := viper.GetString("app.c4gh.passphrase")
```

sda-download/internal/config/config_test.go
```diff
@@ -75,17 +86,18 @@ func (suite *TestSuite) TestAppConfig() {
...
 
 	viper.Set("app.host", "test")
 	viper.Set("app.port", 1234)
-	viper.Set("app.serveUnencryptedData", false)
 	viper.Set("app.servercert", "test")
 	viper.Set("app.serverkey", "test")
 	viper.Set("log.logLevel", "debug")
 	viper.Set("db.sslmode", "disable")
+	viper.Set("app.c4ghPrivateKeyPath", privateKeyFile.Name())
+	viper.Set("app.c4ghPassphrase", "password")
```
jbygdell, line 100:
```suggestion
	viper.Set("app.c4gh.privateKeyPath", privateKeyFile.Name())
	viper.Set("app.c4gh.passphrase", "password")
```

sda-download/internal/config/config_test.go
```diff
@@ -94,15 +106,25 @@ func (suite *TestSuite) TestAppConfig() {
...
 
-	// Check the key that was generated
+	// Check the private key that was loaded by checking the derived public key
 	publicKey, err := base64.StdEncoding.DecodeString(c.App.Crypt4GHPublicKeyB64)
 	assert.Nilf(suite.T(), err, "Incorrect public c4gh key generated (error in base64 encoding)")
 	_, err = keys.ReadPublicKey(bytes.NewReader(publicKey))
 	assert.Nilf(suite.T(), err, "Incorrect public c4gh key generated (bad key)")
+
+	// Check false c4gh key
+	viper.Set("app.c4ghPrivateKeyPath", "some/nonexistent.key")
```
jbygdell, line 119:
```suggestion
	viper.Set("app.c4gh.privateKeyPath", "some/nonexistent.key")
```

sda-download/internal/config/config_test.go
```diff
@@ -94,15 +106,25 @@ func (suite *TestSuite) TestAppConfig() {
...
+	err = c.appConfig()
+	assert.ErrorContains(suite.T(), err, "no such file or directory")
+
+	// Check false c4gh key
+	viper.Set("app.c4ghPrivateKeyPath", privateKeyFile.Name())
```
jbygdell, line 124:
```suggestion
	viper.Set("app.c4gh.privateKeyPath", privateKeyFile.Name())
```

sda-download/internal/config/config_test.go
```diff
@@ -94,15 +106,25 @@ func (suite *TestSuite) TestAppConfig() {
...
+	viper.Set("app.c4ghPassphrase", "blablabla")
```
jbygdell, line 125:
```suggestion
	viper.Set("app.c4gh.passphrase", "blablabla")
```

sda-download/dev_utils/compose-no-tls.yml
```diff
@@ -74,6 +74,34 @@ services:
...
+      s3:
+        condition: service_healthy
+      mockauth:
+        condition: service_started
+    environment:
+      - ARCHIVE_URL=http://s3
+      - ARCHIVE_TYPE=s3
+      - DB_HOST=db
+      - APP_C4GHPRIVATEKEYPATH=/dev_utils/c4gh.sec.pem
+      - APP_C4GHPASSPHRASE=oaagCP1YgAZeEyl2eJAkHv9lkcWXWFgm
```
jbygdell, line 92:
```suggestion
      - APP_C4GH_PRIVATEKEYPATH=/dev_utils/c4gh.sec.pem
      - APP_C4GH_PASSPHRASE=oaagCP1YgAZeEyl2eJAkHv9lkcWXWFgm
```

sda-download/dev_utils/config.yaml
```diff
@@ -4,7 +4,6 @@ app:
   serverkey: "./dev_utils/certs/download-key.pem"
   port: "8443"
   middleware: "default"
```
jbygdell, line 6:
```suggestion
  middleware: "default"
  c4gh:
    passphrase: ""
    privateKeyPath:  ""
```

jbygdell (changes requested):

sda-download/dev_utils/config.yaml
aaperis: not sure whys this is not marked as outdated but it's changed now

sda-download/dev_utils/config.yaml
```diff
@@ -4,7 +4,9 @@ app:
...
-  serveUnencryptedData: true
+  c4gh:
+    passphrase: ""
+    privateKeyPath:  ""
```
jbygdell, line 9: What is the point of this addition?
The crypt4gh key stuff are declared further down in the config file and we only need one.

sda-download/dev_utils/run_integration_test.sh
```diff
@@ -194,7 +194,7 @@ echo "expected file found"
 # Test file can be decrypted
 ## test also the files endpoint
 
-C4GH_PASSPHRASE=$(grep -F passphrase config.yaml | sed -e 's/.* //' -e 's/"//g')
+C4GH_PASSPHRASE=$(grep -F passphrase config.yaml | tail -1 | sed -e 's/.* //' -e 's/"//g')
```
jbygdell, line 197:
```suggestion
C4GH_PASSPHRASE=$(grep -F passphrase config.yaml | sed -e 's/.* //' -e 's/"//g')
```
Shouldn't be needed since the config file only will have one cryptgh config block

sda-download/internal/config/config.go
```diff
@@ -380,9 +376,17 @@ func (c *Map) appConfig() error {
...
 
 	var err error
-	c.App.Crypt4GHPrivateKey, c.App.Crypt4GHPublicKeyB64, err = GenerateC4GHKey()
-	if err != nil {
-		return err
+	if viper.GetString("app.c4gh.privateKeyPath") != "" {
+
+		if !viper.IsSet("app.c4gh.passphrase") {
+			return errors.New("app.c4gh.passphrase is not set")
+		}
```
jbygdell, line 383:
```suggestion
	if viper.GetString("app.c4gh.privateKeyPath") != "" {

		if !viper.IsSet("c4gh.passphrase") {
			return errors.New("c4gh.passphrase is not set")
		}
```

c4gh config is already declared and we only need one entry in the config

sda-download/internal/config/config.go
```diff
@@ -481,25 +485,27 @@ func constructWhitelist(obj []TrustedISS) *jwk.MapWhitelist {
...
-// GeneerateC4GHKey generates a keypair and returns the private key as a byte
-// array and the public key as a base64 encoded string
-func GenerateC4GHKey() ([32]byte, string, error) {
-	log.Info("creating temporary crypt4gh key")
-
-	public, private, err := keys.GenerateKeyPair()
+// GetC4GHKey reads and decrypts and returns the c4gh key
+func GetC4GHKeys() ([32]byte, string, error) {
+	keyPath := viper.GetString("app.c4gh.privateKeyPath")
+	passphrase := viper.GetString("app.c4gh.passphrase")
```
jbygdell, line 491:
```suggestion
	keyPath := viper.GetString("c4gh.privateKeyPath")
	passphrase := viper.GetString("c4gh.passphrase")
```

sda-download/internal/config/config_test.go
```diff
@@ -75,17 +86,18 @@ func (suite *TestSuite) TestAppConfig() {
...
 
 	viper.Set("app.host", "test")
 	viper.Set("app.port", 1234)
-	viper.Set("app.serveUnencryptedData", false)
 	viper.Set("app.servercert", "test")
 	viper.Set("app.serverkey", "test")
 	viper.Set("log.logLevel", "debug")
 	viper.Set("db.sslmode", "disable")
+	viper.Set("app.c4gh.PrivateKeyPath", privateKeyFile.Name())
+	viper.Set("app.c4gh.passphrase", "password")
```
jbygdell, line 100:
```suggestion
	viper.Set("c4gh.PrivateKeyPath", privateKeyFile.Name())
	viper.Set("c4gh.passphrase", "password")
```

sda-download/internal/config/config_test.go
```diff
@@ -94,15 +106,25 @@ func (suite *TestSuite) TestAppConfig() {
...
 
-	// Check the key that was generated
+	// Check the private key that was loaded by checking the derived public key
 	publicKey, err := base64.StdEncoding.DecodeString(c.App.Crypt4GHPublicKeyB64)
 	assert.Nilf(suite.T(), err, "Incorrect public c4gh key generated (error in base64 encoding)")
 	_, err = keys.ReadPublicKey(bytes.NewReader(publicKey))
 	assert.Nilf(suite.T(), err, "Incorrect public c4gh key generated (bad key)")
+
+	// Check false c4gh key
+	viper.Set("app.c4gh.privateKeyPath", "some/nonexistent.key")
```
jbygdell, line 119:
```suggestion
	viper.Set("c4gh.privateKeyPath", "some/nonexistent.key")
```

sda-download/internal/config/config_test.go
```diff
@@ -94,15 +106,25 @@ func (suite *TestSuite) TestAppConfig() {
...
+	err = c.appConfig()
+	assert.ErrorContains(suite.T(), err, "no such file or directory")
+
+	// Check false c4gh key
+	viper.Set("app.c4gh.privateKeyPath", privateKeyFile.Name())
+	viper.Set("app.c4gh.passphrase", "blablabla")
```
jbygdell, line 125:
```suggestion
	viper.Set("c4gh.privateKeyPath", privateKeyFile.Name())
	viper.Set("c4gh.passphrase", "blablabla")
```

sda-download/.github/integration/tests/s3notls/52_check_endpoint.sh
```diff
@@ -70,12 +70,21 @@ echo "expected file found"
 # Test file can be decrypted
 ## test also the files endpoint
 
-C4GH_PASSPHRASE=$(grep -F passphrase config.yaml | sed -e 's/.* //' -e 's/"//g')
+C4GH_PASSPHRASE=$(grep -F passphrase config.yaml | tail -1 | sed -e 's/.* //' -e 's/"//g')
```
jbygdell, line 73:
```suggestion
C4GH_PASSPHRASE=$(grep -F passphrase config.yaml | sed -e 's/.* //' -e 's/"//g')
```

jbygdell (changes requested):
Crypt4gh config already exists, there is not need to add another one since the app will only need one.

Either reuse the existing one or move everything up under the `app` sub heading.

sda-download/dev_utils/config.yaml
aaperis: I would expect that the internal key that `download` uses is not the same as the one that `reencrypt` uses (the latter being the archive's main key).

aaperis: Also, I thought this addition was your suggestion? https://github.com/neicnordic/sensitive-data-archive/pull/1163#discussion_r1883523274

jbygdell: The config file for `sda-download` is not unified with the one for the rest of the sda-stack and if a single file is used to start all services you will get into trouble.

The config file in this folder should be for sda-download and noting else.

aaperis:
I unresolved the initial the conversations that I set as resolved (these are answered) so that @jbygdell can take a look and resolve. According to how we decided to do things on our last meeting about PRs.

sda-download/dev_utils/config.yaml
jbygdell: > I would expect that the internal key that `download` uses is not the same as the one that `reencrypt` uses (the latter being the archive's main key).

It should not be the same, but the config file for sda-download should only have one c4gh entry until it is merged into the `sda` code structure.

sda-download/dev_utils/run_integration_test.sh
jbygdell: If you want to extract a specific entry using `yq` is the simplest solution.
```sh
yq '.app.c4gh.passphrase' config.yaml
```

sda-download/dev_utils/config.yaml
aaperis: I am not following,  the config file here is used for sda-download's integration tests and there you need both sda-download and reencrypt c4gh keys. Please elaborate.

jbygdell: Yes, and that is the root of the problem, there is one config file that is used to configure two services that do not have a unified code to parse the configuration.

aaperis: yes, but that was a problem all along and it will be as long as we don't merge `sda-download` to the stack. I don't see how this is relevant here specifically

aaperis: The code around `sda-download` (integration tests and what lies in the dev_utils directory) is a mess. My point is that fixing this is not part of this PR.

sda-download/internal/config/config.go
aaperis: no it is not, there is no other entry for c4gh keys in `sda-download/internal/config.go`

sda-download/internal/config/config.go
aaperis: same as above

sda-download/internal/config/config_test.go
aaperis: same as above

sda-download/internal/config/config_test.go
aaperis: same as above

sda-download/internal/config/config_test.go
aaperis: same as above

sda-download/dev_utils/config.yaml
jbygdell: Fixing all the tests is out of the scope, but you can take one step in the right direction by moving out everything related to the re-encrypt service from the conifig.yaml. Either in to a separate config file or into the compose file

sda-download/internal/config/config.go
jbygdell: All the issues stems from the fact that the config.yaml is not dedicated to sda-download.
Still, it should not be under the `app` section since that holds the HTTP relates configs for port, tls and so.

charts/sda-svc/values.yaml
```diff
@@ -239,7 +239,9 @@ global:
       iss:
         - iss: "https://login.elixir-czech.org/oidc"
           jku: "https://login.elixir-czech.org/oidc/jwk"
-    serveUnencryptedData: false
+    serveDecrypted:
+      c4ghKeyFile: ""
+      secretName: ""
```
jbygdell, line 244:
```suggestion
    serveDecrypted:
      c4ghKeyFile: ""
      passphrase: ""
      secretName: ""
```

Private key needs a passphrase.

aaperis: The passphrase is intended to be read from the secret, similarly as to how we have it in `ingest-deploy.yaml`, see [here](https://github.com/neicnordic/sensitive-data-archive/blob/1176989fe22f88b526cc9603d861cf112ffdec7e/charts/sda-svc/templates/ingest-deploy.yaml#L212). 
Imo this is better than passing the secret via a `values.yaml` file.

I can either add a note in the README saying that the secretName should include the passphrase
or add the `passphrase` variable as suggested above.

.github/integration/scripts/charts/dependencies.sh
```diff
@@ -31,7 +31,7 @@ else
 fi
 
 # secret for the crypt4gh keypair
-C4GHPASSPHRASE="$(random-string)"
+C4GHPASSPHRASE="ZB8ko3g2P9wEgagpgqCS7vJhJoKrSUfJ"
```
jbygdell, line 34:
```suggestion
C4GHPASSPHRASE="$(random-string)"
```

There is no reason for this change

charts/sda-svc/values.yaml
jbygdell: Right, in the notes it goes then.

.github/integration/scripts/charts/dependencies.sh
nanjiangshu: I also prefer no changes.

sda-download/dev_utils/config.yaml
MalinAhlberg: I find it quite hard to follow this, so I agree that better documentation would be nice :sweat_smile: . To me, it sounds like it's time for a new issue, and only going for a minimal working solution (regarding the config file) in this PR.

sda-download/.github/integration/tests/common/80_check_reencrypt.sh
```diff
@@ -18,16 +18,10 @@ dataset="https://doi.example/ty009.sfrrss/600.45asasga"
 file="dummy_data"
 expected_size=1048605
 
-# Download unencrypted full file,  check file size
-curl --cacert certs/ca.pem -H "Authorization: Bearer $token" "https://localhost:8443/s3/$dataset/$file" --output full1.bam
-
-if [ ! -f "full1.bam" ]; then
```
nanjiangshu, line 24: I think it is better to keep this checking since otherwise it will reported as "Incorrect file size" later.

sda-download/.github/integration/tests/common/80_check_reencrypt.sh
```diff
@@ -51,10 +41,13 @@ fi
...
 # Decrypt the reencrypted file and compare it with the original unencrypted file
 export C4GH_PASSPHRASE="strongpass" # passphrase for the client crypt4gh key
-if ! crypt4gh decrypt --sk client.sec.pem < $reencryptedFile > full2.bam; then
+crypt4gh decrypt -s client.sec.pem -f $reencryptedFile
+if [ ! -f "${reencryptedFile%.c4gh}" ] ; then
     echo "Failed to decrypt re-encrypted file with the client's private key"
     exit 1
 fi
+mv "${reencryptedFile%.c4gh}" full2.bam
+
```
nanjiangshu, line 50:
```suggestion
```

sda-download/.github/integration/tests/common/90_check_s3_errors.sh
```diff
@@ -0,0 +1,92 @@
...
+
+if [ -z "$token" ]; then
+    echo "Failed to obtain token"
+    exit 1
+fi
+
+dataset="https://doi.example/ty009.sfrrss/600.45asasga"
+file="dummy_data"
+clientkey=$(base64 -w0 client.pub.pem)
+bad_token=token2=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJyZXF1ZXN0ZXJAZGVtby5vcmciLCJhdWQiOlsiYXVkMiIsImF1ZDMiXSwiYXpwIjoiYXpwIiwic2NvcGUiOiJvcGVuaWQiLCJpc3MiOiJodHRwczovL2RlbW8uZXhhbXBsZSIsImV4cCI6OTk5OTk5OTk5OSwiaWF0IjoxNTYxNjIxOTEzLCJqdGkiOiI2YWQ3YWE0Mi0zZTljLTQ4MzMtYmQxNi03NjVjYjgwYzIxMDIifQ.ncUyjNytxqS9bqLnsbjv6D839PnHVw-anQS4bFpAs20
```
nanjiangshu, line 20: This is a bit confusing for providing a `bad_token` at the first glance.

nanjiangshu, line 20:
```suggestion
bad_token=BADeyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJyZXF1ZXN0ZXJAZGVtby5vcmciLCJhdWQiOlsiYXVkMiIsImF1ZDMiXSwiYXpwIjoiYXpwIiwic2NvcGUiOiJvcGVuaWQiLCJpc3MiOiJodHRwczovL2RlbW8uZXhhbXBsZSIsImV4cCI6OTk5OTk5OTk5OSwiaWF0IjoxNTYxNjIxOTEzLCJqdGkiOiI2YWQ3YWE0Mi0zZTljLTQ4MzMtYmQxNi03NjVjYjgwYzIxMDIifQ.ncUyjNytxqS9bqLnsbjv6D839PnHVw-anQS4bFpAs20
```

sda-download/.github/integration/tests/s3notls/52_check_endpoint.sh
nanjiangshu: I think it's a valid change in case there are multiple lines with the word `passphrase`. Furthermore, we may need to refactor it when SDA supports multiple keys.

sda-download/api/api.md
```diff
@@ -1,12 +1,9 @@
 # API
 
-
 The Download API service provides functionality for downloading files from the Archive.
-It implements the [Data Out API](https://neic-sda.readthedocs.io/en/latest/dataout/#rest-api-endpoints).
-
-Further, it enables the endpoints `/s3` and `/s3-encrypted`, used for htsget.
+It implements the [Data Out API](https://neic-sda.readthedocs.io/en/latest/dataout/#rest-api-endpoints). Further, it enables the endpoint `/s3` used for htsget and other services that need to interface with an s3-backend storage.
```
nanjiangshu, line 4:
```suggestion
It implements the [Data Out API](https://neic-sda.readthedocs.io/en/latest/dataout/#rest-api-endpoints). Further, it enables the endpoint `/s3`, which is used for htsget and other services that need to interface with an S3-backend storage.
```

sda-download/api/api.md
```diff
@@ -25,10 +22,9 @@ The client can establish a session to bypass time-costly visa validations for fu
 - `/metadata/datasets/*dataset`
 - `/files/:fileid`
 
-**[File download requests, for htsget](#file-download-requests)**
+**[File download requests, for s3 endpoint](#file-download-requests)**
```
nanjiangshu, line 25:
```suggestion
**[File download requests, for `/s3` endpoint](#file-download-requests)**
```

sda-download/api/api.md
```diff
@@ -100,10 +96,11 @@ Parts of a file can be requested with specific byte ranges using `startCoordinat
 ```
 
 ### File download requests
-These endpoints are designed for usage with [htsget](https://samtools.github.io/hts-specs/htsget.html).
+This endpoint is designed for usage with [htsget](https://samtools.github.io/hts-specs/htsget.html) or other external applications that interface with s3-storage backends.
```
nanjiangshu, line 99:
```suggestion
This endpoint is designed for usage with [htsget](https://samtools.github.io/hts-specs/htsget.html) or other external applications that interface with S3-storage backends.
```

sda-download/api/sda/sda.go
```diff
@@ -207,12 +207,22 @@ func Files(c *gin.Context) {
...
+	// when Crypt4GHPublicKeyB64 is not set or empty, but this is not the case for calls to /files endpoint.
+	// So we need this check.
+
+	if c.Param("type") != "encrypted" && config.Config.App.Crypt4GHPublicKeyB64 == "" {
 		c.String(http.StatusBadRequest, "downloading unencrypted data is not supported")
 
 		return
 	}
 
+	if c.Param("type") != "encrypted" && c.GetHeader("Client-Public-Key") != "" {
```
MalinAhlberg, line 220: I'm not following...
if `type=decrypted` (or at least `!=encrypted`), and there is no client public key provided,  then the user has requested un-encrypted data. Why should we give a bad request with  `downloading encrypted data is not supported`?

.github/integration/scripts/charts/dependencies.sh
aaperis: @nanjiangshu  If we don't use a fixed value here then  due to this block [here](https://github.com/neicnordic/sensitive-data-archive/blob/f1979be39c23318500ba58cb340608689431557c/.github/integration/scripts/charts/dependencies.sh#L40), you need to manually remove the keys every time you want to run local tests for Helm charts using the commands from the Makefile. 

But I'll set it back to random and make an issue.

MalinAhlberg (commented):
Looks like good and thorough work..!
I left some questions, but otherwise I think it looks good!

sda-download/internal/config/config.go
nanjiangshu: I think the reason why @aaperis put it under `app` instead of directly under `c4gh` is try to distinguish this private key from the private key used for other services such as `ingest`. There might be some confusing when we try to merge `sda-download` to SDA later on. I'm OK to put at either places as long as they works. But I prefer that they are distinguishable.

nanjiangshu (commented):
Looks good! Well done! I left a few comments and also under discussions. I can approve it when the current conversations are resolved.

.github/integration/scripts/charts/dependencies.sh
aaperis: Change reverted, new issue opened regarding this: #1208.

nanjiangshu: Aha, I don't know about this. That's a good point.

sda-download/internal/config/config.go
jbygdell: As stated elsewhere, do not mix in how the other sda services does things.
There is zero code overlap at the moment and if one would make sure that the issues are kept to a minimum when this gets merged the make sure to follow the given semantics on how things are structured.

.github/integration/scripts/charts/dependencies.sh
aaperis: That's on me, I should have explained it better :smile: .

sda-download/api/api.md
```diff
@@ -100,10 +96,11 @@ Parts of a file can be requested with specific byte ranges using `startCoordinat
...
+
+The `/s3` endpoint accepts the parameters described below. Note that depending on the configuration of the download service, `/s3` may either serve only encrypted or decrypted files.
 
-The `/s3` and `/s3-encrypted` endpoints accept the same parameters, described below.
-Note that the download service may be configured to only allow encrypted file downloads.
+By default,
```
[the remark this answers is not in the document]
aaperis: Oops, good catch!

sda-download/api/sda/sda.go
aaperis: Yeap, this can be confusing:
If `type!=encrypted` then this means that `download` is deployed as serving unencrypted files (see [here](https://github.com/neicnordic/sensitive-data-archive/blob/f1979be39c23318500ba58cb340608689431557c/sda-download/api/s3/s3.go#L250) for example). 

But if the user has provided a Client pub key in the header that means that the user is trying to download encrypted files (i.e. encrypted with their provided pubkey).
Thus, this will throw a bad request error (as per the issue description).

MalinAhlberg: ah, I thought the `type` came from the user's request. But in that case, it makes much sense :+1: .

sda-download/dev_utils/config.yaml
aaperis: As a resolve to this, I opened Issue #1209,  feel free to refine the description of the issue :smile: .

sda-download/internal/config/config.go
aaperis: I would prefer to keep the variable names separated. I don't find it practical to use variable naming that other sda applications use only to change them in the future because they overlap.

jbygdell: Keeping the names unique is a no-brainer since it will make things easier down the road, but it should stil be under the `c4gh` heading to conform with the overall semantics.

So something like this maybe:
```
c4gh.transientKeyPath
c4gh.transientPassphrase
```

sda-download/.github/integration/tests/common/80_check_reencrypt.sh
nanjiangshu: OK. It makes sense to me. We can removed it for easier debugging.

sda-download/.github/integration/tests/s3notls/52_check_endpoint.sh
aaperis: changed all yaml parsing to using `yq` as suggested above

sda-download/dev_utils/run_integration_test.sh
aaperis: changed all similar  to using `yq`

MalinAhlberg (commented):
I've tested this manually, and the `s3` endpoint seems to work as expected :star: . However, the `files` endpoint is not supported when the service is in encrypted mode, the answer is always `downloading unencrypted data is not supported`. Maybe that's intended..? [The docs](https://neic-sda.readthedocs.io/en/latest/dataout/#rest-api-endpoints) says you can provide the header `Public-Key`, but I cannot get that to work.
If this behavior is intended (or if someone can point out what I'm doing wrong), I will approve :)

aaperis:
> I've tested this manually, and the `s3` endpoint seems to work as expected ⭐ . However, the `files` endpoint is not supported when the service is in encrypted mode, the answer is always `downloading unencrypted data is not supported`. Maybe that's intended..? [The docs](https://neic-sda.readthedocs.io/en/latest/dataout/#rest-api-endpoints) says you can provide the header `Public-Key`, but I cannot get that to work. I also tried the `Client-Public-Key` header. If this behavior is intended (or if someone can point out what I'm doing wrong), I will approve :)

Good point @MalinAhlberg, just tried the same thing that you describe with the latest version from `main` and the behavior is the same. So the current behavior is at least consistent. 

Not sure if we ever had `/files` serving encrypted data, I'll check deeper. In any case either the docs are misleading or/and there's something fishier. Great catch!

aaperis:
> I've tested this manually, and the `s3` endpoint seems to work as expected ⭐ . However, the `files` endpoint is not supported when the service is in encrypted mode, the answer is always `downloading unencrypted data is not supported`. Maybe that's intended..? [The docs](https://neic-sda.readthedocs.io/en/latest/dataout/#rest-api-endpoints) says you can provide the header `Public-Key`, but I cannot get that to work. I also tried the `Client-Public-Key` header. If this behavior is intended (or if someone can point out what I'm doing wrong), I will approve :)

Good point @MalinAhlberg, just tried the same thing that you describe with the latest version from main and the behavior is the same. So the current behavior is at least consistent.

Not sure  `/files` ever served encrypted data, from what I see in #695 and #761 all work for re-encryption is done on the `s3` side only. For example see [here](https://github.com/neicnordic/sensitive-data-archive/blob/2747e979973a49de64d6dc7166aff8d356befead/sda-download/api/sda/sda.go#L363). The `c.Param("type") is only set to  `encrypted` if a `s3` endpoint is called. Otherwise, I believe the default case in this switch statement seems to stream decrypted data, see [here](https://github.com/neicnordic/sensitive-data-archive/blob/2747e979973a49de64d6dc7166aff8d356befead/sda-download/api/sda/sda.go#L431).

All in all, we need to discuss about this with the team.
Good catch!

MalinAhlberg (dismissed):

sda-download/dev_utils/compose-no-tls.yml
```diff
@@ -74,6 +74,34 @@ services:
...
+      s3:
+        condition: service_healthy
+      mockauth:
+        condition: service_started
+    environment:
+      - ARCHIVE_URL=http://s3
+      - ARCHIVE_TYPE=s3
+      - DB_HOST=db
+      - APP_C4GH_PRIVATEKEYPATH=/dev_utils/c4gh.sec.pem
+      - APP_C4GH_PASSPHRASE=oaagCP1YgAZeEyl2eJAkHv9lkcWXWFgm
```
jbygdell, line 92:
```suggestion
      - C4GH_TRANSIENTKEYPATH=/dev_utils/c4gh.sec.pem
      - C4GH_TRANSIENTPASSPHRASE=oaagCP1YgAZeEyl2eJAkHv9lkcWXWFgm
```

sda-download/api/sda/sda_test.go
```diff
@@ -689,11 +743,15 @@ func (f *fakeGRPC) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	privateKeyFilePath, err := GenerateTestC4ghKey(t)
+	assert.NoError(t, err)
+
 	// Save original to-be-mocked functions
 	originalCheckFilePermission := database.CheckFilePermission
 	originalGetCacheFromContext := middleware.GetCacheFromContext
 	originalGetFile := database.GetFile
-	originalServeUnencryptedData := config.Config.App.ServeUnencryptedData
+	originalServeUnencryptedDataTrigger := config.Config.App.Crypt4GHPublicKeyB64
+	originalC4ghPrivateKeyFilepath := config.Config.App.Crypt4GHPrivateKey
```
jbygdell, line 754:
```suggestion
	originalServeUnencryptedDataTrigger := config.Config.C4GH.PublicKeyB64
	originalC4ghPrivateKeyFilepath := config.Config.C4GH.PrivateKey
```

sda-download/api/sda/sda_test.go
```diff
@@ -589,7 +640,10 @@ func TestDownload_Fail_OpenFile(t *testing.T) {
 	database.CheckFilePermission = originalCheckFilePermission
 	middleware.GetCacheFromContext = originalGetCacheFromContext
 	database.GetFile = originalGetFile
-	config.Config.App.ServeUnencryptedData = originalServeUnencryptedData
+	config.Config.App.Crypt4GHPublicKeyB64 = originalServeUnencryptedDataTrigger
+	config.Config.App.Crypt4GHPrivateKey = originalC4ghPrivateKeyFilepath
+	viper.Set("app.c4gh.privateKeyPath", "")
+	viper.Set("app.c4gh.passphrase", "")
```
jbygdell, line 646:
```suggestion
	config.Config.C4GH.PublicKeyB64 = originalServeUnencryptedDataTrigger
	config.Config.C4GH.PrivateKey = originalC4ghPrivateKeyFilepath
	viper.Set("c4gh.transientKeyPath", "")
	viper.Set("c4gh.ttransientPassphrase", "")
```

sda-download/api/sda/sda_test.go
```diff
@@ -556,7 +603,11 @@ func TestDownload_Fail_OpenFile(t *testing.T) {
 
 		return fileDetails, nil
 	}
-	config.Config.App.ServeUnencryptedData = true
+
+	viper.Set("app.c4gh.privateKeyPath", privateKeyFilePath)
+	viper.Set("app.c4gh.passphrase", "password")
+	config.Config.App.Crypt4GHPrivateKey, config.Config.App.Crypt4GHPublicKeyB64, err = config.GetC4GHKeys()
```
jbygdell, line 609:
```suggestion
	viper.Set("c4gh.transientKeyPath", privateKeyFilePath)
	viper.Set("c4gh.transientPassphrase", "password")
	config.Config.C4GH.PrivateKey, config.Config.C4GH.PublicKeyB64, err = config.GetC4GHKeys()
```

sda-download/api/sda/sda_test.go
```diff
@@ -525,17 +565,24 @@ func TestDownload_Fail_GetFile(t *testing.T) {
...
+	privateKeyFilePath, err := GenerateTestC4ghKey(t)
+	assert.NoError(t, err)
+
 	// Save original to-be-mocked functions
 	originalCheckFilePermission := database.CheckFilePermission
 	originalGetCacheFromContext := middleware.GetCacheFromContext
 	originalGetFile := database.GetFile
-	originalServeUnencryptedData := config.Config.App.ServeUnencryptedData
+	originalServeUnencryptedDataTrigger := config.Config.App.Crypt4GHPublicKeyB64
+	originalC4ghPrivateKeyFilepath := config.Config.App.Crypt4GHPrivateKey
```
jbygdell, line 585:
```suggestion
	originalServeUnencryptedDataTrigger := config.Config.C4GH.PublicKeyB64
	originalC4ghPrivateKeyFilepath := config.Config.C4GH.PrivateKey
```

sda-download/api/sda/sda_test.go
```diff
@@ -525,17 +565,24 @@ func TestDownload_Fail_GetFile(t *testing.T) {
 	database.CheckFilePermission = originalCheckFilePermission
 	middleware.GetCacheFromContext = originalGetCacheFromContext
 	database.GetFile = originalGetFile
-	config.Config.App.ServeUnencryptedData = originalServeUnencryptedData
+	config.Config.App.Crypt4GHPublicKeyB64 = originalServeUnencryptedDataTrigger
+	config.Config.App.Crypt4GHPrivateKey = originalC4ghPrivateKeyFilepath
+	viper.Set("app.c4gh.privateKeyPath", "")
+	viper.Set("app.c4gh.passphrase", "")
```
jbygdell, line 571:
```suggestion
	config.Config.C4GH.PublicKeyB64 = originalServeUnencryptedDataTrigger
	config.Config.C4GH.PrivateKey = originalC4ghPrivateKeyFilepath
	viper.Set("c4gh.transientKeyPath", "")
	viper.Set("c4gh.transientPassphrase", "")
```

sda-download/api/sda/sda_test.go
```diff
@@ -497,11 +531,17 @@ func TestDownload_Fail_GetFile(t *testing.T) {
 	database.GetFile = func(_ string) (*database.FileDownload, error) {
 		return nil, errors.New("database error")
 	}
-	config.Config.App.ServeUnencryptedData = true
+
+	viper.Set("app.c4gh.privateKeyPath", privateKeyFilePath)
+	viper.Set("app.c4gh.passphrase", "password")
+	config.Config.App.Crypt4GHPrivateKey, config.Config.App.Crypt4GHPublicKeyB64, err = config.GetC4GHKeys()
```
jbygdell, line 537:
```suggestion
	viper.Set("c4gh.transientKeyPath", privateKeyFilePath)
	viper.Set("c4gh.transientPassphrase", "password")
	config.Config.C4GH.PrivateKey, config.Config.C4GH.PublicKeyB64, err = config.GetC4GHKeys()
```

sda-download/api/sda/sda_test.go
```diff
@@ -473,17 +500,24 @@ func TestDownload_Fail_NoPermissions(t *testing.T) {
...
+	privateKeyFilePath, err := GenerateTestC4ghKey(t)
+	assert.NoError(t, err)
+
 	// Save original to-be-mocked functions
 	originalCheckFilePermission := database.CheckFilePermission
 	originalGetCacheFromContext := middleware.GetCacheFromContext
 	originalGetFile := database.GetFile
-	originalServeUnencryptedData := config.Config.App.ServeUnencryptedData
+	originalServeUnencryptedDataTrigger := config.Config.App.Crypt4GHPublicKeyB64
+	originalC4ghPrivateKeyFilepath := config.Config.App.Crypt4GHPrivateKey
```
jbygdell, line 520:
```suggestion
	originalServeUnencryptedDataTrigger := config.Config.C4GH.PublicKeyB64
	originalC4ghPrivateKeyFilepath := config.Config.C4GH.PrivateKey
```

sda-download/api/sda/sda_test.go
```diff
@@ -473,17 +500,24 @@ func TestDownload_Fail_NoPermissions(t *testing.T) {
 	// Return mock functions to originals
 	database.CheckFilePermission = originalCheckFilePermission
 	middleware.GetCacheFromContext = originalGetCacheFromContext
-	config.Config.App.ServeUnencryptedData = originalServeUnencryptedData
+	config.Config.App.Crypt4GHPublicKeyB64 = originalServeUnencryptedDataTrigger
+	config.Config.App.Crypt4GHPrivateKey = originalC4ghPrivateKeyFilepath
+	viper.Set("app.c4gh.privateKeyPath", "")
+	viper.Set("app.c4gh.passphrase", "")
```
jbygdell, line 506:
```suggestion
	config.Config.C4GH.PublicKeyB64 = originalServeUnencryptedDataTrigger
	config.Config.C4GH.PrivateKey = originalC4ghPrivateKeyFilepath
	viper.Set("c4gh.transientKeyPath", "")
	viper.Set("c4gh.transientPassphrase", "")
```

sda-download/api/sda/sda_test.go
```diff
@@ -446,11 +467,17 @@ func TestDownload_Fail_NoPermissions(t *testing.T) {
 	middleware.GetCacheFromContext = func(_ *gin.Context) session.Cache {
 		return session.Cache{}
 	}
-	config.Config.App.ServeUnencryptedData = true
+
+	viper.Set("app.c4gh.privateKeyPath", privateKeyFilePath)
+	viper.Set("app.c4gh.passphrase", "password")
+	config.Config.App.Crypt4GHPrivateKey, config.Config.App.Crypt4GHPublicKeyB64, err = config.GetC4GHKeys()
```
jbygdell, line 473:
```suggestion
	viper.Set("c4gh.transientKeyPath", privateKeyFilePath)
	viper.Set("c4gh.transientPassphrase", "password")
	config.Config.C4GH.PrivateKey, config.Config.C4GH.PublicKeyB64, err = config.GetC4GHKeys()
```

sda-download/api/sda/sda_test.go
```diff
@@ -427,16 +441,23 @@ func TestDownload_Fail_FileNotFound(t *testing.T) {
...
 
+	privateKeyFilePath, err := GenerateTestC4ghKey(t)
+	assert.NoError(t, err)
+
 	// Save original to-be-mocked functions
 	originalCheckFilePermission := database.CheckFilePermission
 	originalGetCacheFromContext := middleware.GetCacheFromContext
-	originalServeUnencryptedData := config.Config.App.ServeUnencryptedData
+	originalServeUnencryptedDataTrigger := config.Config.App.Crypt4GHPublicKeyB64
+	originalC4ghPrivateKeyFilepath := config.Config.App.Crypt4GHPrivateKey
```
jbygdell, line 460:
```suggestion
	originalServeUnencryptedDataTrigger := config.Config.C4GH.PublicKeyB64
	originalC4ghPrivateKeyFilepath := config.Config.C4GH.PrivateKey
```

sda-download/api/sda/sda_test.go
```diff
@@ -427,16 +441,23 @@ func TestDownload_Fail_FileNotFound(t *testing.T) {
 
 	// Return mock functions to originals
 	database.CheckFilePermission = originalCheckFilePermission
-	config.Config.App.ServeUnencryptedData = originalServeUnencryptedData
+	config.Config.App.Crypt4GHPublicKeyB64 = originalServeUnencryptedDataTrigger
+	config.Config.App.Crypt4GHPrivateKey = originalC4ghPrivateKeyFilepath
+	viper.Set("app.c4gh.privateKeyPath", "")
+	viper.Set("app.c4gh.passphrase", "")
```
jbygdell, line 447:
```suggestion
	config.Config.C4GH.PublicKeyB64 = originalServeUnencryptedDataTrigger
	config.Config.C4GH.PrivateKey = originalC4ghPrivateKeyFilepath
	viper.Set("c4gh.transientKeyPath", "")
	viper.Set("c4gh.transientPassphrase", "")
```

sda-download/api/sda/sda_test.go
```diff
@@ -377,35 +371,55 @@ func TestDownload_Fail_UnencryptedDownloadNotAllowed(t *testing.T) {
...
 
 	// Substitute mock functions
 	database.CheckFilePermission = func(_ string) (string, error) {
 		return "", errors.New("file not found")
 	}
-	config.Config.App.ServeUnencryptedData = true
+
+	viper.Set("app.c4gh.privateKeyPath", privateKeyFilePath)
+	viper.Set("app.c4gh.passphrase", "password")
+	config.Config.App.Crypt4GHPrivateKey, config.Config.App.Crypt4GHPublicKeyB64, err = config.GetC4GHKeys()
```
jbygdell, line 415:
```suggestion
	viper.Set("c4gh.transientKeyPath", privateKeyFilePath)
	viper.Set("c4gh.transientPassphrase", "password")
	config.Config.C4GH.PrivateKey, config.Config.C4GH.PublicKeyB64, err = config.GetC4GHKeys()
```

sda-download/api/sda/sda_test.go
```diff
@@ -377,35 +371,55 @@ func TestDownload_Fail_UnencryptedDownloadNotAllowed(t *testing.T) {
...
 
-	// Save original to-be-mocked functions
+	privateKeyFilePath, err := GenerateTestC4ghKey(t)
+	assert.NoError(t, err)
+
+	// Save original to-be-mocked functions and config
 	originalCheckFilePermission := database.CheckFilePermission
-	originalServeUnencryptedData := config.Config.App.ServeUnencryptedData
+	originalServeUnencryptedDataTrigger := config.Config.App.Crypt4GHPublicKeyB64
+	originalC4ghPrivateKeyFilepath := config.Config.App.Crypt4GHPrivateKey
```
jbygdell, line 406:
```suggestion
	originalServeUnencryptedDataTrigger := config.Config.C4GH.PublicKeyB64
	originalC4ghPrivateKeyFilepath := config.Config.C4GH.PrivateKey
```

sda-download/api/sda/sda_test.go
```diff
@@ -338,9 +341,8 @@ func TestFiles_Success(t *testing.T) {
 func TestDownload_Fail_UnencryptedDownloadNotAllowed(t *testing.T) {
 
 	// Save original to-be-mocked config
-	originalServeUnencryptedData := config.Config.App.ServeUnencryptedData
-
-	config.Config.App.ServeUnencryptedData = false
+	originalServeUnencryptedDataTrigger := config.Config.App.Crypt4GHPublicKeyB64
+	config.Config.App.Crypt4GHPublicKeyB64 = ""
```
jbygdell, line 345:
```suggestion
	originalServeUnencryptedDataTrigger := config.Config.C4GH.PublicKeyB64
	config.Config.C4GH.PublicKeyB64 = ""
```

sda-download/api/s3/s3.go
```diff
@@ -329,7 +330,7 @@ func Download(c *gin.Context) {
 		ListBuckets(c)
 
 	case c.Param("filename") != "":
-		if strings.HasPrefix(c.Request.URL.Path, "/s3-encrypted") {
+		if config.Config.App.Crypt4GHPublicKeyB64 == "" {
```
jbygdell, line 333:
```suggestion
		if config.Config.C4GH.PublicKeyB64 == "" {
```

sda-download/api/sda/sda.go
```diff
@@ -207,12 +207,22 @@ func Files(c *gin.Context) {
 
 // Download serves file contents as bytes
 func Download(c *gin.Context) {
-	if c.Param("type") != "encrypted" && !config.Config.App.ServeUnencryptedData {
+	// This conditional should always be satisfied for /s3 since the c.Param is set to  encrypted
+	// when Crypt4GHPublicKeyB64 is not set or empty, but this is not the case for calls to /files endpoint.
+	// So we need this check.
+
+	if c.Param("type") != "encrypted" && config.Config.App.Crypt4GHPublicKeyB64 == "" {
```
jbygdell, line 214:
```suggestion
	if c.Param("type") != "encrypted" && config.Config.CGH.PublicKeyB64 == "" {
```

nanjiangshu (approved):
Looks good!

aaperis:
All suggestions done so hopefully this can move forward.

P.S. It would have spared me some garbage time if [this](https://github.com/neicnordic/sensitive-data-archive/pull/1163#discussion_r1906858944) had been consistent with [this](https://github.com/neicnordic/sensitive-data-archive/pull/1163#discussion_r1906863320), but I guess that's on me :smile:.

MalinAhlberg (commented):
One last comment, then I'm ready to approve

sda-download/api/sda/sda.go
```diff
@@ -213,7 +213,7 @@ func Download(c *gin.Context) {
 	// So we need this check.
 	// Checking the type instead of the field S3 is better because it also provides a sanity check for the /s3 case.
 
-	if c.Param("type") != "encrypted" && config.Config.App.Crypt4GHPublicKeyB64 == "" {
+	if c.Param("type") != "encrypted" && config.Config.C4GH.PublicKeyB64 == "" {
```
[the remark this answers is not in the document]
aaperis: oo daang! I fixed this is now included in  https://github.com/neicnordic/sensitive-data-archive/pull/1163/commits/107a01aa426b1870b0d75bcc379730af431a5b65 :smile:

MalinAhlberg (approved):

sda-download/api/sda/sda_test.go
```diff
@@ -354,21 +356,13 @@ func TestDownload_Fail_UnencryptedDownloadNotAllowed(t *testing.T) {
...
-	}
+	assert.Equal(t, expectedStatusCode, response.StatusCode, "Unexpected status code from download")
+	assert.Equal(t, expectedBody, body, "Unexpected body from download")
 
 	// Test downloading of unencrypted file from the s3 endpoint, should fail
-	// encrypted
 	w = httptest.NewRecorder()
 	c, _ = gin.CreateTestContext(w)
-	c.Request = &http.Request{Method: "GET", URL: &url.URL{Path: "/s3/somepath", RawQuery: "filename=somepath"}}
+	c.Request = &http.Request{Method: "GET", URL: &url.URL{Path: "/mocks3/somepath", RawQuery: "filename=somepath"}}
```
jbygdell, line 365: Is this supposed to be `mocks3` ?

sda-download/internal/config/config.go
```diff
@@ -380,9 +379,17 @@ func (c *Map) appConfig() error {
...
 	var err error
-	c.App.Crypt4GHPrivateKey, c.App.Crypt4GHPublicKeyB64, err = GenerateC4GHKey()
-	if err != nil {
-		return err
+	if viper.GetString("c4gh.transientKeyPath") != "" {
+
+		if !viper.IsSet("c4gh.transientPassphrase") {
+			return errors.New("c4gh.transientPassphrase is not set")
+		}
+
```
jbygdell, line 387:
```suggestion
	if viper.IsSet("c4gh.transientKeyPath") && viper.IsSet("c4gh.transientPassphrase") {
```

No need to do extra checks since the error will come from the `GetC4GHKeys` function anyways.

aaperis: yes, but it is preferable to fail there imo

sda-download/api/sda/sda_test.go
jbygdell: Stupid tests ignore inputs..

aaperis: I just used the same naming as elsewhere in the tests with similar code. The tests could use a refactoring at some point but not here :smile:

nanjiangshu (approved):

sda-download/internal/config/config.go
jbygdell: Clearly we have a different view on what's preferable.

jbygdell (approved):

PR #1162, opened 2024-12-04
Docs: How to run SDA stack

nanjiangshu:
**Related issue(s) and PR(s)**  
This PR closes #1153 

**Changes made**
- Fixed some bugs in the Makefile
- Moved the instructions to run the SDA stack from GETTINGSTARTED.md to README.md and updated the instructions.

PR #1167, opened 2024-12-08
Run services manually with go run

### LINKED ISSUE #1155: Be able to run a service with go run
As a developer, I want to be able to run a service be just using go run and a configuration file so that to be faster to check things when I am working on a service.

**Acceptance criteria**
- [ ] Documentation on how to do that (example of how to do for one service)

**Additional context**
Everything else should be run with docker compose as usual except the one service where an ongoing work is happening.

Have a demo for the sprint review

nanjiangshu:
This PR closes #1155

**Description**

**Additional notes**
The bug for the integration test of sda-download has been there for quite some time. 

**How to test**
Follow the instructions in [DEVELOPMENT.md](https://github.com/neicnordic/sensitive-data-archive/blob/feature/manual-go-run/DEVELOPMENT.md) and verify that everything functions as expected.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -41,10 +41,10 @@ If successful, the curl command should output a JSON body containing:
 
 Running any of the SDA services located in the `sda` subfolder requires that the service specific credentials and RabbitMQ configurations are set as ENVs. Here, we'll use `ingest` as an example.
 
-- Bring up all SDA services with the S3 backend and populate them with test data by running the following command in the root folder of the repository:
+- Bring up all SDA services with the S3 backend by running the following command in the root folder of the repository:
 
 ```sh
-make integrationtest-sda-s3-run 
+make sda-s3-up
 ```
 
 - When the previous command is finished, bring down the `ingest` service using:
@@ -88,11 +88,14 @@ sda-cli -config /tmp/shared/s3cfg upload -encrypt-with-key /tmp/shared/c4gh.pub.
 # use sda-admin to check if t1.txt has been uploaded
 export API_HOST=http://localhost:8090
 export ACCESS_TOKEN=$(curl -s -k http://localhost:8080/tokens | jq -r '.[0]') 
-sda-admin file list -user test@dummy.org # file t1.txt should have fileStatus 'uploaded'
+sda-admin file list -user test@dummy.org # file test_dummy.org/t1.txt.c4gh should have fileStatus 'uploaded'
+
+# register the Crypt4GH key
+curl -H "Authorization: Bearer $ACCESS_TOKEN" -H "Content-Type: application/json" -X POST -d '{"pubkey": "'"$( base64 -w0 /tmp/shared/c4gh.pub.pem)"'", "description": "pubkey"}' http://localhost:8090/c4gh-keys/add
 
 # use sda-admin to ingest the file t1.txt
 sda-admin file ingest -filepath test_dummy.org/t1.txt.c4gh -user test@dummy.org  
 
 # verify that t1.txt has been ingested using sda-admin
-sda-admin file list -user test@dummy.org # file t1.txt should have fileStatus 'verified'
+sda-admin file list -user test@dummy.org # file test_dummy.org/t1.txt.c4gh should have fileStatus 'verified'
 ```
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -8,25 +8,16 @@ This section explains how to run some of the services using `go run` instead of 
 make integrationtest-sda-s3-run 
 ```
 
-2. Change to the folder `sda-download` and then create the credentials using: 
-```sh
-pushd dev_utils
-bash make_certs.sh
-chmod 600 certs/*-key.*
-popd
-```
-
-3. Update your `/etc/hosts` file: Add the following line to ensure the service can resolve OIDC locally: 
-```
-127.0.0.1 oidc
-```
-
-4. Start the `sda-download` service using:
+2. Change to the folder `sda-download` and start the `sda-download` service using:
 ```sh
 CONFIGFILE=dev_utils/config-notls_local.yaml go run cmd/main.go
 ```
+> If the host `oidc` cannot be accessed on the local machine, update your `/etc/hosts` file by adding the following line to ensure the service can resolve `oidc` locally:
+> ```
+> 127.0.0.1 oidc
+> ```
 
-5. Check if `sda-download` works as expected using:
+3. Check if `sda-download` works as expected using:
 ```sh
 curl -o /dev/null -s -w "%{http_code}\n" http://localhost:18080/health
 ```
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,59 +1,68 @@
-## Run services with `go run`
+# Run services with `go run`
 
 This section explains how to run some of the services using `go run` instead of the Docker setup to facilitate development.
 
-### Running `sda-download` with `go run`
-1. Bring up all SDA services with the S3 backend and populate them with test data by running the following command in the root folder of the repository:
+## Running `sda-download` with `go run`
+
+- Bring up all SDA services with the S3 backend and populate them with test data by running the following command in the root folder of the repository:
+
 ```sh
 make integrationtest-sda-s3-run 
 ```
 
-2. Change to the folder `sda-download` and start the `sda-download` service using:
+- Change to the folder `sda-download` and start the `sda-download` service using:
+
 ```sh
 CONFIGFILE=dev_utils/config-notls_local.yaml go run cmd/main.go
 ```
-> If the host `oidc` cannot be accessed on the local machine, update your `/etc/hosts` file by adding the following line to ensure the service can resolve `oidc` locally:
-> ```
-> 127.0.0.1 oidc
-> ```
 
-3. Check if `sda-download` works as expected using:
+- Check if `sda-download` works as expected using:
+
 ```sh
 curl -o /dev/null -s -w "%{http_code}\n" http://localhost:18080/health
 ```
+
 If successful, the curl command should output the HTTP code `200`.
 
 You can further check the endpoint `/metadata/datasets` using:
+
 ```sh
 token=$(curl -s -k http://localhost:8080/tokens | jq -r '.[0]') 
 curl -H "Authorization: Bearer $token" http://localhost:18080/metadata/datasets
 ```
+
 If successful, the curl command should output a JSON body containing:
+
 ```json
 ["EGAD74900000101"]
 ```
 
+## Running other SDA services with `go run`
 
-### Running other SDA services with `go run`
-Running other SDA services located in the `sda` subfolder, such as `ingest` or `verify`, differs slightly from running the `sda-download` service. Here, we'll use `ingest` as an example.
+Running any of the SDA services located in the `sda` subfolder requires that the service specific credentials and RabbitMQ configurations are set as ENVs. Here, we'll use `ingest` as an example.
 
-1. Bring up all SDA services with the S3 backend and populate them with test data by running the following command in the root folder of the repository:
+- Bring up all SDA services with the S3 backend and populate them with test data by running the following command in the root folder of the repository:
+
 ```sh
 make integrationtest-sda-s3-run 
 ```
 
-2. Bring down the `ingest` service using:
+- When the previous command is finished, bring down the `ingest` service using:
+
 ```sh
-PR_NUMBER=$(/bin/date +%F)  docker compose -f .github/integration/sda-s3-integration.yml stop ingest  
+docker stop ingest
 ```
 
-3. Change to the folder `sda`, then copy keys and other information from the shared folder of the container using:
+- Copy keys and other information from the shared folder of the container using:
+
 ```sh
 docker cp verify:/shared /tmp/
 ```
-This will copy all data from the container's `/shared` folder to `/tmp/shared` on your local machine.
 
-4. Start the `ingest` service using:
+This will copy all data from the container's `/shared` folder to `/tmp/shared` on your local machine, so that we have access to all the required auto generated files that will be required.
+
+- Change to the folder `sda` and start the `ingest` service using:
+
 ```sh
 export BROKER_PASSWORD=ingest
 export BROKER_USER=ingest
@@ -64,7 +73,8 @@ export DB_USER=ingest
 CONFIGFILE=config_local.yaml go run cmd/ingest/ingest.go
 ```
 
-5. Check if the `ingest` service works as expected by following these steps
+- Check if the `ingest` service works as expected by following these steps
+
 ```sh
 # create a test file
 seq 10 > /tmp/t1.txt
@@ -83,6 +93,6 @@ sda-admin file list -user test@dummy.org # file t1.txt should have fileStatus 'u
 # use sda-admin to ingest the file t1.txt
 sda-admin file ingest -filepath test_dummy.org/t1.txt.c4gh -user test@dummy.org  
 
-# verify that t1.txt has been uploaded using sda-admin 
+# verify that t1.txt has been ingested using sda-admin
 sda-admin file list -user test@dummy.org # file t1.txt should have fileStatus 'verified'
 ```
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -23,13 +23,19 @@ popd
 
 4. Start the `sda-download` service using:
 ```sh
-APP_PORT=8553 CONFIGFILE=dev_utils/config-notls_local.yaml go run cmd/main.go
+CONFIGFILE=dev_utils/config-notls_local.yaml go run cmd/main.go
 ```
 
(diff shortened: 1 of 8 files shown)


DEVELOPMENT.md
```diff
@@ -0,0 +1,97 @@
...
+If successful, the curl command should output a JSON body containing:
+```json
+["EGAD74900000101"]
+```
+
+
+### Running other SDA services with `go run`
+Running other SDA services located in the `sda` subfolder, such as `ingest` or `verify`, differs slightly from running the `sda-download` service. Here, we'll use `ingest` as an example.
+
+1. Bring up all SDA services with the S3 backend and populate them with test data by running the following command in the root folder of the repository (this step is the same as for `sda-download`):
```
jbygdell, line 43: If you describe how to run a single command from the sda stack there is no need to reference `sda-download` and compare the differences. Keep it strictly for the sda stack.

DEVELOPMENT.md
```diff
@@ -0,0 +1,97 @@
...
+2. Bring down the `ingest` service using:
+```sh
+PR_NUMBER=$(/bin/date +%F)  docker compose -f .github/integration/sda-s3-integration.yml stop ingest  
+```
+
+3. Change to the folder `sda`, then copy keys and other information from the shared folder of the container using:
+```sh
+mkdir -p tmp
+docker cp verify:/shared tmp/
+```
```
jbygdell, line 57: Do not copy temp files into the current workdir, always use the system temp folder.

```suggestion
3. Change to the folder `sda`, then copy keys and other information from the shared folder of the container using:
```sh
docker cp verify:/shared /tmp/shared
```
```

File: DEVELOPMENT.md [hunk 3]
```diff
@@ -0,0 +1,97 @@
...
+This will copy all data from the container's `/shared` folder to `tmp/shared` on your local machine.
```
jbygdell, line 58:
```suggestion
This will copy all data from the container's `/shared` folder to `/tmp/shared` on your local machine.
```

DEVELOPMENT.md
```diff
@@ -0,0 +1,97 @@
...
+
+4. Copy schemas to `/schemas` on the localhost using:
+```sh
+sudo rsync -arvz schemas/ /schemas/ 
+```
+
```
jbygdell, line 64:
```suggestion
```

This should not be needed, the schema path is a configurable option, although not directly so that is something we should fix to make this easier.

DEVELOPMENT.md
```diff
@@ -0,0 +1,97 @@
...
+CONFIGFILE=config_local.yaml go run cmd/ingest/ingest.go
+```
+
+6. Check if the `ingest` service works as expected by following these steps
+```sh
+# create a test file
+seq 10 > tmp/t1.txt  
+
+# update the s3cmd config file
+sed -i '/host_/s/s3inbox:8000/localhost:18000/g' tmp/shared/s3cfg
```
jbygdell, line 82:
```suggestion
sed -i '/host_/s/s3inbox:8000/localhost:18000/g' /tmp/shared/s3cfg
```

DEVELOPMENT.md
```diff
@@ -0,0 +1,97 @@
...
+
+# upload tmp/t1.txt to s3inbox by sda-cli
+sda-cli -config tmp/shared/s3cfg upload -encrypt-with-key tmp/shared/c4gh.pub.pem tmp/t1.txt
```
jbygdell, line 85:
```suggestion
sda-cli -config /tmp/shared/s3cfg upload -encrypt-with-key /tmp/shared/c4gh.pub.pem tmp/t1.txt 
```

DEVELOPMENT.md
```diff
@@ -0,0 +1,97 @@
...
+export BROKER_ROUTINGKEY=archived
+export DB_PASSWORD=ingest
+export DB_USER=ingest 
+CONFIGFILE=config_local.yaml go run cmd/ingest/ingest.go
+```
+
+6. Check if the `ingest` service works as expected by following these steps
+```sh
+# create a test file
+seq 10 > tmp/t1.txt
```
jbygdell, line 79:
```suggestion
seq 10 > /tmp/t1.txt  
```

jbygdell (changes requested):

DEVELOPMENT.md
nanjiangshu: The problem is nontrivial. Currently, the `SchemasPath` is hardcoded based on the value of `schema.type`:
https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/internal/config/config.go#L919-L925

I have tried modifying this function to set `SchemasPath` to a custom path, but it still doesn't work.

We should create an issue to make `SchemasPath` customizable. In the meantime, we need to manually copy the schemas to make it work.

DEVELOPMENT.md
```diff
@@ -0,0 +1,97 @@
...
+bash make_certs.sh
+chmod 600 certs/*-key.*
+popd
+```
+
+3. Update your `/etc/hosts` file: Add the following line to ensure the service can resolve OIDC locally: 
+```
+127.0.0.1 oidc
+```
+
```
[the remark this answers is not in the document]
nanjiangshu: Do you mean something like this in `sda-s3-integration.yml`? Doesn't work. 

```yaml
  oidc:
    container_name: oidc
    command:
      - /bin/sh
      - -c
      - |
        pip install --upgrade pip
        pip install aiohttp Authlib joserfc requests
        python -u /oidc.py
    depends_on:
      credentials:
        condition: service_completed_successfully
    healthcheck:
      test: ["CMD", "python3", "-c", 'import requests; print(requests.get(url = "http://localhost:8080/jwk").text)']
      interval: 10s
      timeout: 2s
      retries: 6
    image: python:3.11-slim
    ports:
      - "8080:8080"
    extra_hosts:
      - "oidc:host-gateway"
    restart: always
    volumes:
      - ./sda/oidc.py:/oidc.py
      - shared:/shared
```

DEVELOPMENT.md
jbygdell:
```
func (c *Config) configSchemas() {
	if viper.GetString("schema.type") == "federated" {
		c.Broker.SchemasPath = "/schemas/federated/"
	} else {
		c.Broker.SchemasPath = "/schemas/isolated/"
	}

	if viper.IsSet("schema.path") {
		c.Broker.SchemasPath = viper.GetString("schema.path")
	}
}
```

DEVELOPMENT.md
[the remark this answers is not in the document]
jbygdell:
```yaml
extra_hosts:
  - "localhost:host-gateway"
```

In the correct place will make things accessible in the same way from the developers perspective as from a container. If you are having issues, look in the GDI repo for hints on how to get it to work.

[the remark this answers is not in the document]
nanjiangshu: I think `host-gateway` can solve the problem to access a localhost from inside a container but not to access e.g. http://oidc:8080 from outside of the container, which is the case when we run `sda-download` with `go run`, right`? So this solution won't work.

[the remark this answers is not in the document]
jbygdell: The whole point with the `localhost:host-gateway` mapping is to make a container callable by the same name from both the shell of the developer and from within another container.

It's possible to use another name than localhost, as long as that name exists in the `etc/hosts` file of the host machine, like for example the value of `$HOSTNAME`.

DEVELOPMENT.md
nanjiangshu: This works well. When I tested yesterday, I think the reason why it failed was because the MQ message are not cleaned and therefore the old schema path was used still. 
Fixed in the commits 17e9d1d2 and 73f8a274

DEVELOPMENT.md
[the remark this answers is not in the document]
nanjiangshu: For some reason, the solutions for host-gateway, e.g. in this PR https://github.com/GenomicDataInfrastructure/starter-kit-storage-and-interfaces/pull/57/files, never worked on my Linux setup. Neither for Ubuntu 22.04 nor 18.04 on SNIC cloud or EGI cloud. If it works for you, cloud you please make the changes to `.github/integration/sda-s3-integration.yml`? Thanks.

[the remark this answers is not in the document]
jbygdell: I can look in to it once i'm in a place to commit things for the k8s documentation.

sda/config_local.yaml
```diff
@@ -0,0 +1,108 @@
...
+auth:
+  cega:
+    authUrl: "http://localhost:8443/username/"
+    id:
+    secret:
+  infoText: "About service text"
+  infoURL: "http://example.org/about"
+  jwt:
+    issuer: "https://localhost:8888"
+    privateKey: "tmp/shared/keys/jwt.key"
```
MalinAhlberg, line 26:
```suggestion
    privateKey: "/tmp/shared/keys/jwt.key"
```
And similar for the other cases.
Needed for me to make the services run.

MalinAhlberg (approved):
Great! Really nice to have this documented! Everything works well for me, except for that I need to add a `/` before `tmp` in the conf (see comment).

sda/config_local.yaml
nanjiangshu: Good catch! fixed in deaf0a34

DEVELOPMENT.md
```diff
@@ -0,0 +1,97 @@
...
+make integrationtest-sda-s3-run 
+```
+
+2. Change to the folder `sda-download` and then create the credentials using: 
+```sh
+pushd dev_utils
+bash make_certs.sh
+chmod 600 certs/*-key.*
+popd
+```
```
jbygdell, line 17:
```suggestion
```

Certificates are not needed since it specifically says to use the `no_tls` config file

DEVELOPMENT.md
```diff
@@ -0,0 +1,88 @@
...
+If successful, the curl command should output a JSON body containing:
+```json
+["EGAD74900000101"]
+```
+
+
+### Running other SDA services with `go run`
+Running other SDA services located in the `sda` subfolder, such as `ingest` or `verify`, differs slightly from running the `sda-download` service. Here, we'll use `ingest` as an example.
+
```
jbygdell, line 39:
```suggestion

### Running a separate SDA service with `go run`
Running any of the SDA services located in the `sda` subfolder requires that the service specific credentials and RabbitMQ configurations are set as ENVs. Here, we'll use `ingest` as an example.

```

DEVELOPMENT.md
```diff
@@ -0,0 +1,88 @@
...
+1. Bring up all SDA services with the S3 backend and populate them with test data by running the following command in the root folder of the repository:
+```sh
+make integrationtest-sda-s3-run
```
jbygdell, line 42:
```suggestion
make sda-s3-up
```

This should probably use just the startup command that is part of PR #1162  since this will start the test script and shutting down ingest later will cause it to fail.

DEVELOPMENT.md
```diff
@@ -0,0 +1,88 @@
...
+Running other SDA services located in the `sda` subfolder, such as `ingest` or `verify`, differs slightly from running the `sda-download` service. Here, we'll use `ingest` as an example.
+
+1. Bring up all SDA services with the S3 backend and populate them with test data by running the following command in the root folder of the repository:
+```sh
+make integrationtest-sda-s3-run 
+```
+
+2. Bring down the `ingest` service using:
+```sh
+PR_NUMBER=$(/bin/date +%F)  docker compose -f .github/integration/sda-s3-integration.yml stop ingest
```
jbygdell, line 47:
```suggestion
docker stop ingest  
```

You only need to call docker stop since the containers have unique names.

DEVELOPMENT.md
```diff
@@ -0,0 +1,88 @@
...
+2. Bring down the `ingest` service using:
+```sh
+PR_NUMBER=$(/bin/date +%F)  docker compose -f .github/integration/sda-s3-integration.yml stop ingest  
+```
+
+3. Change to the folder `sda`, then copy keys and other information from the shared folder of the container using:
+```sh
+docker cp verify:/shared /tmp/
+```
+This will copy all data from the container's `/shared` folder to `/tmp/shared` on your local machine.
```
jbygdell, line 54:
```suggestion
This will copy all data from the container's `/shared` folder to `/tmp/shared` on your local machine, so that we have access to all the required auto generated files that will be required.
```

DEVELOPMENT.md
```diff
@@ -0,0 +1,88 @@
...
+
+# use sda-admin to check if t1.txt has been uploaded
+export API_HOST=http://localhost:8090
+export ACCESS_TOKEN=$(curl -s -k http://localhost:8080/tokens | jq -r '.[0]') 
+sda-admin file list -user test@dummy.org # file t1.txt should have fileStatus 'uploaded'
+
+# use sda-admin to ingest the file t1.txt
+sda-admin file ingest -filepath test_dummy.org/t1.txt.c4gh -user test@dummy.org  
+
+# verify that t1.txt has been uploaded using sda-admin
```
jbygdell, line 86:
```suggestion
# verify that t1.txt has been ingested using sda-admin 
```

jbygdell (changes requested):
The DEVELOPMENT.md file has a ver different layout compared to the other markdown files. It would be better if this file also followed the same layout i.e remove all the numbered list items for starter.
Secondly it would be much easier if you started using a markdown linter in order to catch all the other issues, mainly missing empty lines between sections and trailing white spaces.

DEVELOPMENT.md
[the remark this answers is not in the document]
jbygdell: I pushed a commit that makes this work from my local shell.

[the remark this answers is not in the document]
nanjiangshu: OK. So I will keep the instruction for adding "127.0.0.1 oidc" to /etc/hosts

DEVELOPMENT.md
nanjiangshu: If you just bring up the services with `sda-s3-up`, the verification test mentioned in the last point will not work. Also, if you wait until `make integrationtest-sda-s3-run` is finished, running docker stop ingest will not cause it to fail. Perhaps I should add a note to wait until the command finishes?

DEVELOPMENT.md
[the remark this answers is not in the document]
jbygdell: No it should be removed, my commit fixed it.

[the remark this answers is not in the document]
nanjiangshu: fixed in 7f1134bb

DEVELOPMENT.md
nanjiangshu: fixed in 7f1134bb

DEVELOPMENT.md
nanjiangshu: fixed in 7f1134bb

DEVELOPMENT.md
nanjiangshu: fixed in 7f1134bb

DEVELOPMENT.md
nanjiangshu: fixed in 7f1134bb

DEVELOPMENT.md
nanjiangshu: I create a bug report https://github.com/neicnordic/sensitive-data-archive/issues/1184

jbygdell: It's not a bug, it's you who haven't registered the key hash in the DB before you try to ingest the file.

nanjiangshu: Thanks for pointing this out. Fixed in the commit 2cefebe

DEVELOPMENT.md
```diff
@@ -0,0 +1,101 @@
...
+docker stop ingest
+```
+
+- Copy keys and other information from the shared folder of the container using:
+
+```sh
+docker cp verify:/shared /tmp/
+```
+
+This will copy all data from the container's `/shared` folder to `/tmp/shared` on your local machine, so that we have access to all the required auto generated files that will be required.
```
jbygdell, line 62:
```suggestion
This will copy all data from the container's `/shared` folder to `/tmp/shared` on your local machine, so that we have access to all the auto generated files that will be required.
```

jbygdell (approved):

DEVELOPMENT.md
```diff
@@ -0,0 +1,101 @@
+# Run services with `go run`
+
+This section explains how to run some of the services using `go run` instead of the Docker setup to facilitate development.
+
+## Running `sda-download` with `go run`
+
+- Bring up all SDA services with the S3 backend and populate them with test data by running the following command in the root folder of the repository:
+
+```sh
+make integrationtest-sda-s3-run
```
aaperis, line 10: This command will build everything and run the tests and in the end will remove all containers.
So, there won't be anything running to test with.

nanjiangshu: The command `make integration-sda-s3` will remove all containers when it is finished, but not this one 😄 
see https://github.com/neicnordic/sensitive-data-archive/blob/main/Makefile#L118-L119

aaperis: Ooh, right! Should have read more carefully :blush: :

Issue #901, opened 2024-06-11
Sda-doa integration


Parisa68:
- The entire `.github` folder must be removed so we don't get any conflicts with regards to depenabot and such.
- The relevant parts of the workflows should be merged into the corresponding file of the root of the repo.

- The docker compose files should be moved into `.github/integration` in the root of the repo and be reworked similar to the existing compose files.
 
- Keys, certificates and such should be generated dynamically if needed and not be committed to the repository.
 
- Relevant options for this app should be added to the `Makefile`

_Originally posted by @jbygdell in https://github.com/neicnordic/sensitive-data-archive/pull/876#pullrequestreview-2087372218_

PR #1187, opened 2024-12-11
Last commit to sync with the old SDA-DOA repo

Parisa68:
Last commit to sync with the old repo. contains:
- use reference id for exporting data
- Change crypt4gh and clearinghouse settings to use fega-norway repo
- rename vars to increase readability map the datasets with reference to datasets stable id
- Handle the cases where the visas are a mix of Fega and non-fega visas
- add tests for dataset reference
- add tests for visa token

## Related issue(s): 
#901 
.


kjellp (approved):
LGTM

sda-doa/test/mock_auth.py
```diff
@@ -143,7 +158,8 @@ async def userinfo(request):
                 DATA[3],
                 DATA[4],
                 DATA[5],
-                DATA[6]
+                DATA[6],
+                DATA[7]
```
joshbaskaran, line 162: Not worth blocking the PR over this but perhaps adding a comment as to what the expected order of tokens might help us troubleshoot down the line should things change.

joshbaskaran (approved):

PR #1189, opened 2024-12-12
[ingest] Check for keyhash on startup

### LINKED ISSUE #1164: ingest logs fills up memory with error logs
**Describe the bug**
When running the data_loader in starter-kit-storage-and-interfaces, no key hash is inserted before ingestion starts. That makes ingest loops infinitely, eating all memory while producing logs like this:
```
ingest  | {"level":"info","msg":"Received work (corr-id: 5dd81f8c-9c5c-46a1-ae90-09d13da1dac6, filepath: htsnexus_test_NA12878.bam.c4gh, user: dummy@gdi.eu)","time":"2024-12-06T08:04:56Z"}
ingest  | {"level":"error","msg":"Key hash a7db730b021d942d12ac2a8d02ba77c8c91fe4ca61681f9e1450536c546ae72a could not be set for fileID 5dd81f8c-9c5c-46a1-ae90-09d13da1dac6: (pq: insert or update on table \"files\" violates foreign key constraint \"files_key_hash_fkey\")","time":"2024-12-06T08:04:57Z"}
```

**Steps to reproduce**
Run `docker compose -f docker-compose-demo.yml up -d` in `starter-kit-storage-and-interfaces`, or run the steps manually without the starter kit.

**Expected behavior**
- [ ] One error is reported
- [ ] Logs are not exploding

**Additional context**
See also #1037 and #1027

jbygdell:
## Related issue(s) and PR(s)
This PR closes #1164 .

## Description
Check that a keyhash exists in the database before start processing any messages. A timer counts down for 5 minutes and if no keyhash exists by then an error will be logged and ingest will restart.

No messages will be processed before this check completes.

## How to test
make build-all
make sda-s3-up
docker logs -f ingest

"no crypt4gh key hash registered" will be printed every 30 seconds and after 5 minutes the container will restart.

If `make integrationtest-sda` is used ingest will Print the same message a few times until the tests reaches `11_api_test`. Then it will process the messages as normal for the ingestion test.

### DIFF
--- a/sda/cmd/ingest/ingest.go
+++ b/sda/cmd/ingest/ingest.go
@@ -14,6 +14,7 @@ import (
 	"os/signal"
 	"strings"
 	"syscall"
+	"time"
 
 	"github.com/neicnordic/crypt4gh/keys"
 	"github.com/neicnordic/crypt4gh/model/headers"
@@ -100,6 +101,24 @@ func main() {
 	var message schema.IngestionTrigger
 
 	go func() {
+		start := time.Now()
+		for i := 1; i > 0; i++ {
+			h, err := db.ListKeyHashes()
+			if err != nil {
+				log.Errorln(err.Error())
+			}
+			if len(h) != 0 {
+				break
+			}
+
+			time.Sleep(time.Duration(30 * time.Second))
+			if time.Since(start).Seconds() >= float64(300) {
+				log.Errorln("no crypt4gh key hash registered, restarting")
+				forever <- false
+			}
+			log.Errorln("no crypt4gh key hash registered")
+		}
+
 		messages, err := mq.GetMessages(conf.Broker.Queue)
 		if err != nil {
 			log.Fatal(err)
--- a/.github/integration/scripts/charts/dependencies.sh
+++ b/.github/integration/scripts/charts/dependencies.sh
@@ -33,7 +33,7 @@ fi
 # secret for the crypt4gh keypair
 C4GHPASSPHRASE="$(random-string)"
 export C4GHPASSPHRASE
-dir=$CWD
+dir=$PWD
 if [ -n "$1" ]; then
         dir=/tmp
 fi
--- a/.github/integration/scripts/charts/k3d.sh
+++ b/.github/integration/scripts/charts/k3d.sh
@@ -16,3 +16,6 @@ sudo mv ./kubectl /usr/local/bin/kubectl
 k3d cluster create sda --image=rancher/k3s:"$k8s"-k3s1 --wait --timeout 10m
 k3d kubeconfig merge sda --kubeconfig-switch-context
 mkdir -p ~/.kube/ && cp ~/.config/kubeconfig-sda.yaml ~/.kube/config
+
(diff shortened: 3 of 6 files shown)


sda/cmd/ingest/ingest.go
```diff
@@ -100,6 +101,24 @@ func main() {
 	var message schema.IngestionTrigger
 
 	go func() {
+		start := time.Now()
+		for i := 0; i <= 10; i++ {
```
nanjiangshu, line 105: Maybe better to define a variable or constant instead of hard coding the number of loops.

sda/cmd/ingest/ingest.go
```diff
@@ -100,6 +101,24 @@ func main() {
...
+			h, err := db.ListKeyHashes()
+			if err != nil {
+				log.Errorln(err.Error())
+			}
+			if len(h) != 0 {
+				break
+			}
+
+			time.Sleep(time.Duration(30 * time.Second))
```
nanjiangshu, line 114: So as the sleep duration, better to avoid hard coding

sda/cmd/ingest/ingest.go
```diff
@@ -100,6 +101,24 @@ func main() {
...
+			if time.Since(start).Seconds() >= float64(300) {
+				log.Errorln("no crypt4gh key hash registered, restarting")
+				forever <- false
+			}
+			log.Errorln("no crypt4gh key hash registered")
```
nanjiangshu, line 119: This message will be repeated, that is, for every 30 seconds, the message is printed twice. Move it to the line in front of `forever <- false`

sda/cmd/ingest/ingest.go
```diff
@@ -100,6 +101,24 @@ func main() {
...
 	go func() {
+		start := time.Now()
+		for i := 0; i <= 10; i++ {
+			h, err := db.ListKeyHashes()
+			if err != nil {
+				log.Errorln(err.Error())
+			}
+			if len(h) != 0 {
+				break
+			}
```
nanjiangshu, line 112:
```suggestion
			if err != nil {
				log.Errorln(err.Error())
			} else if len(h) != 0 {
				break
			}
```

Without if else, it assumes that ListKeyHashes() returns a valid slice even if an error occurs. This may lead to problem  in edge cases.

nanjiangshu (commented):
Except for some minor comments, the PR works as expected. Great job!

sda/cmd/ingest/ingest.go
jbygdell: `ListKeyHashes()` always returns a valid slice according to the DB tests.

sda/cmd/ingest/ingest.go
jbygdell: Sleep is there so as not to put a strain on the DB connection and not spam the log messages.

sda/cmd/ingest/ingest.go
jbygdell: It's printed once every 30s for 300s then the container will restart.

nanjiangshu (dismissed):
Looks good!

jbygdell:
Rebased so needs to be reapproved.

aaperis (approved):
LGTM

nanjiangshu (approved):

PR #1228, opened 2024-12-30
[internal/database] GetCorrID needs to accept multiple rows as result

jbygdell:
## Description

QueryRow selects the first result returned, but the file_event_log always will return multiple rows. This is due to the `register_file` function  in our DB schema that stores the first log event without any correlation ID.
Combine that with the quasi randomness that Postgres returns if no sorting is applied and we have the current state.

Running the commands below on master failed 5 out of 6 times on average.

## How to test

The following execution should not fail.
```cmd
make build-all
make integrationtest-sda-s3-run
sda-admin/sda-admin file ingest -filepath requester_demo.org/data/file1.c4gh -user requester@demo.org
```

### DIFF
--- a/sda/internal/database/db_functions.go
+++ b/sda/internal/database/db_functions.go
@@ -732,9 +732,6 @@ func (dbs *SDAdb) getCorrID(user, path, accession string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	if rows.Err() != nil {
-		return "", rows.Err()
-	}
 	defer rows.Close()
 
 	var corrID sql.NullString
@@ -746,6 +743,9 @@ func (dbs *SDAdb) getCorrID(user, path, accession string) (string, error) {
 		if corrID.Valid {
 			return corrID.String, nil
 		}
+	}
+	if rows.Err() != nil {
+		return "", rows.Err()
 	}
 
 	return "", fmt.Errorf("sql: no rows in result set")
--- a/sda/internal/database/db_functions.go
+++ b/sda/internal/database/db_functions.go
@@ -6,6 +6,7 @@ import (
 	"database/sql"
 	"encoding/hex"
 	"errors"
+	"fmt"
 	"math"
 	"strings"
 	"time"
@@ -725,15 +726,29 @@ func (dbs *SDAdb) getCorrID(user, path, accession string) (string, error) {
 	dbs.checkAndReconnectIfNeeded()
 	db := dbs.DB
 	const query = "SELECT DISTINCT correlation_id FROM sda.file_event_log e " +
-		"RIGHT JOIN sda.files f ON e.file_id = f.id WHERE f.submission_file_path = $1 AND f.submission_user = $2 AND COALESCE(f.stable_id, '')= $3;"
-
-	var corrID string
-	err := db.QueryRow(query, path, user, accession).Scan(&corrID)
+		"RIGHT JOIN sda.files f ON e.file_id = f.id " +
+		"WHERE f.submission_file_path = $1 AND f.submission_user = $2 AND COALESCE(f.stable_id, '') = $3;"
+	rows, err := db.Query(query, path, user, accession)
 	if err != nil {
 		return "", err
 	}
+	if rows.Err() != nil {
+		return "", rows.Err()
+	}
+	defer rows.Close()
 
-	return corrID, nil
+	var corrID sql.NullString
+	for rows.Next() {
+		err := rows.Scan(&corrID)
+		if err != nil {
+			return "", err
+		}
+		if corrID.Valid {
+			return corrID.String, nil
+		}
+	}
+
+	return "", fmt.Errorf("sql: no rows in result set")
 }
 
 // list all users with files not yet assigned to a dataset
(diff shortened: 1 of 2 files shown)


pahatz (approved):
It seems to work well.
fyi, I updated the description for a typo.

sda/internal/database/db_functions.go
```diff
@@ -725,15 +726,29 @@ func (dbs *SDAdb) getCorrID(user, path, accession string) (string, error) {
...
-	err := db.QueryRow(query, path, user, accession).Scan(&corrID)
+		"RIGHT JOIN sda.files f ON e.file_id = f.id " +
+		"WHERE f.submission_file_path = $1 AND f.submission_user = $2 AND COALESCE(f.stable_id, '') = $3;"
+	rows, err := db.Query(query, path, user, accession)
 	if err != nil {
 		return "", err
 	}
+	if rows.Err() != nil {
+		return "", rows.Err()
+	}
```
nanjiangshu, line 737: Checking `rows.Err()` should be placed in the code block 
```go
for rows.Next() {
		err := rows.Scan(&corrID)
		if err != nil {
			return "", err
		}
		if corrID.Valid {
			return corrID.String, nil
		}
	}
```

nanjiangshu:
```suggestion
```

sda/internal/database/db_functions.go
```diff
@@ -725,15 +726,29 @@ func (dbs *SDAdb) getCorrID(user, path, accession string) (string, error) {
...
+	var corrID sql.NullString
+	for rows.Next() {
+		err := rows.Scan(&corrID)
+		if err != nil {
+			return "", err
+		}
+		if corrID.Valid {
+			return corrID.String, nil
+		}
+	}
```
nanjiangshu, line 749:
```suggestion
	for rows.Next() {
		err := rows.Scan(&corrID)
		if err != nil {
			return "", err
		}
		if rows.Err() != nil {
		        return "", rows.Err()
	        }
		if corrID.Valid {
			return corrID.String, nil
		}
	}
```

sda/internal/database/db_functions.go
```diff
@@ -725,15 +726,29 @@ func (dbs *SDAdb) getCorrID(user, path, accession string) (string, error) {
...
 
-	return corrID, nil
+	return "", fmt.Errorf("sql: no rows in result set")
```
nanjiangshu, line 751: Maybe better to provide more info e.g. `path`, `user`, `accession` in the error message?

aaperis (approved):
Nice.

sda/internal/database/db_functions.go
jbygdell: At this point the eror should be as it is since this is used in multiple places, any extra information needs to be added (if necessary) at the stage where the eror is handled by the calling function.

sda/internal/database/db_functions_test.go
```diff
@@ -529,6 +529,23 @@ func (suite *DatabaseTests) TestGetCorrID_sameFilePath() {
...
+	filePath := "/testuser/file10.c4gh"
+	user := "testuser"
+
+	fileID, err := db.RegisterFile(filePath, user)
+	assert.NoError(suite.T(), err, "failed to register file in database")
+	err = db.UpdateFileEventLog(fileID, "uploaded", fileID, user, "{}", "{}")
+	assert.NoError(suite.T(), err, "failed to update satus of file in database")
+
+	corrID, err := db.GetCorrID(user, "/testuser/file20.c4gh", "")
+	assert.EqualError(suite.T(), err, "sql: no rows in result set")
```
nanjiangshu, line 545: Maybe change to a more generic assertion, e.g. 
```go
assert.Error(suite.T(), err, "expected an error for a nonexistent file path")
```
Since otherwise, this always need to be changed in case the error message from the function `GetCorrID` is changed in the future.

sda/internal/database/db_functions.go
jbygdell: Checking for `rows:Err()` only needs to be done once, on the outside, as per library documentation.

nanjiangshu (commented):
Works great! However, I have a concern about the placement of the `rows.Err()` check. Please see my inline comment.

sda/internal/database/db_functions_test.go
jbygdell: That error should be an SQL error, so unless the library changes nothings needs to be done.

sda/internal/database/db_functions_test.go
```diff
@@ -529,6 +529,23 @@ func (suite *DatabaseTests) TestGetCorrID_sameFilePath() {
...
+	db, err := NewSDAdb(suite.dbConf)
+	assert.NoError(suite.T(), err, "got (%v) when creating new connection", err)
+
+	filePath := "/testuser/file10.c4gh"
+	user := "testuser"
+
+	fileID, err := db.RegisterFile(filePath, user)
+	assert.NoError(suite.T(), err, "failed to register file in database")
+	err = db.UpdateFileEventLog(fileID, "uploaded", fileID, user, "{}", "{}")
+	assert.NoError(suite.T(), err, "failed to update satus of file in database")
```
MalinAhlberg, line 542:
```suggestion
	assert.NoError(suite.T(), err, "failed to update status of file in database")
```

MalinAhlberg (approved):
Nice :clap: . I found a small typo, otherwise it looks good!

kostas-kou (approved):

nanjiangshu (approved):
LGTM!

PR #1235, opened 2025-01-07
`verify` supports multiple private keys

### LINKED ISSUE #1087: [verify] Support multiple encryption keys
# Main user story
As a security conscious user of the sensitive data arcive,
I want to see that the archive can rotate encryption keys
so that security can be maintained.

## Sub user story
As an sda-dev
I want to implement multiple-key support for the verify services
so that repository key rotation can be enabled.

# Description

* Verify should be able to load a directory of public/private key pairs instead of just a specific one (same configuration as ingest).
* Since verify reads all the keys from the directory should keep all the key hashes in memory for finding the private key to use without trying all the private keys. 
* For verify to pass for a file it should look up the hash of the key used to encrypt a file from the database and should then try to decrypt the file with that specific key.
* 

# Acceptance criteria

* Demonstrate ingestion and verification of encrypted files with two different archive c4gh private keys.
* Tests verifying the changes added

# Extra information

This issue was split off from issue #787.

Note: key rotation miro board [here](https://miro.com/app/board/uXjVNz-UuhY=/)

nanjiangshu:
## Related issue(s) and PR(s)
This PR closes #1087 

## Description
Verify reads a list of key files and their corresponding passphrases provided in the config file under `c4gh.keys`, for example
```yaml
c4gh:
   privateKeys:
   - filePath: /shared/c4gh.sec.pem
     passphrase: "c4ghpass"
   - filePath: /shared/c4gh1.sec.pem
     passphrase: "c4ghpass"
```

The old method, which provides only one key file and its corresponding passphrase, is still retained for components other than `ingest` and `verify`. This method should be removed once all components have been updated to support multiple keys. 

In addition, setting multiple keys and passphrases via environmental variables is not supported for the moment. This will be addressed when all components are updated. The README for each service should be updated accordingly at last. 

## How to test
Verify that the integration tests pass.

If you want to test manually, do as follows
- Start the docker stack:
```
make build-all && make sda-s3-up
```
- Prepare data
```
for i in $(seq 2); do seq $i > /tmp/t$i.txt; done
docker cp verify:/shared /tmp/
user=test@dummy.org
export ACCESS_TOKEN=$(curl -s -k http://localhost:8080/tokens | jq -r '.[0]')
token=$ACCESS_TOKEN
sed -i '/host_/s/s3inbox:8000/localhost:18000/g' /tmp/shared/s3cfg
rm -f /tmp/t*.c4gh
```
- Upload a file encrypted with one of the keys under `c4gh.privateKeys` and then ingest it.
Note that the status of the file should be changed from `uploaded` to `verified`.
```sh
sda-cli -config /tmp/shared/s3cfg upload -encrypt-with-key /tmp/shared/c4gh.pub.pem /tmp/t1.txt
sda-admin file list -user test@dummy.org
curl -H "Authorization: Bearer $token" -H "Content-Type: application/json" -X POST -d '{"pubkey": "'"$( base64 -w0 /tmp/shared/c4gh.pub.pem)"'", "description": "c4gh pub key 0"}' http://localhost:8090/c4gh-keys/add
sda-admin file ingest -filepath test_dummy.org/t1.txt.c4gh -user test@dummy.org
```

- Upload another file encrypted with the other key under `c4gh.privateKeys` and then ingest it.
Note that the status of the file should also be changed from `uploaded` to `verified`.
 
```sh 
curl -H "Authorization: Bearer $token" -H "Content-Type: application/json" -X POST -d '{"pubkey": "'"$( base64 -w0 /tmp/shared/c4gh1.pub.pem)"'", "description": "c4gh pub key 1"}' http://localhost:8090/c4gh-keys/add  
sda-cli -config /tmp/shared/s3cfg upload -encrypt-with-key /tmp/shared/c4gh1.pub.pem /tmp/t2.txt
sda-admin file ingest -filepath test_dummy.org/t2.txt.c4gh -user test@dummy.org 
```

### DIFF
--- a/sda/cmd/verify/verify.go
+++ b/sda/cmd/verify/verify.go
@@ -206,19 +206,17 @@ func main() {
 			}
 
 			var key *[32]byte
-			var foundKey bool
-			for _, key = range archiveKeyList {
-				size, err := headers.EncryptedSegmentSize(header, *key)
+			for _, k := range archiveKeyList {
+				size, err := headers.EncryptedSegmentSize(header, *k)
 				if (err == nil) && (size != 0) {
-					foundKey = true
+					key = k
 
 					break
 				}
-				log.Warnf("Failed to get encrypted segment size with key: %v, trying next key. Reason: %s", *key, err.Error())
 			}
 
-			if !foundKey {
-				log.Errorf("Failed to get encrypted segment size with all available key(s).")
+			if key == nil {
+				log.Errorf("no matching key found for file: %s.", message.ArchivePath)
 
 				continue
 			}
@@ -226,7 +224,7 @@ func main() {
 			mr := io.MultiReader(bytes.NewReader(header), io.TeeReader(f, archiveFileHash))
 			c4ghr, err := streaming.NewCrypt4GHReader(mr, *key, nil)
 			if err != nil {
-				log.Errorf("Failed to open C4GH decryptor stream with key: %v", *key)
+				log.Errorf("failed to open c4gh decryptor stream, reson: %s", err.Error())
 
 				continue
 			}
--- a/sda/cmd/verify/verify.go
+++ b/sda/cmd/verify/verify.go
@@ -39,7 +39,7 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
-	key, err := config.GetC4GHKey()
+	archiveKeyList, err := config.GetC4GHprivateKeys()
 	if err != nil {
 		log.Fatal(err)
 	}
@@ -204,13 +204,20 @@ func main() {
 				continue
 			}
 
-			hr := bytes.NewReader(header)
-			// Feed everything read from the archive file to archiveFileHash
-			mr := io.MultiReader(hr, io.TeeReader(f, archiveFileHash))
+			var c4ghr *streaming.Crypt4GHReader
 
-			c4ghr, err := streaming.NewCrypt4GHReader(mr, *key, nil)
-			if err != nil {
-				log.Errorf("Failed to open c4gh decryptor stream, reson: (%s)", err.Error())
+			for _, key := range archiveKeyList {
+				// Feed data read from the archive file into archiveFileHash
+				mr := io.MultiReader(bytes.NewReader(header), io.TeeReader(f, archiveFileHash))
+
+				c4ghr, err = streaming.NewCrypt4GHReader(mr, *key, nil)
+				if (err == nil) && (c4ghr != nil) {
+					break
+				}
+				log.Warnf("Failed to open C4GH decryptor with key: %v, trying next key. Reason: %s", *key, err.Error())
+			}
+			if c4ghr == nil {
+				log.Errorf("Failed to open C4GH decryptor stream with all available key(s)")
 
 				continue
 			}
--- a/sda/cmd/verify/verify.go
+++ b/sda/cmd/verify/verify.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"strings"
 
+	"github.com/neicnordic/crypt4gh/model/headers"
 	"github.com/neicnordic/crypt4gh/streaming"
 	"github.com/neicnordic/sensitive-data-archive/internal/broker"
 	"github.com/neicnordic/sensitive-data-archive/internal/config"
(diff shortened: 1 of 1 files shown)


sda/cmd/verify/verify.go
```diff
@@ -204,13 +204,20 @@ func main() {
 				continue
 			}
 
-			hr := bytes.NewReader(header)
-			// Feed everything read from the archive file to archiveFileHash
-			mr := io.MultiReader(hr, io.TeeReader(f, archiveFileHash))
+			var c4ghr *streaming.Crypt4GHReader
```
jbygdell, line 207:
```suggestion
```

sda/cmd/verify/verify.go
```diff
@@ -204,13 +204,20 @@ func main() {
...
+			for _, key := range archiveKeyList {
+				// Feed data read from the archive file into archiveFileHash
+				mr := io.MultiReader(bytes.NewReader(header), io.TeeReader(f, archiveFileHash))
+
+				c4ghr, err = streaming.NewCrypt4GHReader(mr, *key, nil)
+				if (err == nil) && (c4ghr != nil) {
+					break
+				}
+				log.Warnf("Failed to open C4GH decryptor with key: %v, trying next key. Reason: %s", *key, err.Error())
+			}
```
jbygdell, line 218:
```suggestion
			var key *[32]byte
			for _, key = range archiveKeyList {
				size, err := headers.EncryptedSegmentSize(header, *key)
				if (err == nil) && (size != 0) {
					break
				}
			}

			mr := io.MultiReader(bytes.NewReader(header), io.TeeReader(f, archiveFileHash))
			c4ghr, err := streaming.NewCrypt4GHReader(mr, *key, nil)
			if err != nil {
```

I'm thinking something like this would suffice.
There is not really any point in creating the readers on every iteration in the for loop.

sda/cmd/verify/verify.go
```diff
@@ -204,13 +205,45 @@ func main() {
...
+			// 	if (err == nil) && (c4ghr != nil) {
+			// 		break
+			// 	}
+			// 	log.Warnf("Failed to open C4GH decryptor with key: %v, trying next key. Reason: %s", *key, err.Error())
+			// }
+			// if c4ghr == nil {
+			// 	log.Errorf("Failed to open C4GH decryptor stream with all available key(s)")
+
+			// 	continue
+			// }
```
jbygdell, line 224:
```suggestion
```

Don't keep code that is commented out.

sda/cmd/verify/verify.go
```diff
@@ -204,13 +205,45 @@ func main() {
...
+				log.Errorf("Failed to get encrypted segment size with all available key(s).")
+
+				continue
+			}
 
+			mr := io.MultiReader(bytes.NewReader(header), io.TeeReader(f, archiveFileHash))
 			c4ghr, err := streaming.NewCrypt4GHReader(mr, *key, nil)
 			if err != nil {
-				log.Errorf("Failed to open c4gh decryptor stream, reson: (%s)", err.Error())
+				log.Errorf("Failed to open C4GH decryptor stream with key: %v", *key)
```
jbygdell, line 246:
```suggestion
				log.Errorf("failed to open c4gh decryptor stream, reson: (%s)", err.Error())
```
Do NOT leak the crypt4gh key!!

sda/cmd/verify/verify.go
```diff
@@ -204,13 +205,45 @@ func main() {
...
+			// 	}
+			// 	log.Warnf("Failed to open C4GH decryptor with key: %v, trying next key. Reason: %s", *key, err.Error())
+			// }
+			// if c4ghr == nil {
+			// 	log.Errorf("Failed to open C4GH decryptor stream with all available key(s)")
+
+			// 	continue
+			// }
+			var key *[32]byte
+			var foundKey bool
```
jbygdell, line 226:
```suggestion
```

sda/cmd/verify/verify.go
```diff
@@ -204,13 +205,45 @@ func main() {
...
+			for _, key = range archiveKeyList {
+				size, err := headers.EncryptedSegmentSize(header, *key)
+				if (err == nil) && (size != 0) {
+					foundKey = true
```
jbygdell, line 230:
```suggestion
			for _, k := range archiveKeyList {
				size, err := headers.EncryptedSegmentSize(header, *k)
				if (err == nil) && (size != 0) {
					key = k
```

Instead of these bool variables use a temporary variable inside the loop instead.

sda/cmd/verify/verify.go
```diff
@@ -204,13 +205,45 @@ func main() {
...
+					break
+				}
+				log.Warnf("Failed to get encrypted segment size with key: %v, trying next key. Reason: %s", *key, err.Error())
+			}
+
+			if !foundKey {
+				log.Errorf("Failed to get encrypted segment size with all available key(s).")
+
+				continue
+			}
```
jbygdell, line 241:
```suggestion
			if key == nil {
				log.Errorf("no matching key found for file: %s.",message.ArchivePath)

				continue
			}
```
The log messages should reflect what we try to do, not necessary what the functions are called.

MalinAhlberg (approved):
Looks good and works well when tested!

sda/cmd/verify/verify.go
```diff
@@ -204,13 +205,45 @@ func main() {
...
+			var key *[32]byte
+			var foundKey bool
+			for _, key = range archiveKeyList {
+				size, err := headers.EncryptedSegmentSize(header, *key)
+				if (err == nil) && (size != 0) {
+					foundKey = true
+
+					break
+				}
+				log.Warnf("Failed to get encrypted segment size with key: %v, trying next key. Reason: %s", *key, err.Error())
```
jbygdell, line 234:
```suggestion
```
This is taken care of if we fail to find a key to use.

jbygdell (changes requested):

sda/cmd/verify/verify.go
```diff
@@ -204,13 +205,45 @@ func main() {
 				continue
 			}
 
-			hr := bytes.NewReader(header)
-			// Feed everything read from the archive file to archiveFileHash
-			mr := io.MultiReader(hr, io.TeeReader(f, archiveFileHash))
+			// var c4ghr *streaming.Crypt4GHReader
```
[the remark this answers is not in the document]
nanjiangshu: Good catch! removed

sda/cmd/verify/verify.go
nanjiangshu: fixed in 6a59d828

sda/cmd/verify/verify.go
nanjiangshu: fixed in 6a59d828

sda/cmd/verify/verify.go
nanjiangshu: fixed in 6a59d828

MalinAhlberg (approved):

jbygdell (approved):

PR #1252, opened 2025-01-10
Use range properly in sda-download

pontus:
sda-download accepts and parses a `Range` header if any. Unfortunately, it decides whatever it's serving an entire file before doing so, ending up in serving from the start of the object request.

This PR moves that decision until after the `Range` header has been considered. Also adds a test to check that getting from something other than the start of the object works when using `Range` and extends some tests to check additional things.


jbygdell (dismissed):

MalinAhlberg (dismissed):
:+1: 
A note, unrelated to this change: this is not tested with htsget as download currently does not serve unencrypted .bai files, which htsget requires.

jbygdell (approved):

MalinAhlberg:
After discussions with @jbygdell, here are some assumptions that we believe should be true. Note that all of them are not true for the current version, but for the future they seem to be sound.
Would you agree with these, @pontus ? If not, please elaborate :)

- the coordinates are for the decrypted file
- the start/end position be relative to the actual file or the crypt4gh header+file (note, a change in htsget would be needed here)
- the crypt4gh header should always be intact
- the start position, if inside of the file, should always adhere to block boundaries
- the end position, should always adhere to block boundaries

PR #1336, opened 2025-01-30
[Test] Add NGINX to k3d

jbygdell:
## Description
This PR adds an NGINX ingress controller to the k3d setup so that it will be possible to test ingress routing on a local setup.

## How to test

When using cURL it's possible to add the hosts header like this:

```
curl -H "Host: test" http://localhost/
```

If a browser is needed the `/etc/hosts` file needs to be modified to contain lines like this for each endpoint that are needed.

```
127.0.0.1 HOSTNAME
```

Or it is possible to use host names like this when deploying the stack `login.127.0.0.1.nip.io`, this will then resolve to localhost when used, both by cURL and any browser.

### DIFF
--- a/.github/integration/scripts/charts/dependencies.sh
+++ b/.github/integration/scripts/charts/dependencies.sh
@@ -9,7 +9,7 @@ random-string() {
 }
 
 if [ "$1" == "local" ]; then
-        if [ ! "$(command crypt4gh)" ]; then
+        if [ ! "$(command crypt4gh --version)" ]; then
                 echo "crypt4gh not installed, get it from here: https://github.com/neicnordic/crypt4gh/releases/latest"
                 exit 1
         elif [ "$(crypt4gh --version | cut -d ' ' -f1)" == "GA4GH" ]; then
@@ -18,8 +18,18 @@ if [ "$1" == "local" ]; then
                 exit 1
         fi
 
-        if [ ! "$(command yq)" ]; then
+        if [ ! "$(command yq --version)" ]; then
                 echo "yq not installed, get it from here: https://github.com/mikefarah/yq/releases/latest"
+                exit 1
+        fi
+
+        if [ ! "$(command jq --version)" ]; then
+                echo "jq not installed"
+                exit 1
+        fi
+
+        if [ ! "$(command xxd --version 2>&1)" ]; then
+                echo "xxd not installed"
                 exit 1
         fi
 else
--- a/README.md
+++ b/README.md
@@ -176,19 +176,25 @@ wget -q -O - https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | bash
 curl -s https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | bash
 ```
 
-Once installed a cluster named `test-cluster` can be created as such:
+#### Create a cluster
 
-```sh
-k3d cluster create test-cluster
-```
-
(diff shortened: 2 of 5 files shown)


.github/integration/scripts/charts/dependencies.sh
```diff
@@ -22,6 +22,16 @@ if [ "$1" == "local" ]; then
                 echo "yq not installed, get it from here: https://github.com/mikefarah/yq/releases/latest"
                 exit 1
         fi
+
+        if [ ! "$(command jq)" ]; then
```
nanjiangshu, line 26: `command jq`  tries to run the commands instead of checking if they exist. This will fail if the command requires arguments, so as for this case.

It's better to change all usage of `command aaa` to improve the robustness.

nanjiangshu, line 26:
```suggestion
        if ! command -v jq >/dev/null 2>&1; then
```

.github/integration/scripts/charts/dependencies.sh
```diff
@@ -22,6 +22,16 @@ if [ "$1" == "local" ]; then
...
+                echo "jq not installed"
+                exit 1
+        fi
+
+        if [ ! "$(command xxd)" ]; then
```
nanjiangshu, line 31:
```suggestion
        if ! command -v xxd >/dev/null 2>&1; then
```

README.md
```diff
@@ -176,6 +176,8 @@ wget -q -O - https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | bash
 curl -s https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | bash
 ```
 
+#### Create a cluster
```
nanjiangshu, line 179: If the cluster is created by 
```
k3d cluster create test-cluster
```
The Nginx ingress controller will not be deployed. So maybe better to clarify that how one can create a cluster with a different name but still want to deploy the Nginx ingress controller. 

Or just remove it so that a new cluster should always be created by `make k3d-create-cluster`

README.md
```diff
@@ -190,6 +192,16 @@ The new cluster's connection details will automatically be merged into your defa
 kubectl get nodes
 ```
 
+The Nginx ingress controller is deployed and will bind to port 80 and 443 of the host system. As such a deployed service with an ingress definition can then be targeted by setting the `Host: HOSTNAME` header fo that service.
```
nanjiangshu, line 195:
```suggestion
The Nginx Ingress controller is deployed and will bind to ports 80 and 443 of the host system. As such, a deployed service with an Ingress definition can be targeted by setting the `Host: HOSTNAME` header for that service.
```

README.md
```diff
@@ -190,6 +192,16 @@ The new cluster's connection details will automatically be merged into your defa
...
+
+```sh
+curl -H "Host: test" http://localhost/
+```
+
+For testing ingress endpoints with other applications like a web browser, the hosts file `/etc/hosts` will ned to be edited unless some other form of DNS modifivcaion is used.
```
nanjiangshu, line 201:
```suggestion
For testing Ingress endpoints with other applications like a web browser, the hosts file `/etc/hosts` will need to be edited unless some other form of DNS modification is used.
```

nanjiangshu (approved):
Looks good!

aaperis (approved):
looks good

pahatz:
okay, merged! never mind then : )

PR #1357, opened 2025-02-04
bugfix/persist fileID over changes of the service instance

aaperis:
## Related issue(s) and PR(s)
This PR closes a bug first reported by our endusers: #1358.

## Description
It provides a fix for the case when for example a client loses connection to the server while uploading a file and when connection is re-established the upload continues through a different s3inbox instance.  In such a case, the registered `fileID` for the file in question would not have been persisted on the new instance, but now it does.

## How to test
I found it somewhat difficult to come up with an automated test for this without a heavy refactoring of the code into small batches of separate functions (that would otherwise not serve any purpose) . Any ideas welcome!

I tested it locally by temporarily adding `p.fileIds[r.URL.Path] = ""` in line 202, building the image and bringing the sda stack up using the `make` commands, uploading a file to the inbox and seeing that the logs show no errors. Without the fix of the PR the previous procedure will reproduce the error seen on the cluster. Make sure to upload a file from scratch because otherwise the state of the database will render this test obsolete (as it happened when testing my initial fix..).

This should be fixed now. So PR open again.

Note: the failing test is not related to this PR per se.

### DIFF
--- a/sda/cmd/s3inbox/proxy.go
+++ b/sda/cmd/s3inbox/proxy.go
@@ -177,6 +177,7 @@ func (p *Proxy) allowedResponse(w http.ResponseWriter, r *http.Request, token jw
 	}
 
 	// Send message to upstream and set file as uploaded in the database
+	// nolint: nestif // We need a nested if statement for checking whether fileId is persisted during possible reconnections
 	if p.uploadFinishedSuccessfully(r, s3response) {
 		log.Debug("create message")
 		message, err := p.CreateMessageFromRequest(r, token)
--- a/sda/cmd/s3inbox/proxy.go
+++ b/sda/cmd/s3inbox/proxy.go
@@ -203,7 +203,6 @@ func (p *Proxy) allowedResponse(w http.ResponseWriter, r *http.Request, token jw
 		// The following block is for treating the case when the client loses connection to the server and then it reconnects to a
 		// different instance of s3inbox. For more details see #1358.
 		if p.fileIds[r.URL.Path] == "" {
-
 			p.fileIds[r.URL.Path], err = p.database.GetFileIDByUserPathAndStatus(username, filepath, "registered")
 			if err != nil {
 				p.internalServerError(w, r, fmt.Sprintf("failed to retrieve fileID from database: %v", err))
--- a/sda/cmd/s3inbox/proxy.go
+++ b/sda/cmd/s3inbox/proxy.go
@@ -204,14 +204,7 @@ func (p *Proxy) allowedResponse(w http.ResponseWriter, r *http.Request, token jw
 		// different instance of s3inbox. For more details see #1358.
 		if p.fileIds[r.URL.Path] == "" {
 
-			corrID, err := p.database.GetCorrID(username, filepath, "")
-			if err != nil {
-				p.internalServerError(w, r, fmt.Sprintf("failed to retrieve corrID from database: %v", err))
-
-				return
-			}
-
-			p.fileIds[r.URL.Path], err = p.database.GetFileID(corrID)
+			p.fileIds[r.URL.Path], err = p.database.GetFileIDByUserPathAndStatus(username, filepath, "registered")
 			if err != nil {
 				p.internalServerError(w, r, fmt.Sprintf("failed to retrieve fileID from database: %v", err))
 
--- a/sda/cmd/s3inbox/proxy.go
+++ b/sda/cmd/s3inbox/proxy.go
@@ -201,7 +201,7 @@ func (p *Proxy) allowedResponse(w http.ResponseWriter, r *http.Request, token jw
 		}
 
 		// The following block is for treating the case when the client loses connection to the server and then it reconnects to a
-		// different instance of s3inbox. For more details see https://github.com/NBISweden/BigPicture-Deployment/issues/283.
+		// different instance of s3inbox. For more details see #1358.
 		if p.fileIds[r.URL.Path] == "" {
 
 			corrID, err := p.database.GetCorrID(username, filepath, "")
--- a/sda/cmd/s3inbox/proxy.go
+++ b/sda/cmd/s3inbox/proxy.go
@@ -199,6 +199,27 @@ func (p *Proxy) allowedResponse(w http.ResponseWriter, r *http.Request, token jw
 			return
 		}
 
+		// The following block is for treating the case when the client loses connection to the server and then it reconnects to a
+		// different instance of s3inbox. For more details see https://github.com/NBISweden/BigPicture-Deployment/issues/283.
+		if p.fileIds[r.URL.Path] == "" {
+
+			corrID, err := p.database.GetCorrID(username, filepath, "")
+			if err != nil {
+				p.internalServerError(w, r, fmt.Sprintf("failed to retrieve corrID from database: %v", err))
+
+				return
+			}
+
+			p.fileIds[r.URL.Path], err = p.database.GetFileID(corrID)
+			if err != nil {
+				p.internalServerError(w, r, fmt.Sprintf("failed to retrieve fileID from database: %v", err))
+
+				return
+			}
+
+			log.Debugf("resuming work on file with fileId: %v", p.fileIds[r.URL.Path])
+		}
+
 		log.Debugf("marking file %v as 'uploaded' in database", p.fileIds[r.URL.Path])
 		err = p.database.UpdateFileEventLog(p.fileIds[r.URL.Path], "uploaded", p.fileIds[r.URL.Path], "inbox", "{}", string(jsonMessage))
 		if err != nil {
--- a/sda/internal/database/db_functions.go
+++ b/sda/internal/database/db_functions.go
@@ -110,7 +110,7 @@ func (dbs *SDAdb) GetFileIDByUserPathAndStatus(submissionUser, filePath, status 
 	// 2, 4, 8, 16, 32 seconds between each retry event.
 	for count := 1; count <= RetryTimes; count++ {
 		fileID, err = dbs.getFileIDByUserPathAndStatus(submissionUser, filePath, status)
-		if err == nil {
+		if err == nil || strings.Contains(err.Error(), "sql: no rows in result set") {
 			break
 		}
 		time.Sleep(time.Duration(math.Pow(2, float64(count))) * time.Second)
--- a/sda/internal/database/db_functions.go
+++ b/sda/internal/database/db_functions.go
@@ -100,6 +100,45 @@ func (dbs *SDAdb) getInboxFilePathFromID(submissionUser, fileID string) (string,
 	return filePath, nil
 }
 
+// GetFileIDByUserPathAndStatus checks if a file exists in the database for a given user and submission filepath
+// and returns its fileID for the latest specified status
+func (dbs *SDAdb) GetFileIDByUserPathAndStatus(submissionUser, filePath, status string) (string, error) {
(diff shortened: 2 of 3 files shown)


sda/cmd/s3inbox/proxy.go
```diff
@@ -199,6 +200,27 @@ func (p *Proxy) allowedResponse(w http.ResponseWriter, r *http.Request, token jw
...
+
+			p.fileIds[r.URL.Path], err = p.database.GetFileID(corrID)
+			if err != nil {
+				p.internalServerError(w, r, fmt.Sprintf("failed to retrieve fileID from database: %v", err))
+
+				return
+			}
+
+			log.Debugf("resuming work on file with fileId: %v", p.fileIds[r.URL.Path])
+		}
```
jbygdell, line 222:
```suggestion
		corrID, err := p.database.GetCorrID(username, filepath, "")
		if err != nil {
			p.internalServerError(w, r, fmt.Sprintf("failed to retrieve corrID from database: %v", err))

			return
		}

		p.fileIds[r.URL.Path], err = p.database.GetFileID(corrID)
		if err != nil {
			p.internalServerError(w, r, fmt.Sprintf("failed to retrieve fileID from database: %v", err))

			return
		}
		log.Debugf("resuming work on file with fileId: %v", p.fileIds[r.URL.Path])
```

Technically we only need this since `fileIds` is a `map[string]sting` and as such can not hold duplicate entries.

One thing we need to take into account is memory leak, unless the `p.fileIds` is cleaned it will continue to increase in size until the app is restarted.

sda/cmd/s3inbox/proxy.go
```diff
@@ -199,6 +200,27 @@ func (p *Proxy) allowedResponse(w http.ResponseWriter, r *http.Request, token jw
 			return
 		}
 
+		// The following block is for treating the case when the client loses connection to the server and then it reconnects to a
+		// different instance of s3inbox. For more details see https://github.com/NBISweden/BigPicture-Deployment/issues/283.
```
jbygdell, line 204:
```suggestion
		// different instance of s3inbox. For more details see https://github.com/NBISweden/BigPicture-Deployment/issues/283.
```
Don't link to private repositories, create/transfer the issue to this repo instead.

sda/cmd/s3inbox/proxy.go
```diff
@@ -177,6 +177,7 @@ func (p *Proxy) allowedResponse(w http.ResponseWriter, r *http.Request, token jw
 	}
 
 	// Send message to upstream and set file as uploaded in the database
+	// nolint: nestif // We need a nested if statement for checking whether fileId is persisted during possible reconnections
```
jbygdell, line 180:
```suggestion
```

Should not be needed based on the comments below.

jbygdell (changes requested):

sda/cmd/s3inbox/proxy.go
```diff
@@ -199,6 +200,27 @@ func (p *Proxy) allowedResponse(w http.ResponseWriter, r *http.Request, token jw
...
+			p.fileIds[r.URL.Path], err = p.database.GetFileID(corrID)
+			if err != nil {
+				p.internalServerError(w, r, fmt.Sprintf("failed to retrieve fileID from database: %v", err))
+
+				return
+			}
+
+			log.Debugf("resuming work on file with fileId: %v", p.fileIds[r.URL.Path])
+		}
+
```
jbygdell, line 223:
```suggestion
		corrID, err := p.database.GetCorrID(username, filepath, "")
		if err != nil {
			p.internalServerError(w, r, fmt.Sprintf("failed to retrieve corrID from database: %v", err))

			return
		}

		p.fileIds[r.URL.Path], err = p.database.GetFileID(corrID)
		if err != nil {
			p.internalServerError(w, r, fmt.Sprintf("failed to retrieve fileID from database: %v", err))

			return
		}
		log.Debugf("resuming work on file with fileId: %v", p.fileIds[r.URL.Path])
```

Technically we only need this since `fileIds` is a `map[string]sting` and as such can not hold duplicate entries.

One thing we need to take into account is memory leak, unless the `p.fileIds` is cleaned it will continue to increase in size until the app is restarted.

aaperis: The reason why I added the if clause is to avoid a couple of extra calls to the database when they are not needed. So I would suggest we keep a bit of more complexity in favor of performance, but otherwise I agree that it is not strictly needed.

Good point about the memory leak. Worst case, won't the pod become OOM and restarted by k8s?

jbygdell: >Good point about the memory leak. Worst case, won't the pod become OOM and restarted by k8s?

Yes, which might break the upload for someone.

jbygdell: >The reason why I added the if clause is to avoid a couple of extra calls to the database when they are not needed. So I would suggest we keep a bit of more complexity in favor of performance, but otherwise I agree that it is not strictly needed.

These two calls only happen when a file uploaded is completed, it's not going to decrease the performance.

aaperis: > > Good point about the memory leak. Worst case, won't the pod become OOM and restarted by k8s?
> 
> Yes, which might break the upload for someone.

indeed, but with this fix the upload should continue with the other pod (if we have deployed one). We should fixing eitherway, I am just theoritizing here :-)

aaperis: > > The reason why I added the if clause is to avoid a couple of extra calls to the database when they are not needed. So I would suggest we keep a bit of more complexity in favor of performance, but otherwise I agree that it is not strictly needed.
> 
> These two calls only happen when a file uploaded is completed, it's not going to decrease the performance.

I still don't see why add two more calls to the database when they are not needed instead of a cheap check for an empty field.

aaperis: > One thing we need to take into account is memory leak, unless the `p.fileIds` is cleaned it will continue to increase in size until the app is restarted. Which will happen if the start and end action is on different pods.

I just noticed that a cleanup is done a bit below, [here](https://github.com/neicnordic/sensitive-data-archive/blob/591e5d0c7656410cf617a94589cf0919b959eb15/sda/cmd/s3inbox/proxy.go#L210). This looks enough?

jbygdell: That is still inside the `completed uploads` block

sda/internal/database/db_functions_test.go
```diff
@@ -1228,3 +1228,36 @@ func (suite *DatabaseTests) TestGetInboxFilePathFromID() {
...
+	filePath := fmt.Sprintf("/%v/Deletefile1.c4gh", user)
+	fileID, err := db.RegisterFile(filePath, user)
+	if err != nil {
+		suite.FailNow("Failed to register file")
+	}
+
+	// check happy path
+	fileID2, err := db.getFileIDByUserPathAndStatus(user, filePath, "registered")
+	assert.NoError(suite.T(), err)
+	assert.Equal(suite.T(), fileID, fileID2)
```
jbygdell, line 1246:
```suggestion
	// sanity check - shold fail
	fileID2, err := db.getFileIDByUserPathAndStatus("wrong-user", filePath, "registered")
	assert.EqualError(suite.T(), err, "sql: no rows in result set")

	// check happy path
	fileID2, err := db.getFileIDByUserPathAndStatus(user, filePath, "registered")
	assert.NoError(suite.T(), err)
	assert.Equal(suite.T(), fileID, fileID2)
```

jbygdell (changes requested):

sda/internal/database/db_functions.go
```diff
@@ -100,6 +100,45 @@ func (dbs *SDAdb) getInboxFilePathFromID(submissionUser, fileID string) (string,
...
+// and returns its fileID for the latest specified status
+func (dbs *SDAdb) GetFileIDByUserPathAndStatus(submissionUser, filePath, status string) (string, error) {
+	var (
+		err    error
+		fileID string
+	)
+	// 2, 4, 8, 16, 32 seconds between each retry event.
+	for count := 1; count <= RetryTimes; count++ {
+		fileID, err = dbs.getFileIDByUserPathAndStatus(submissionUser, filePath, status)
+		if err == nil {
```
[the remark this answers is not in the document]
aaperis: good point!

sda/internal/database/db_functions.go
```diff
@@ -100,6 +100,45 @@ func (dbs *SDAdb) getInboxFilePathFromID(submissionUser, fileID string) (string,
...
+func (dbs *SDAdb) GetFileIDByUserPathAndStatus(submissionUser, filePath, status string) (string, error) {
+	var (
+		err    error
+		fileID string
+	)
+	// 2, 4, 8, 16, 32 seconds between each retry event.
+	for count := 1; count <= RetryTimes; count++ {
+		fileID, err = dbs.getFileIDByUserPathAndStatus(submissionUser, filePath, status)
+		if err == nil || strings.Contains(err.Error(), "sql: no rows in result set") {
+			break
```
nanjiangshu, line 114: When `err == nil`, the actual error caused by `sql: no rows in result set` will not be returned. 
Maybe change to 
```go
return "", sql.ErrNoRows
```

nanjiangshu: Or Maybe 

```go
	if errors.Is(err, sql.ErrNoRows) {
		return "", sql.ErrNoRows
	}
```

will work. Not tested yet.

nanjiangshu (commented):
Works well when I tested according to the description. Just left a minor comment about how the proper error should be returned.

sda/internal/database/db_functions.go
aaperis: The error is captured and returned correctly, see [here](https://github.com/neicnordic/sensitive-data-archive/blob/8d0998f20b55035cd166543a6707a77288cf54fc/sda/internal/database/db_functions_test.go#L1243) and the line below for the corresponding test.

nanjiangshu (approved):

sda/cmd/s3inbox/proxy.go
```diff
@@ -199,6 +200,19 @@ func (p *Proxy) allowedResponse(w http.ResponseWriter, r *http.Request, token jw
 			return
 		}
 
+		// The following block is for treating the case when the client loses connection to the server and then it reconnects to a
+		// different instance of s3inbox. For more details see #1358.
+		if p.fileIds[r.URL.Path] == "" {
+			p.fileIds[r.URL.Path], err = p.database.GetFileIDByUserPathAndStatus(username, filepath, "registered")
+			if err != nil {
+				p.internalServerError(w, r, fmt.Sprintf("failed to retrieve fileID from database: %v", err))
```
jbygdell, line 208: Was thinking on how smart it is to return all base errors to the client but it is done in so many places that we need to do it all at once.

jbygdell (approved):

sda/cmd/s3inbox/proxy.go
aaperis: Indeed, we shouldn't propagate the error msg to the client. It should be enough to report the errorCode.  This can change centrally by deprecating [this](https://github.com/neicnordic/sensitive-data-archive/blob/8d0998f20b55035cd166543a6707a77288cf54fc/sda/cmd/s3inbox/proxy.go#L538). I see now that the `log(err)` is done in two places as well.

But better leave this for a new issue so that the team can discuss on it. I opened #1380 about it.

PR #1362, opened 2025-02-06
[auth] Serve both raw (for download) and resigned (for upload) tokens

### LINKED ISSUE #1337: [auth] Serve both raw (for download) and resigned (for upload) tokens
As a user 
I wand a clear GUI where I can fetch credentials for uploading/downloading files to/from the archive
so that I do not use the wrong token for a task.

Description:

The resigned token from SDA-auth doesnot work with SDA-download since the latter needs the ls-aai raw token.To mitigate this, sda-auth should serve both tokens (when deployed in resignJwt=true mode) and for compatibility with sda-cli, two s3config files, one for uploading and one for downloading data.

Also, while at it, do some stylistic changes to the webpage like removing the visas from display.

A/C:
- auth result page has two buttons: 
  - "Download credentials for uploading to Inbox"
  This returns the s3cmd file (name it `s3cmd-inbox.conf`) that contains the resigned token 
  - "Download credentials for downloading from the archive"
  This returns the s3cmd file (name it `s3cmd-download.conf`) that contains the ls-aai access token
- display both tokens on the page (instead of only the access token)
- if auth is deployed with `resignJwt=false` then serve the credentials with the access token in both places
- tests verifying that this works
- Optional: while at it, remove the visas from display

nanjiangshu:
## Related issue(s) and PR(s)
This PR closes #1337 

## Description
This PR allows `auth` to serve both raw and resigned tokens as well as their corresponding s3 config files.
This PR is based on the branch `tests/mock-aai` in which the service `mock-aai` is implemented. Nevertheless, the changes of this PR and the changes in the branch `tests/mock-aai` are in different folders of code so there's no conflict when merging to the main branch.

## How to test
1. Start the SDA services with resignJWT enabled by 
setting `AUTH_RESIGNJWT=true` in the file `.github/integration/sda-s3-integration.yml`
and `server.jwtpubkeypath : "/shared/keys/pub/"` in the config file `.github/integration/sda/config.yaml`.

then run 
```sh
make sda-s3-up
```

2. When the services are up, visit http://localhost:8801 and login with the `Test User`, check that there are two different tokens and also click the download buttons for two different S3 config files, that is, `s3cmd-inbox.conf` and `s3cmd-download.conf`.

3. Modify bucket base and use_https for local testing.
For Mac:
```sh
sed -i '' '/host_/s/inbox:8000/localhost:18000/g' s3cmd-inbox.conf
sed -i '' '/host_/s/inbox:8000/localhost:18000/g' s3cmd-download.conf
sed -i '' 's/use_https = True/use_https = False/g' s3cmd-inbox.conf
sed -i '' 's/use_https = True/use_https = False/g' s3cmd-download.conf
```
For Linux:
```sh
sed -i  '/host_/s/inbox:8000/localhost:18000/g' s3cmd-inbox.conf 
sed -i  '/host_/s/inbox:8000/localhost:18000/g' s3cmd-download.conf
sed -i  's/use_https = True/use_https = False/g' s3cmd-inbox.conf
sed -i  's/use_https = True/use_https = False/g' s3cmd-download.conf
```

4. create test files
```sh
for i in $(seq 2); do seq $i > /tmp/t$i.txt; done
```
**Test uploading**
5. Test uploading with the config file `s3cmd-inbox.conf` and it should work.
```sh
sda-cli -config s3cmd-inbox.conf upload -encrypt-with-key mykey.pub.pem /tmp/t1.txt
```

6. Test uploading with the config file `s3cmd-download.conf` and it should fail.
```sh
sda-cli -config s3cmd-download.conf upload -encrypt-with-key mykey.pub.pem /tmp/t3.txt  

Consider renewing the token.
Error: listing uploaded files: failed to list objects, reason: Unauthorized: not authorized
	status code: 401, request id: , host id: 
```

**Test downloading**
7. Start the download service by running the following command in the folder `sda-download`
```sh
 CONFIGFILE=dev_utils/config-notls_local.yaml go run cmd/main.go 
 ```
8. Test accessing the download endpoint with the resigned token and it should fail with `get visas failed`
```sh
token=$(grep token s3cmd-inbox.conf | awk '{print $3}')
curl  -H "Authorization: Bearer $token" http://localhost:18080/metadata/datasets    
```

9. Test accessing the download endpoint with the raw token and it should work (return an empty list) 
```sh
token=$(grep token s3cmd-download.conf | awk '{print $3}')
curl  -H "Authorization: Bearer $token" http://localhost:18080/metadata/datasets  
```

### DIFF
--- a/sda/cmd/auth/frontend/templates/oidc.html
+++ b/sda/cmd/auth/frontend/templates/oidc.html
@@ -32,18 +32,18 @@
             </p>
             {{if .ResignedToken}}
             <p class="text-center">
-              Your access token {{if .ExpDateResigned}}(expires at {{.ExpDateResigned}} UTC){{end}} for uploading to the Inbox is:
+              For <strong>uploading</strong> to the Inbox, your access token {{if .ExpDateResigned}}(expires at {{.ExpDateResigned}} UTC){{end}} is:
             </p>
             <pre class="border border-secondary rounded py-2 px-3 my-4" id="logintext">{{.ResignedToken}}</pre>
             {{end}}
+            <a href="/oidc/s3conf-inbox" class="btn btn-primary btn-block mb-5">Download credentials to upload to the Inbox</a>
             {{if .RawToken}}
             <p class="text-center">
-              Your access token {{if .ExpDateRaw}}(expires at {{.ExpDateRaw}} UTC){{end}} for downloading from the Archive is:
+              For <strong>downloading</strong> from the Archive, your access token {{if .ExpDateRaw}}(expires at {{.ExpDateRaw}} UTC){{end}} is:
             </p>
             <pre class="border border-secondary rounded py-2 px-3 my-4" id="logintext">{{.RawToken}}</pre>
             {{end}}
-            <a href="/oidc/s3conf-inbox" class="btn btn-primary btn-block">Download credentials for uploading to the Inbox</a>
-            <a href="/oidc/s3conf-download" class="btn btn-primary btn-block">Download credentials for downloading from the Archive</a>
+            <a href="/oidc/s3conf-download" class="btn btn-primary btn-block mb-5">Download credentials to access the Archive</a>
             <a href="/" class="btn btn-primary btn-block">Continue</a>
       </div>
     </div>
--- a/sda/cmd/auth/frontend/templates/oidc.html
+++ b/sda/cmd/auth/frontend/templates/oidc.html
@@ -28,7 +28,7 @@
     <div class="jumbotron" role="region">
       <div class="container" id="loginbox1">
             <p class="lead text-center">
-              Welcome, {{.User}}!
+              Welcome, {{.Profile}}!
             </p>
             {{if .ResignedToken}}
             <p class="text-center">
--- a/sda/cmd/auth/frontend/templates/oidc.html
+++ b/sda/cmd/auth/frontend/templates/oidc.html
@@ -38,19 +38,20 @@
             <pre class="border border-secondary rounded py-2 px-3 my-4" id="logintext">{{.}}</pre>
             {{ end }}
             {{ end }}
-            {{if .Token}}
+            {{if .ResignedToken}}
             <p class="text-center">
-              Your access token is:
+              Your access token {{if .ExpDateResigned}}(expires at {{.ExpDateResigned}} UTC){{end}} for uploading to the Inbox is:
             </p>
-            <pre class="border border-secondary rounded py-2 px-3 my-4" id="logintext">{{.Token}}</pre>
+            <pre class="border border-secondary rounded py-2 px-3 my-4" id="logintext">{{.ResignedToken}}</pre>
             {{end}}
-            {{if .ExpDate}}
+            {{if .RawToken}}
             <p class="text-center">
-              Your access token expires (UTC):
+              Your access token {{if .ExpDateRaw}}(expires at {{.ExpDateRaw}} UTC){{end}} for downloading from the Archive is:
             </p>
-            <pre class="text-center" id="logintext">{{.ExpDate}}</pre>
+            <pre class="border border-secondary rounded py-2 px-3 my-4" id="logintext">{{.RawToken}}</pre>
             {{end}}
-            <a href="/oidc/s3conf" class="btn btn-primary btn-block">Download inbox s3cmd credentials</a>
+            <a href="/oidc/s3conf-inbox" class="btn btn-primary btn-block">Download credentials for uploading to the Inbox</a>
+            <a href="/oidc/s3conf-download" class="btn btn-primary btn-block">Download credentials for downloading from the Archive</a>
             <a href="/" class="btn btn-primary btn-block">Continue</a>
       </div>
     </div>
--- a/sda/cmd/auth/frontend/templates/oidc.html
+++ b/sda/cmd/auth/frontend/templates/oidc.html
@@ -30,13 +30,15 @@
             <p class="lead text-center">
               Welcome, {{.Fullname}}!
             </p>
-            {{if .ResignedToken}}
-            <p class="text-center">
-              For <strong>uploading</strong> to the Inbox, your access token {{if .ExpDateResigned}}(expires at {{.ExpDateResigned}} UTC){{end}} is:
-            </p>
-            <pre class="border border-secondary rounded py-2 px-3 my-4" id="logintext">{{.ResignedToken}}</pre>
+            {{if (not .cegaID)}}
+              {{if .ResignedToken}}
+              <p class="text-center">
+                For <strong>uploading</strong> to the Inbox, your access token {{if .ExpDateResigned}}(expires at {{.ExpDateResigned}} UTC){{end}} is:
+              </p>
+              <pre class="border border-secondary rounded py-2 px-3 my-4" id="logintext">{{.ResignedToken}}</pre>
+              {{end}}
+              <a href="/oidc/s3conf-inbox" class="btn btn-primary btn-block mb-5">Download credentials to upload to the Inbox</a>
             {{end}}
-            <a href="/oidc/s3conf-inbox" class="btn btn-primary btn-block mb-5">Download credentials to upload to the Inbox</a>
             {{if .RawToken}}
             <p class="text-center">
               For <strong>downloading</strong> from the Archive, your access token {{if .ExpDateRaw}}(expires at {{.ExpDateRaw}} UTC){{end}} is:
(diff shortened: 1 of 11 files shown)


.github/integration/sda/aai-mock/userinfos/dummy-user.yaml
```diff
@@ -2,7 +2,7 @@ web-display-name: "Dummy User - submitter"
 
 sub: "test@dummy.org"
 
-name: "Dummy User"
+name: "test@dummy.org"
```
jbygdell, line 5:
```suggestion
name: "Dummy User"
```

This should be as it is.

.github/integration/sda/aai-mock/userinfos/test-user.yaml
```diff
@@ -2,7 +2,7 @@ web-display-name: "Test User - admin"
 
 sub: "testu@lifescience-ri.eu"
 
-name: "Test User"
+name: "testu@lifescience-ri.eu"
```
jbygdell, line 5:
```suggestion
name: "Test User"
```
This should be as it is.

.github/integration/sda/config.yaml
```diff
@@ -86,8 +86,8 @@ oidc:
 server:
   cert: ""
   key: ""
-  # jwtpubkeypath: "/shared/keys/pub/"
-  jwtpubkeyurl: "http://localhost:8800/oidc/jwk"
+  jwtpubkeypath: "/shared/keys/pub/"
+  # jwtpubkeyurl: "http://localhost:8800/oidc/jwk"
```
jbygdell, line 90:
```suggestion
  # jwtpubkeypath: "/shared/keys/pub/"
  jwtpubkeyurl: "http://localhost:8800/oidc/jwk"
```

Leave this as it is or the automated test will have issues.

sda/cmd/auth/frontend/templates/oidc.html
```diff
@@ -30,27 +30,20 @@
             <p class="lead text-center">
               Welcome, {{.User}}!
             </p>
-            {{if .Passport}}
+            {{if .ResignedToken}}
```
jbygdell, line 33: Putting the comment here as the line above that needs fixing can't be commented on.

the `Welcome, {{.User}}!` hsould be `Welcome, {{.Profile}}!` in order to display  the users full name instead of the username.

sda/cmd/auth/main.go
```diff
@@ -270,6 +275,7 @@ func (auth AuthHandler) elixirLogin(ctx iris.Context) *OIDCData {
 	}
 
 	if auth.Config.ResignJwt {
+		log.Debugf("Resigning token for user %s", idStruct.User)
```
jbygdell, line 278: Putting the comment here instead.

the line `jwt.SubjectKey:    idStruct.Profile,` below should be `jwt.SubjectKey:    idStruct.User,`

jbygdell (changes requested):
1. Don't make changes to files in the base branch unless that one is `main`. Your Base branch is a branch that belongs to another PR.
2. The root cause of all your mishaps is due to the `OIDCIdentity` struct having rather bad names for what data it holds. It should actually look like this:
```
type OIDCIdentity struct {
	Subject              string
	Passport             []string
	RawToken             string
	ResignedToken        string
	Fullname             string
	Email                string
	EdupersonEntitlement []string
	ExpDateRaw           string
	ExpDateResigned      string
}
```

.github/integration/sda/config.yaml
nanjiangshu: I noticed it. Have reverted this commit.

sda/cmd/auth/main.go
nanjiangshu: fixed in ba8e75f8

.github/integration/sda/aai-mock/userinfos/test-user.yaml
nanjiangshu: reverted the commit ea0908da

sda/cmd/auth/frontend/templates/oidc.html
nanjiangshu: Fixed in 476555bf

.github/integration/sda/aai-mock/userinfos/dummy-user.yaml
nanjiangshu: reverted the commit ea0908da

jbygdell (commented):
remove the following commits:
ea0908daffdb47b0ce75a6fc51ba18e3fe35a976 : Fix name of the test user for aai mock
d69100a7224f7ec94828c8a66c2a1d73668565cf : Use jwtpubkeypath instead of jwtpubkeyurl
c516ed6a0303f69ddcf8e51e5b1c37a8b4d0bc8e : Revert "Use jwtpubkeypath instead of jwtpubkeyurl"
b78c70ca1a4ec99c01012862920db8bffd60c1b1 : Revert "Fix name of the test user for aai mock"

Rebase this branch on main, should only be your 10 commits by then.

sda/cmd/auth/frontend/templates/oidc.html
```diff
@@ -28,29 +28,22 @@
...
-            {{ end }}
-            {{if .Token}}
-            <p class="text-center">
-              Your access token is:
-            </p>
-            <pre class="border border-secondary rounded py-2 px-3 my-4" id="logintext">{{.Token}}</pre>
+            <pre class="border border-secondary rounded py-2 px-3 my-4" id="logintext">{{.ResignedToken}}</pre>
             {{end}}
-            {{if .ExpDate}}
+            <a href="/oidc/s3conf-inbox" class="btn btn-primary btn-block mb-5">Download credentials to upload to the Inbox</a>
```
jbygdell, line 39: This should not be visible if the CEGA login is enabled.

sda/cmd/auth/main.go
```diff
@@ -264,30 +269,32 @@ func (auth AuthHandler) elixirLogin(ctx iris.Context) *OIDCData {
...
-		idStruct.Token = token
-		idStruct.ExpDate = expDate
+		idStruct.ResignedToken = token
+		idStruct.ExpDateResigned = expDate
 	}
 
 	log.WithFields(log.Fields{"authType": "oidc", "user": idStruct.User}).Infof("User was authenticated")
-	s3conf := getS3ConfigMap(idStruct.Token, auth.Config.S3Inbox, idStruct.User)
+	s3confInbox := getS3ConfigMap(idStruct.ResignedToken, auth.Config.S3Inbox, idStruct.User)
+	s3confDownload := getS3ConfigMap(idStruct.RawToken, auth.Config.S3Inbox, idStruct.User)
```
aaperis, line 295: Here it would be nice for download's config file to not contain the `host_base` and `host_bucket` of the inbox. 
In the future the download module of sda-cli could use this info instead for download's base url.

This could be a new issue that would be about what we want to include/exclude in download's s3cmd.conf.

nanjiangshu: Good point! I think we may need some discussion to decide the exact information needed for it. I created the issue https://github.com/neicnordic/sensitive-data-archive/issues/1395 to address this problem.

sda/cmd/auth/frontend/templates/oidc.html
nanjiangshu: If I understand correctly, after CEGA login, the page will be [ega.html](https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/cmd/auth/frontend/templates/ega.html), not oidc.html. 

Also when testing the EGA login, I found a bug at https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/cmd/auth/main.go#L208 which makes `loginform.html` failed to render. There's also redundant code in this function which needs to be refactored. I guess it is better to be addressed in a new issue?

jbygdell: In a F-EGA setup the ega-login is used for uploading data.
But in order to download data the user will have to authenticate using the LS-AAI login. And then the user should not see the 'upload credentials' section of the LS-login response page.

sda/cmd/auth/main.go
jbygdell: > Here it would be nice for download's config file to not contain the `host_base` and `host_bucket` of the inbox. 
> In the future the download module of sda-cli could use this info instead for download's base url.
> 
> This could be a new issue that would be about what we want to include/exclude in download's s3cmd.conf. 

If the download service is 'S3' compatible it should be usable with standard S3 tools and then all relevant information should exist in the config file.

Sda-cli might not be picky about what's in the config file but s3cmd is.

jbygdell (approved):

MalinAhlberg (approved):
Good work, and thanks for the very helpful descriptions on how to test :+1: !
All works well for me (minus some initial confusion around that there are two variables called `AUTH_RESIGNJWT` in `.github/integration/sda-s3-integration.yml`).

aaperis (approved):

PR #1394, opened 2025-02-15
Add Download API specification

### LINKED ISSUE #1378: Simple Download API specification
Make a proposal for download (encrypted) API. Should include:
* Download files
* List Files
* Download partial files

A/C:
* Show in sprint review.

jbygdell:
## Related issue(s) and PR(s)
This PR closes #1378 .

## Description
The `/file/{fileId}:` is optional since the same functionality exists in the `/file` endpoint. Although it provides a more convenient way to get a file out.

### Endpoints

`/file` - endpoint that takes query arguments to return a file by ID or path, the dataset name is a required parameter. 
`/file/{file_stable_id}` - endpoint that returns the file withe the corresponding stable_id

`/info/datasets` - returns a list of datasets the user has access to
`/info/datasets/{dataset_name}` - returns information about the requested dataset
`/info/datasets/{dataset_name}/files` - returns information about the files in the dataset

## How to test
Paste the contents of the yaml file into: https://editor-next.swagger.io/ for a nicer view

### DIFF
--- a/sda/cmd/download/swagger_v1.yml
+++ b/sda/cmd/download/swagger_v1.yml
@@ -13,12 +13,6 @@ paths:
           required: true
           schema:
             type: string
-        - in: header
-          description: "The JWT used to authenticate the user"
-          name: token
-          required: true
-          schema:
-            type: string
         - in: query
           name: dataset
           schema:
@@ -90,12 +84,6 @@ paths:
           required: true
           schema:
             type: string
-        - in: header
-          description: "The JWT used to authenticate the user"
-          name: token
-          required: true
-          schema:
-            type: string
         - in: path
           name: fileId
           schema:
@@ -107,7 +95,8 @@ paths:
           required: false
           schema:
             type: string
-          description: Byte range of the file that is to be delivered
+          description: 0-indexed byte range of the file that is to be delivered.
+          example: "Range: bytes=20-200, 300-400"
       responses:
         "200":
           content:
@@ -230,7 +219,7 @@ paths:
         - bearerAuth: []
   /ready:
     get:
-      description: Returns the status of the application, used by the kubernets API to decide if traffic should be routed to the app
+      description: Returns the status of the application, used by the Kubernets API to decide if traffic should be routed to the app
       responses:
         "200":
           description: The service is operational
--- a/sda/cmd/download/swagger_v1.yml
+++ b/sda/cmd/download/swagger_v1.yml
@@ -59,14 +59,21 @@ paths:
                 type: string
                 format: binary
           description: successful operation
+        "206":
+          content:
+            application/octet-stream:
+              schema:
+                type: string
+                format: binary
+          description: successful partial file delivery
         "400":
-          description: Invalid file reference
+          description: Bad request i.e start larger than end
         "401":
           description: Authentication failure
         "403":
           description: The user does not have access to the dataset
         "416":
-          description: Range Not Satisfiable
+          description: Unsatisfiable range i.e. end larger than file size
         "500":
           description: Internal application error
       security:
@@ -93,20 +100,11 @@ paths:
             type: string
           required: true
           description: Stable ID of the file to download
-        - in: query
-          name: end
+        - in: header
+          name: Range
           required: false
           schema:
-            default: 0
-            minimum: 0
-            type: integer
-        - in: query
-          name: start
-          required: false
-          schema:
-            default: 0
-            minimum: 0
-            type: integer
+            type: string
       responses:
         "200":
           content:
@@ -115,12 +113,21 @@ paths:
                 type: string
                 format: binary
           description: successful operation
+        "206":
+          content:
+            application/octet-stream:
+              schema:
+                type: string
+                format: binary
+          description: successful partial file delivery
         "400":
-          description: Invalid file reference
+          description: Bad request i.e start larger than end
         "401":
           description: Authentication failure
         "403":
           description: The user does not have access to the dataset
+        "416":
+          description: Unsatisfiable range i.e. end larger than file size
         "500":
           description: Internal application error
       security:
@@ -138,6 +145,8 @@ paths:
                 type: array
         "401":
           description: Authentication failure
+        "500":
+          description: Internal application error
       security:
         - bearerAuth: []
   /info/datasets/{datasetId}:
--- a/sda/cmd/download/swagger_v1.yml
+++ b/sda/cmd/download/swagger_v1.yml
@@ -8,7 +8,7 @@ paths:
       description: Returns the requested file to the user
       parameters:
         - in: header
-          description: "Public key used to re-encrypt the file header with"
+          description: "Public key used to re-encrypt the file (header) with. This should be supplied as the base64 encoding of the PEM (RFC 7468) encoding of the public key"
           name: public_key
           required: true
           schema:
@@ -30,18 +30,20 @@ paths:
           schema:
             type: string
           required: false
-          description: Stable ID of the file to download
+          description: Stable ID of the file to download. It is an error to supply both filePath and fileId
         - in: query
           name: filePath
           schema:
             type: string
           required: false
-          description: File path of the file to download
+          description: File path of the file to download. It is an error to supply both filePath and fileId
         - in: query
           name: end
+          description: 'End of requested data in the form of the 0-index of the first octet not included in the
+                                 response. That is, data delivered will have octet indices in the closed-open set
+                                 [start,end). Defaults to end of file if not set'
           required: false
           schema:
-            default: 0
             minimum: 0
             type: integer
         - in: query
--- a/sda/cmd/download/swagger_v1.yml
+++ b/sda/cmd/download/swagger_v1.yml
@@ -60,14 +60,14 @@ paths:
               schema:
                 type: string
                 format: binary
-          description: successful operation
+          description: Successful operation
         "206":
           content:
             application/octet-stream:
               schema:
                 type: string
                 format: binary
-          description: successful partial file delivery
+          description: Successful partial file delivery
         "400":
           description: Bad request i.e start larger than end
         "401":
@@ -107,6 +107,7 @@ paths:
           required: false
           schema:
             type: string
+          description: Byte range of the file that is to be delivered
       responses:
         "200":
           content:
@@ -114,14 +115,14 @@ paths:
               schema:
                 type: string
                 format: binary
-          description: successful operation
+          description: Successful operation
         "206":
           content:
             application/octet-stream:
               schema:
                 type: string
                 format: binary
-          description: successful partial file delivery
+          description: Successful partial file delivery
         "400":
           description: Bad request i.e start larger than end
         "401":
@@ -139,7 +140,7 @@ paths:
       description: Returns a list of the datasets the user has access to, an empty list is a valid response.
       responses:
         "200":
-          description: successful operation
+          description: Successful operation
           content:
             application/json:
               example: ["aa-dataset-123456-asdfgh", "bb-dataset-123456-asdfgh"]
@@ -179,7 +180,7 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/DatasetInfo"
-          description: successful operation
+          description: Successful operation
         "401":
           description: Authentication failure
         "403":
@@ -218,7 +219,7 @@ paths:
                 type: array
                 items:
                   $ref: "#/components/schemas/FileInfo"
-          description: successful operation
+          description: Successful operation
         "401":
           description: Authentication failure
         "403":
--- a/sda/cmd/download/swagger_v1.yml
+++ b/sda/cmd/download/swagger_v1.yml
@@ -20,6 +20,12 @@ paths:
           schema:
             type: string
         - in: query
+          name: dataset
+          schema:
+            type: string
+          required: true
+          description: Stable ID of the dataset
+        - in: query
           name: fileId
           schema:
             type: string
--- a/sda/cmd/download/swagger_v1.yml
+++ b/sda/cmd/download/swagger_v1.yml
@@ -1,0 +1,258 @@
+openapi: 3.0.3
+info:
+  title: SDA file download API
+  version: "1.0"
+paths:
+  /file:
+    get:
+      description: Returns the requested file to the user
+      parameters:
+        - in: header
+          description: "Public key used to re-encrypt the file header with"
+          name: public_key
+          required: true
+          schema:
+            type: string
+        - in: header
+          description: "The JWT used to authenticate the user"
+          name: token
+          required: true
+          schema:
+            type: string
+        - in: query
+          name: fileId
+          schema:
+            type: string
+          required: false
+          description: Stable ID of the file to download
+        - in: query
+          name: filePath
+          schema:
+            type: string
+          required: false
+          description: File path of the file to download
+        - in: query
+          name: end
+          required: false
+          schema:
+            default: 0
+            minimum: 0
+            type: integer
+        - in: query
+          name: start
+          required: false
+          schema:
+            default: 0
+            minimum: 0
+            type: integer
+      responses:
+        "200":
+          content:
+            application/octet-stream:
+              schema:
+                type: string
+                format: binary
+          description: successful operation
+        "400":
+          description: Invalid file reference
+        "401":
+          description: Authentication failure
+        "403":
+          description: The user does not have access to the dataset
+        "416":
+          description: Range Not Satisfiable
+        "500":
+          description: Internal application error
+      security:
+        - bearerAuth: []
+  /file/{fileId}:
+    get:
+      description: Returns the requested file to the user
+      parameters:
+        - in: header
+          description: "Public key used to re-encrypt the file header with"
+          name: public_key
+          required: true
+          schema:
+            type: string
+        - in: header
+          description: "The JWT used to authenticate the user"
+          name: token
+          required: true
+          schema:
+            type: string
+        - in: path
+          name: fileId
+          schema:
+            type: string
+          required: true
+          description: Stable ID of the file to download
+        - in: query
+          name: end
+          required: false
+          schema:
+            default: 0
+            minimum: 0
+            type: integer
+        - in: query
+          name: start
+          required: false
+          schema:
+            default: 0
+            minimum: 0
+            type: integer
+      responses:
+        "200":
+          content:
+            application/octet-stream:
+              schema:
+                type: string
+                format: binary
+          description: successful operation
+        "400":
+          description: Invalid file reference
+        "401":
+          description: Authentication failure
+        "403":
+          description: The user does not have access to the dataset
+        "500":
+          description: Internal application error
+      security:
+        - bearerAuth: []
+  /info/datasets:
+    get:
+      description: Returns a list of the datasets the user has access to, an empty list is a valid response.
+      responses:
+        "200":
+          description: successful operation
+          content:
+            application/json:
+              example: ["aa-dataset-123456-asdfgh", "bb-dataset-123456-asdfgh"]
+              schema:
+                type: array
+        "401":
+          description: Authentication failure
+      security:
+        - bearerAuth: []
+  /info/datasets/{datasetId}:
+    get:
+      description: Returns an array with metadata about the dataset
+      parameters:
+        - in: header
+          description: "Public key used to re-encrypt the file header with"
+          name: public_key
+          required: true
+          schema:
+            type: string
+        - in: header
+          description: "The JWT used to authenticate the user"
+          name: token
+          required: true
+          schema:
+            type: string
+        - in: path
+          name: datasetId
+          schema:
+            type: string
+          required: true
+          description: Stable ID of the dataset of interest
+      responses:
+        "200":
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/DatasetInfo"
+          description: successful operation
+        "401":
+          description: Authentication failure
+        "403":
+          description: The user does not have access to the dataset
+        "500":
+          description: Internal application error
+      security:
+        - bearerAuth: []
+  /info/datasets/{datasetId}/files:
+    get:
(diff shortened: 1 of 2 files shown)


sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,273 @@
...
+info:
+  title: SDA file download API
+  version: "1.0"
+paths:
+  /file:
+    get:
+      description: Returns the requested file to the user
+      parameters:
+        - in: header
+          description: "Public key used to re-encrypt the file header with"
```
pontus, line 11:
```suggestion
          description: "Public key used to re-encrypt the file (header) with. This should be supplied as the base64 encoding of the PEM (RFC 7468) encoding of the public key"
```

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,273 @@
...
+          schema:
+            type: string
+          required: true
+          description: Stable ID of the dataset
+        - in: query
+          name: fileId
+          schema:
+            type: string
+          required: false
+          description: Stable ID of the file to download
```
pontus, line 33:
```suggestion
          description: Stable ID of the file to download. It is an error to supply both filePath and fileId
```

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,273 @@
...
+        - in: query
+          name: filePath
+          schema:
+            type: string
+          required: false
+          description: File path of the file to download
```
pontus, line 39:
```suggestion
          description: File path of the file to download. It is an error to supply both filePath and fileId
```

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,273 @@
...
+        - in: query
+          name: start
+          required: false
+          schema:
+            default: 0
+            minimum: 0
+            type: integer
+      responses:
+        "200":
+          content:
```
pontus, line 56:
```suggestion
         headers:
           'Content-Disposition: attachment; filename="filePath"'
          content:
```

?

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,273 @@
...
+          name: filePath
+          schema:
+            type: string
+          required: false
+          description: File path of the file to download
+        - in: query
+          name: end
+          required: false
+          schema:
+            default: 0
```
pontus, line 44:
```suggestion
```

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,273 @@
...
+          required: false
+          description: Stable ID of the file to download
+        - in: query
+          name: filePath
+          schema:
+            type: string
+          required: false
+          description: File path of the file to download
+        - in: query
+          name: end
```
pontus, line 41:
```suggestion
          name: end
          description: 'End of requested data in the form of the 0-index of the first octet not included in the
                                 response. That is, data delivered will have octet indices in the closed-open set
                                 [start,end). Defaults to end of file if not set'
```

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,273 @@
...
+          description: Authentication failure
+        "403":
+          description: The user does not have access to the dataset
+        "416":
+          description: Unsatisfiable range i.e. end larger than file size
+        "500":
+          description: Internal application error
+      security:
+        - bearerAuth: []
+  /file/{fileId}:
```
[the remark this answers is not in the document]
jbygdell: File stable ID is a unique entity and since the authorization step checks whether or not the user has permission to  access the file or not this should be ok.

sda/cmd/download/swagger_v1.yml
jbygdell: Not really, if you request a file by stable ID that should be the filename in the header.

sda/cmd/download/swagger_v1.yml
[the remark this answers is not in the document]
jbygdell: Adding the `datasetID` would make the authorization step easier/faster but then again is it really needed?
One problem that might arrive here is if the dataset ID is an URL.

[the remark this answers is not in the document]
pontus: I think we agree that it's not strictly needed, but the flow gets easier (should not be overly complex but compared to other things in download it's a lot). We should maybe mull on it for a while. 

(Values for `datasetID` can be managed through encoding.)

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,275 @@
...
+          description: successful operation
+        "401":
+          description: Authentication failure
+        "403":
+          description: The user does not have access to the dataset
+        "500":
+          description: Internal application error
+      security:
+        - bearerAuth: []
+  /ready:
```
MalinAhlberg, line 230: Not `/health`? https://github.com/neicnordic/sensitive-data-archive/blob/5218110d6ccc75d8682d42229b0b5d03ef3801b5/sda-download/api/api.go#L74

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,275 @@
...
+          description: Stable ID of the dataset of interest
+      responses:
+        "200":
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/FileInfo"
+          description: successful operation
```
MalinAhlberg, line 221:
```suggestion
          description: Successful operation
```

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,275 @@
...
+            type: string
+          required: true
+          description: Stable ID of the dataset of interest
+      responses:
+        "200":
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/DatasetInfo"
+          description: successful operation
```
MalinAhlberg, line 182:
```suggestion
          description: Successful operation
```

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,275 @@
...
+        "500":
+          description: Internal application error
+      security:
+        - bearerAuth: []
+  /info/datasets:
+    get:
+      description: Returns a list of the datasets the user has access to, an empty list is a valid response.
+      responses:
+        "200":
+          description: successful operation
```
MalinAhlberg, line 142:
```suggestion
          description: Successful operation
```

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,275 @@
...
+                type: string
+                format: binary
+          description: successful operation
+        "206":
+          content:
+            application/octet-stream:
+              schema:
+                type: string
+                format: binary
+          description: successful partial file delivery
```
MalinAhlberg, line 124:
```suggestion
          description: Successful partial file delivery
```

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,275 @@
...
```
MalinAhlberg, line 70:
```suggestion
          description: Successful partial file delivery
```
Uppercasing the messages for 200 (here and below) for consistency.

sda/cmd/download/swagger_v1.yml
jbygdell: `health` or `ready` serve the same purpose

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,275 @@
...
+          description: Authentication failure
+        "403":
+          description: The user does not have access to the dataset
+        "416":
+          description: Unsatisfiable range i.e. end larger than file size
+        "500":
+          description: Internal application error
+      security:
+        - bearerAuth: []
+  /info/datasets:
```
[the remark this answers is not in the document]
jbygdell: Yes, a token is required since we need the user-info token in order to get the visas so we know which datasets the user has access to.

sda/cmd/download/swagger_v1.yml
MalinAhlberg: I know, I was just wondering if it would make sense to keep the same endpoint name :upside_down_face:

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,276 @@
...
+          description: Authentication failure
+        "403":
+          description: The user does not have access to the dataset
+        "500":
+          description: Internal application error
+      security:
+        - bearerAuth: []
+  /ready:
+    get:
+      description: Returns the status of the application, used by the kubernets API to decide if traffic should be routed to the app
```
pahatz, line 233:
```suggestion
      description: Returns the status of the application, used by the kubernetes API to decide if traffic should be routed to the app
```

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,276 @@
...
+          type: string
+          format: date-time
+          example: "2025-02-14T14:51:26.639Z"
+        files:
+          type: integer
+          format: int32
+          example: 1234
+        name:
+          type: string
+          example: aa-dataset-123456-asdfgh
```
pahatz, line 254:
```suggestion
          example: aa-Dataset-123456-asdfgh
```

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,276 @@
...
+        filePath:
+          type: string
+          example: "samples/controls/sample1.cram.c4gh"
+        size:
+          type: integer
+          format: int64
+          example: 56623104
+        stableId:
+          type: string
+          example: aa-file-123456-asdfgh
```
pahatz, line 271:
```suggestion
          example: aa-File-123456-asdfgh
```

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,276 @@
...
+        - bearerAuth: []
+  /info/datasets:
+    get:
+      description: Returns a list of the datasets the user has access to, an empty list is a valid response.
+      responses:
+        "200":
+          description: Successful operation
+          content:
+            application/json:
+              example: ["aa-dataset-123456-asdfgh", "bb-dataset-123456-asdfgh"]
```
pahatz, line 146:
```suggestion
              example: ["aa-Dataset-123456-asdfgh", "bb-Dataset-123456-asdfgh"]
```

pahatz (dismissed):
Looks good!

MalinAhlberg:
In general, I think this looks very good!
 I'm a bit confused about the `start`/`end` parameters vs the `Range` header, though. Are they to be used in exactly the same way? Ie is this
>"End of requested data in the form of the 0-index of the first octet not included in the response. That is, data delivered will have octet indices in the closed-open set [start,end). Defaults to end of file if not set"

true for the end of the range to? And exactly how should the range be formatted? `"Range: bytes=20-200"`?

jbygdell:
> In general, I think this looks very good! I'm a bit confused about the `start`/`end` parameters vs the `Range` header, though. Are they to be used in exactly the same way? Ie is this
> 
> > "End of requested data in the form of the 0-index of the first octet not included in the response. That is, data delivered will have octet indices in the closed-open set [start,end). Defaults to end of file if not set"
> 
> true for the end of the range to? And exactly how should the range be formatted? `"Range: bytes=20-200"`?

Range headers has to adhere to the specification like what you wrote above. But also this is correct `"Range: bytes=20-200, 30-400"`

The query parameters on the other hand you can supply one or both, hence the logic about inferring the missing one.

sda/cmd/download/swagger_v1.yml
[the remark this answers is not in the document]
jbygdell: That is actually handled by this:
```
      security:
        - bearerAuth: []
```

So the real issue is that the `token` header should be removed from the `file/*` endpoints

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,276 @@
...
+          schema:
+            type: string
+          required: true
+          description: Stable ID of the file to download
+        - in: header
+          name: Range
+          required: false
+          schema:
+            type: string
+          description: Byte range of the file that is to be delivered
```
MalinAhlberg, line 110:
```suggestion
          description: 0-indexed byte range of the file that is to be delivered. The end position is the first octet not included in the response.
          example: "Range: bytes=20-200, 30-400"
```

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,265 @@
+openapi: 3.0.3
+info:
+  title: SDA file download API
+  version: "1.0"
```
nanjiangshu, line 4: Since this is premature, would it be better to use version `0.x`?

MalinAhlberg (approved):
Nice!

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,265 @@
...
+paths:
+  /file:
+    get:
+      description: Returns the requested file to the user
+      parameters:
+        - in: header
+          description: "Public key used to re-encrypt the file (header) with. This should be supplied as the base64 encoding of the PEM (RFC 7468) encoding of the public key"
```
nanjiangshu, line 11: The description `Public key used to re-encrypt the file (header) with.` is ambiguous, unclear whether the public key is used to encrypt just the file header or the entire file.

 ```suggestion
          description: "Public key used to re-encrypt the file header server-side before transmission. This should be supplied as the base64 encoding of the PEM (RFC 7468) encoding of the public key"
```

File: sda/cmd/download/swagger_v1.yml [hunk 22]
```diff
@@ -0,0 +1,265 @@
...
+          description: Successful operation
+        "206":
+          content:
+            application/octet-stream:
+              schema:
+                type: string
+                format: binary
+          description: Successful partial file delivery
+        "400":
+          description: Bad request i.e start larger than end
```
nanjiangshu, line 116: Is `start lager than end` the only condition for an `400` error? If not, better to use `e.g.` instead of `i.e.`.

If yes, change `i.e` -> `i.e.`

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,265 @@
...
+        "401":
+          description: Authentication failure
+        "403":
+          description: The user does not have access to the dataset
+        "416":
+          description: Unsatisfiable range i.e. end larger than file size
```
nanjiangshu, line 122: Similar in this case, use `e.g.` if `end larger than file size` is not the only condition for `416`.

sda/cmd/download/swagger_v1.yml
nanjiangshu: If there's no special reason, I agree with @MalinAhlberg that it is better to keep the same endpoint name.

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,265 @@
...
+components:
+  schemas:
+    DatasetInfo:
+      type: object
+      properties:
+        date:
+          type: string
+          format: date-time
+          example: "2025-02-14T14:51:26.639Z"
+        files:
```
nanjiangshu, line 237: `files` represents number of files, right? Then maybe to use a clearer name, such as `fileCount`.

nanjiangshu (dismissed):

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,265 @@
...
+          schema:
+            type: string
+          required: true
+          description: Stable ID of the file to download
+        - in: header
+          name: Range
+          required: false
+          schema:
+            type: string
+          description: 0-indexed byte range of the file that is to be delivered.
```
pontus, line 98:
```suggestion
          description: 0-indexed byte range of the file that is to be delivered as per RFC7233.
```

sda/cmd/download/swagger_v1.yml
pontus: Fair, would this be useful enough that we should put the basename there or shall we skip?

jbygdell: Let's skip this one. It's not likely that someone will try to download a file this way using a browser anyways.

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,265 @@
...
+  /info/datasets/{datasetId}:
+    get:
+      description: Returns an array with metadata about the dataset
+      parameters:
+        - in: header
+          description: "Public key used to re-encrypt the file header with"
+          name: public_key
+          required: true
+          schema:
+            type: string
```
pontus, line 153:
```suggestion
```

Not applicable.

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,265 @@
...
+        - in: header
+          description: "The JWT used to authenticate the user"
+          name: token
+          required: true
+          schema:
+            type: string
```
pontus, line 159:
```suggestion
```

Handled by the `security` listing `bearerAuth`.

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,265 @@
...
+                type: array
+        "401":
+          description: Authentication failure
+        "500":
+          description: Internal application error
+      security:
+        - bearerAuth: []
+  /info/datasets/{datasetId}:
+    get:
+      description: Returns an array with metadata about the dataset
```
pontus, line 146: Would it possible to link to a schema for this?

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,265 @@
...
+  /info/datasets/{datasetId}/files:
+    get:
+      description: Returns an array with metadata about all files in a dataset
+      parameters:
+        - in: header
+          description: "Public key used to re-encrypt the file header with"
+          name: public_key
+          required: true
+          schema:
+            type: string
```
pontus, line 190:
```suggestion
```

Not applicable.

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,265 @@
...
+        - in: header
+          description: "The JWT used to authenticate the user"
+          name: token
+          required: true
+          schema:
+            type: string
```
pontus, line 196:
```suggestion
```

Managed with `security`.

sda/cmd/download/swagger_v1.yml
pontus: I can't read, apologies.

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,241 @@
+openapi: 3.0.3
+info:
+  title: SDA file download API
+  version: "1.0"
```
pontus, line 4:
```suggestion
  version: "1.0"
  description: |
     The SDA file download API enables querying of dataset content metadata and content retrieval.
     Note: unless otherwise specified, whenever coordinates are used in this specification, they reference the
     coordinate space of the re-encrypted header followed by the archive content.  
```

?

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,241 @@
...
+        "500":
+          description: Internal application error
+      security:
+        - bearerAuth: []
+  /file/{fileId}:
+    get:
+      description: Returns the requested file to the user
+      parameters:
+        - in: header
+          description: "Public key used to re-encrypt the file header with"
```
pontus, line 82:
```suggestion
          description:  "Public key used to re-encrypt the file (header) with. This should be supplied as the base64 encoding of the PEM (RFC 7468) encoding of the public key"
```

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,241 @@
...
+        filePath:
+          type: string
+          example: "samples/controls/sample1.cram.c4gh"
+        size:
+          type: integer
+          format: int64
+          example: 56623104
+        stableId:
+          type: string
+          example: aa-file-123456-asdfgh
```
pontus, line 236:
```suggestion
        stableId:
          type: string
          example: aa-file-123456-asdfgh
        lastModified:
          type: string
          format: date-time
          example: "2025-02-21T10:51:26.639Z"
```
If we mean to serve s3 from on top of this, we'll want at least to be able to construct something to put offer up as last modified timestamp as well as etag. Supposedly the `stableId` could serve for etag purposes, but a timestamp would be good I think.

jbygdell: Any S3 implementation will have it's own response specification.

pontus: Yes, but earlier we discussed having this API as a stepping stone that other services could build on (e.g s3, or htsget or whatever speciality service there is a big enough need for).

MalinAhlberg (approved):

sda/cmd/download/swagger_v1.yml
```diff
@@ -0,0 +1,256 @@
...
+          type: integer
+          format: int32
+          example: 1234
+        name:
+          type: string
+          example: aa-dataset-123456-asdfgh
+        size:
+          type: integer
+          format: int64
+          example: 6597069766656
```
pontus, line 238: What is this, the combined sum of all dataset contents (with headers?). I'd be happy to have a description.

pontus:
I'd still prefer to see https://github.com/neicnordic/sensitive-data-archive/pull/1394/files#r1967221350 addressed, but am fine with merging as is. Similarly, I think it would be helpful with a description for the size of a dataset (https://github.com/neicnordic/sensitive-data-archive/pull/1394#discussion_r1969662615) but am fine with that being closed/moved to issue.

sda/cmd/download/swagger_v1.yml
jbygdell: The only relevant size would be the total size of all encrypted files.

pontus: With the corresponding header sizes (pre-rerencyption) added?

PR #1419, opened 2025-02-19
Bugfix in the sda-svc helm chart

### LINKED ISSUE #1069: [charts] fix url concatenation so we don't get `//` in the urls
**Describe the bug**
Url paths are doen by simply joining the strings which in some cases generates double slashes.
OIDC hostnames needs to end with `/` or the matching will fail.
**Steps to reproduce**

**Expected behavior**
- [ ] charts can be deployed with no issues

**Additional context**

**Estimation of size**: small

**Estimation of priority**: low

jbygdell:
Closes #1069

## Description
This PR fixes some issues with the sda-svc chart
- SCHEMA_TYPE is added to the list of ENVs for the API service so that it will work properly in any deployment scenario.
- OIDC_CONFIGURATION_URL in the download service strips the `/` as the last character in the OIDC URL. This is to prevent a string concatenation error.

## How to test
make k3d-create-cluster
make k3d-import-images
make k3d-deploy-dependencies
make k3d-deploy-postgres
make 3d-deploy-rabbitmq
make k3d-deploy-sda-s3

### DIFF
--- a/charts/sda-svc/Chart.yaml
+++ b/charts/sda-svc/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: sda-svc
-version: 0.30.5
+version: 0.30.6
 appVersion: v0.3.196
 kubeVersion: '>= 1.26.0'
 description: Components for Sensitive Data Archive (SDA) installation
--- a/charts/sda-svc/templates/api-deploy.yaml
+++ b/charts/sda-svc/templates/api-deploy.yaml
@@ -189,6 +189,8 @@ spec:
         - name: LOG_LEVEL
           value: {{ .Values.global.log.level | quote }}
     {{- end }}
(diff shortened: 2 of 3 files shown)


jbygdell:
Waiting for the chart bump to be merged

aaperis (approved):

MalinAhlberg (commented):
Seems to work well!
Except for auth, where I get redirected to `http://oidc:8080/authorize?..` which cannot be accessed. But maybe that wasn't working before either? In that case I'm ready to approve.

Another thing, not related to this PR, but that would be nice to get fixed, is the three typos of `deploy` in the README (`deplploy-postgres`, `k3d-deplpoy-sda-posix` ad `k3d-deplpoy-sda-s3`)

MalinAhlberg (approved):

PR #1460, opened 2025-03-05
fix blank page on failed ega login

### LINKED ISSUE #1416: [Auth] EGA login doesn't work when an error occurs
## Describe the bug

Authenticating with the EGA option results in a blank page.

## Steps to reproduce

make build-all
make sda-s3-up

Authenticate with wrong user/password and the response is a blank page. This should at the least show a page with an authentication error i.e. `incorrect username or password supplied`

## Expected behavior
- [ ] Properly report to the user that something went wrong
  - [ ] wrong username and/or password (401)
  - [ ] badly configured auth service i.e responses of 404, 503, 502 and so on from the backend

## Estimation of size
small/medium/big

## Estimation of priority
low/medium/high

MalinAhlberg:
## Related issue(s) and PR(s)
This PR closes #1416 .

## Description
This PR
- fixes the bug that caused the blank page when logging in with bad credentials from the user. _Provided credentials are not valid_ is shown
- fixes the bug that caused the blank page when auth service is badly configured (`auth_cega_id` or `auth_cega_secret` is wrong). _Problems connecting to EGA authenticaton server_ is shown
- fixes the bug that lead to _This page isn't worknig_ when `cega-nss` is down. _EGA authentication server could not be contacted_ is shown.
- fixes some unrelated typos in the README

## How to test
- [ ] `make sda-s3-up`
- [ ] navigate to `localhost:8888`, choose the EGA login, and try to log in with a bad username/password
- [ ] verify that the page is not blank, but contains an error message and a new login form
- [ ] Stop `cega-nss` and verify that the page does not break when trying to log in
- [ ] Change `AUTH_CEGA_SECRET` in `.github/integration/sda-s3-integration.yml` and verify that the page does not turn blank or break when trying to log in.
- [ ] Please test anything else you might come up with

**Additional notes**
No tests are added, manual testing of the webpage will still be the standard.

### DIFF
--- a/sda/cmd/auth/main.go
+++ b/sda/cmd/auth/main.go
@@ -118,13 +118,16 @@ func (auth AuthHandler) postEGA(ctx iris.Context) {
 
 	res, err := authenticateWithCEGA(auth.Config.Cega, username)
 
+	var statusCode int
 	if err != nil {
 		log.Error(err)
+		statusCode = 404
+	} else {
+		statusCode = res.StatusCode
+		defer res.Body.Close()
 	}
 
-	defer res.Body.Close()
-
-	switch res.StatusCode {
+	switch statusCode {
 	case 200:
 		if err != nil {
 			log.Error(err)
--- a/sda/cmd/auth/main.go
+++ b/sda/cmd/auth/main.go
@@ -208,27 +208,15 @@ func (auth AuthHandler) postEGA(ctx iris.Context) {
 func (auth AuthHandler) getEGALogin(ctx iris.Context) {
 
 	s := sessions.Get(ctx)
-	message := s.GetFlashString("message")
-	if message == "" {
-		ctx.ViewData("infoUrl", auth.Config.InfoURL)
-		ctx.ViewData("infoText", auth.Config.InfoText)
-		err := ctx.View("loginform.html")
-		if err != nil {
-			log.Error("Failed to return to login form: ", err)
-
-			return
-		}
-
-		return
-	}
 	ctx.ViewData("infoUrl", auth.Config.InfoURL)
 	ctx.ViewData("infoText", auth.Config.InfoText)
-	ctx.ViewData("Reason", message)
+	message := s.GetFlashString("message")
+	if message != "" {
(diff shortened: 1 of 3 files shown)


aaperis (approved):
Works great! :smile:

MalinAhlberg:
> Works great! 😄

Thanks for the quick review! I realized there were some more problems though, they are now fixed in bd52543  and 45654b4.

sda/cmd/auth/main.go
```diff
@@ -118,13 +118,16 @@ func (auth AuthHandler) postEGA(ctx iris.Context) {
 
 	res, err := authenticateWithCEGA(auth.Config.Cega, username)
 
+	var statusCode int
 	if err != nil {
 		log.Error(err)
+		statusCode = 404
+	} else {
+		statusCode = res.StatusCode
+		defer res.Body.Close()
```
jbygdell, line 127:
```suggestion
	if err != nil {
		log.Errorf("cega connection error: %v",err)
		res = &http.Response{
			Body: io.NopCloser(nil),
			StatusCode: http.StatusInternalServerError,
		}
```

This also means that the 404 case below needs to be replaced with:
`case 500,502,503:` so that we cover all bases when the NSS server misbehaves.

sda/cmd/auth/main.go
```diff
@@ -118,13 +118,16 @@ func (auth AuthHandler) postEGA(ctx iris.Context) {
...
 	}
 
-	defer res.Body.Close()
-
-	switch res.StatusCode {
+	switch statusCode {
```
jbygdell, line 130:
```suggestion
	defer res.Body.Close()

	switch res.StatusCode {
```

jbygdell (changes requested):

MalinAhlberg:
I did the refactoring suggested by @jbygdell in a7999d3 and added some extra checks b456ef0. The last commit is another small refactoring, everything else is just rebasing.

jbygdell (approved):

nanjiangshu (approved):
Looks good. Good job!

PR #1483, opened 2025-03-17
Ingest waits for delays from s3

### LINKED ISSUE #1430: Ingest should wait for s3 backend when reading file
## Describe the bug

When ingest writes the file to archive and then tries to read it again it fails

This happens when failing fast when trying to read the file, because the metadata to s3 has not been written yet.
Might be able to fix this by removing this:

https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/internal/storage/storage.go#L308-L310

## Steps to reproduce

This happens when there is a large amount of files is in the queue

## Expected behavior

- [ ] We don't get the retries in the logs

## Estimation of size

small

## Estimation of priority

medium

MalinAhlberg:
## Related issue(s) and PR(s)
This PR closes #1430 .

## Description
Ingest needs the `GetFileSize` to accept some delay from s3, so that it does not immediately fail when receiving `NoSuchKey` or `NotFound`. At other points, however, this extra waiting time is not wanted, for example when calling the function from [`verify`](https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/cmd/verify/verify.go#L160), since that would give 2 minutes extra waiting time before the call fails (if the file is not present in the storage). Notably, [this test](https://github.com/neicnordic/sensitive-data-archive/blob/main/.github/integration/tests/sda/22_error_test.sh#L120) would require a lot of retries. 

My assumption is that it should be optional whether a delay from s3 should be expected or not.
This PR therefore adds an extra parameter to allow for this.
**If you have other ideas, please feel very welcome to post them here.**

## How to test
Make sure all tests pass. You could also mock [this call](https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/internal/storage/storage.go#L293) so that it does not find the queried file at once.

### DIFF
--- a/sda/internal/storage/storage_test.go
+++ b/sda/internal/storage/storage_test.go
@@ -256,7 +256,7 @@ func (suite *StorageTestSuite) TestPosixBackend() {
 	assert.Equal(suite.T(), writeData, readBackBuffer[:readBack], "did not read back data as expected")
 	assert.Nil(suite.T(), err, "unexpected error when reading back data")
 
-	size, err := backend.GetFileSize("testFile")
+	size, err := backend.GetFileSize("testFile", false)
 	assert.Nil(suite.T(), err, "posix NewFileReader failed when it should work")
 	assert.NotNil(suite.T(), size, "Got a nil size for posix")
 
@@ -270,7 +270,7 @@ func (suite *StorageTestSuite) TestPosixBackend() {
 	assert.NotZero(suite.T(), buf.Len(), "Expected warning missing")
 
 	buf.Reset()
-	_, err = backend.GetFileSize("posixDoesNotExist")
+	_, err = backend.GetFileSize("posixDoesNotExist", false)
 	assert.Error(suite.T(), err, "posix GetFileSize worked when it should not")
 	assert.NotZero(suite.T(), buf.Len(), "Expected warning missing")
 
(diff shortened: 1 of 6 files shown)


pahatz (approved):
Looks like it should work fine.

sda/internal/storage/storage_test.go
```diff
@@ -323,11 +323,17 @@ func (suite *StorageTestSuite) TestS3Backend() {
...
 	assert.Nil(suite.T(), err, "s3 GetFileSize failed when it should work")
 	assert.NotNil(suite.T(), size, "Got a nil size for s3")
 	assert.Equal(suite.T(), int64(len(writeData)), size, "Got an incorrect file size")
 
+	// make sure expectDelay=true works
+	size, err = s3back.GetFileSize("s3Creatable", true)
+	assert.Nil(suite.T(), err, "s3 GetFileSize with expected delay failed when it should work")
+	assert.NotNil(suite.T(), size, "Got a nil size for s3")
+	assert.Equal(suite.T(), int64(len(writeData)), size, "Got an incorrect file size")
+
```
jbygdell, line 336: Shouldn't this use a freshly uploaded file to realy see if it works as expected?

MalinAhlberg: Added in 496df2d (and then rebased on main).

jbygdell (approved):

nanjiangshu (approved):
Looks great!

PR #1506, opened 2025-03-20
[API] return stable ID as part of the file info response.

### LINKED ISSUE #1417: [admin-api] Return accession IDs via admin API

As a SDA-admin, I want to be able to fetch accession IDs (or stable IDs) via the admin API when accession IDs are mapped to files so that I can continue with dataset creation. 
## Acceptance criteria
- [ ] Some database code that we use for admin-api needs to be split. We need to separate the calls made by submitters vs. calls made by helpdesk or repository admins.
- [ ] They should be returned when we list the files for a user, if there is a stable_id return the accession id
- [ ] The endpoint should **NOT** display the accession ids for submission users.

## Additional context
Related comment https://github.com/NBISweden/BigPicture-Deployment/issues/282#issuecomment-2663320060
## Estimation of size

small/medium/big

## Estimation of priority

low/medium/high

jbygdell:
## Related issue(s) and PR(s)
This PR closes #1417.

## Description
This expands the `SubmissionFileInfo` struct to also contain the file stable IDs but only return them to the queries made to the `/users/:username/files` endpoint

### DIFF
--- a/sda/cmd/api/api.go
+++ b/sda/cmd/api/api.go
@@ -232,7 +232,7 @@ func getFiles(c *gin.Context) {
 		return
 	}
 
-	files, err := Conf.API.DB.GetUserFiles(token.Subject())
+	files, err := Conf.API.DB.GetUserFiles(token.Subject(), false)
 	if err != nil {
 		// something went wrong with querying or parsing rows
 		c.JSON(502, err.Error())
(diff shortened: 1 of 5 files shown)


MalinAhlberg (approved):
:+1:

nanjiangshu (commented):
Looks good and works well!

I just have one question, why `accessionID` is not shown for the `/files` endpoint?
I know that this is mainly for admin of the database, but I guess showing to the user will not leak any sensitive information, right?  In addition, this information might be useful when the user submit bug report, or?

nanjiangshu (approved):

PR #1507, opened 2025-03-20
[internal/database] add test step to append files to a dataset.

### LINKED ISSUE #1503: [mapper] Mapper should be able to add files to an already existing dataset
## Backround

It should be possible to extend a dataset with more files, currently mapper throws an error in case the dataset already exists.

## Description

When `mapper` gets a message for dataset creation together with a list of files it should first check if the dataset exists and in that case it should just proceed to add the files to that dataset.

## Out of scope

For this task we are assuming that we have already validated that the new files originate from the same data controller as the already existing dataset.

jbygdell:
## Related issue(s) and PR(s)
This PR closes #1503.

## Description
This adds a missing step for appending files to a dataset in the `MapFilesToDataset` DB call.

### DIFF
--- a/.github/integration/sda/rbac.json
+++ b/.github/integration/sda/rbac.json
@@ -12,8 +12,8 @@
       },
       {
          "role": "admin",
-         "path": "/dataset/verify/:dataset",
-         "action": "PUT"
+         "path": "/dataset/*",
+         "action": "(POST)|(PUT)"
       },
        {
          "role": "submission",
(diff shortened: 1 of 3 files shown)


MalinAhlberg (approved):
I can confirm that you can add more files to a dataset already.

However, it would be nice to have something like
```
      {
         "role": "submission",
         "path": "/dataset/*",
         "action": "POST"
       }
```
in `.github/integration/sda/rbac.json` so that this is easier to test manually.

MalinAhlberg (approved):

pahatz (approved):

PR #1520, opened 2025-03-26
Add Cron job to run large submission tests

### LINKED ISSUE #1504: [tests] Add tests for large datasets in terms of files
## Background

In NBIS we've encoutering problems with the pipeline when trying to ingest and create datasets when the datasets have 30k files or more. In order to investigate this problem further we need a couple of test cases. It seems like we don't need the files actually in the system to do these tests as it's about sending and/or handling lots of messages and not about doing the actual work.

## Description

- [ ] Test case where we submit 100k files to the ingest service (pre-create the database entries automatically)
- [ ] Test case where we submit 100k files to the verify service (pre-create the database entries automatically)
- [ ] Test case where we create a dataset with 100k files (pre-create the database entries automatically)

In these cases a successfull test fails because the ingest/verify services won't find the files in the filesystem, there should be no other errors.

## Out of scope

A test case with 100k actual files.

jbygdell:
The test for finalize is expected to fail for now.

## Related issue(s) and PR(s)
This PR closes #1504 .

## Description
These tests create DB entries for 100k files, sends the messages to the MQ.

The test suite is run as a cronjob on Saturdays, all 4 tests are run in parallel in order to cope with the max runtime of 6h for a job.

### Expected results

- finalize - failed test due to timeout
- ingest - passed test, 100k messages sent to the error queue
- mapper- passed test, no errors
- verify - passed test, 100k messages sent to the error queue

## How to test
Once merged there is an option to manually trigger the tests.

### DIFF
--- a/.github/integration/scripts/large_submission/finalize.sh
+++ b/.github/integration/scripts/large_submission/finalize.sh
@@ -1,0 +1,111 @@
+#!/bin/sh
+set -e
+
+for t in curl jq postgresql-client uuid-runtime; do
+    if [ ! "$(command -v $t)" ]; then
+        if [ "$(id -u)" != 0 ]; then
+            echo "$t is missing, unable to install it"
+            exit 1
+        fi
+
+        apt-get -o DPkg::Lock::Timeout=60 update >/dev/null
+        apt-get -o DPkg::Lock::Timeout=60 install -y "$t" >/dev/null
+    fi
+done
+
+submission_size=100000
+
+## empty all queues ##
+for q in accession archived backup completed inbox ingest mappings verified; do
+    curl -s -k -u guest:guest -X DELETE "http://rabbitmq:15672/api/queues/sda/$q/contents"
+done
+## truncate database
+psql -U postgres -h postgres -d sda -At -c "TRUNCATE TABLE sda.files CASCADE;"
+
+stream_size=$(curl -s -u guest:guest http://rabbitmq:15672/api/queues/sda/error_stream | jq '.messages_ready')
+
+i=1
+while [ $i -le $((submission_size)) ]; do
+    user="test@dummy.org"
+    inbox_path="inbox/test-file-$i.c4gh"
+    fileID=$(psql -U postgres -h postgres -d sda -At -c "SELECT sda.register_file('$inbox_path', '$user');")
+    if [ -z "$fileID" ]; then
+        echo "register_file failed"
+        exit 1
+    fi
+    resp=$(psql -U postgres -h postgres -d sda -At -c "UPDATE sda.files SET archive_file_path = '$fileID', archive_file_size = '$i' WHERE id = '$fileID';")
+    if [ "$resp" != "UPDATE 1" ]; then
+        echo "failed to update file $resp"
+        exit 1
+    fi
+    psql -U postgres -h postgres -d sda -At -c "INSERT INTO sda.file_event_log(file_id, event, correlation_id, user_id, message) VALUES('$fileID', 'verified', '$fileID', 'test-user', '{\"uploaded\": \"message\"}');" >/dev/null
+
+    properties=$(
+        jq -c -n \
+            --argjson delivery_mode 2 \
+            --arg correlation_id "$fileID" \
+            --arg content_encoding UTF-8 \
+            --arg content_type application/json \
+            '$ARGS.named'
+    )
+
+    DEC_SHA=$(echo $i | sha256sum | cut -d' ' -f 1)
+    decrypted_checksums=$(
+        jq -c -n \
+            --arg sha256 "$DEC_SHA" \
+            '$ARGS.named|to_entries|map(with_entries(select(.key=="key").key="type"))'
+    )
+
+    accession_id="urn:uuid:$(uuidgen)"
+    accession_payload=$(
+        jq -r -c -n \
+            --arg type accession \
+            --arg user "$user" \
+            --arg filepath "$inbox_path" \
+            --arg accession_id "$accession_id" \
+            --argjson decrypted_checksums "$decrypted_checksums" \
+            '$ARGS.named|@base64'
+    )
+
+    accession_body=$(
+        jq -c -n \
+            --arg vhost sda \
+            --arg name sda \
+            --argjson properties "$properties" \
+            --arg routing_key "accession" \
+            --arg payload_encoding base64 \
+            --arg payload "$accession_payload" \
+            '$ARGS.named'
+    )
+
+    curl -s -u guest:guest "http://rabbitmq:15672/api/exchanges/sda/sda/publish" \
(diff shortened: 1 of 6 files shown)


MalinAhlberg (commented):
In general, this looks very good to me!

Some details are not clear to me, however. Shouldn't we check what error messages we get in the error queue? To make sure that they are due to files that can't be found, and not something else?

jbygdell:
> In general, this looks very good to me!
> 
> Some details are not clear to me, however. Shouldn't we check what error messages we get in the error queue? To make sure that they are due to files that can't be found, and not something else?

Well, the only test that is of relevance is the test for `mapper`, the other ones will behave the same regardless of the number of messages/files.

pahatz (approved):
Looks good!

.github/integration/scripts/large_submission/mapper.sh
```diff
@@ -0,0 +1,95 @@
...
+submission_size=100000
+
+## empty all queues ##
+for q in accession archived backup completed inbox ingest mappings verified; do
+    curl -s -k -u guest:guest -X DELETE "http://rabbitmq:15672/api/queues/sda/$q/contents"
+done
+## truncate database
+psql -U postgres -h postgres -d sda -At -c "TRUNCATE TABLE sda.files CASCADE;"
+
+# stream_size=$(curl -s -u guest:guest http://rabbitmq:15672/api/queues/sda/error_stream | jq '.messages_ready')
```
aaperis, line 25:
```suggestion
```

.github/integration/scripts/large_submission/sda-s3.yml
```diff
@@ -0,0 +1,394 @@
...
+      - "localhost:host-gateway"
+    environment:
+      - PGPASSWORD=rootpasswd
+      - STORAGETYPE=s3
+    image: python:3.11-slim-bullseye
+    profiles:
+      - tests
+    volumes:
+      - shared:/shared
+      - ../../tests:/tests
```
aaperis, line 387:
```suggestion
```

.github/integration/scripts/large_submission/finalize.sh
```diff
@@ -0,0 +1,111 @@
...
+
+    i=$((i + 1))
+done
+
+RETRY_TIMES=0
+until [ "$(curl -s -u guest:guest http://rabbitmq:15672/api/queues/sda/accession | jq '.messages')" -eq 0 ]; do
+    echo "waiting for messages to be processed"
+    RETRY_TIMES=$((RETRY_TIMES + 1))
+    if [ "$RETRY_TIMES" -eq 30 ]; then
+        echo "::error::Time out while waiting for finalize to compete the work"
```
aaperis, line 93:
```suggestion
        echo "::error::Time out while waiting for finalize to complete the work"
```

.github/integration/scripts/large_submission/ingest.sh
```diff
@@ -0,0 +1,106 @@
...
+    i=$((i + 1))
+done
+
+sleep 10
+RETRY_TIMES=0
+until [ "$(curl -s -u guest:guest http://rabbitmq:15672/api/queues/sda/ingest | jq '.messages')" -eq 0 ]; do
+    echo "waiting for messages to be processed"
+    RETRY_TIMES=$((RETRY_TIMES + 1))
+    if [ "$RETRY_TIMES" -eq 30 ]; then
+        echo "::error::Time out while waiting for ingest to compete the work"
```
aaperis, line 90:
```suggestion
        echo "::error::Time out while waiting for ingest to complete the work"
```

.github/integration/scripts/large_submission/mapper.sh
```diff
@@ -0,0 +1,95 @@
...
+curl -s -u guest:guest "http://rabbitmq:15672/api/exchanges/sda/sda/publish" \
+    -H 'Content-Type: application/json;charset=UTF-8' \
+    -d "@/shared/message.json" >/dev/null
+
+RETRY_TIMES=0
+until [ "$(curl -s -u guest:guest http://rabbitmq:15672/api/queues/sda/mappings | jq '.messages')" -eq 0 ]; do
+    echo "waiting for dataset be registered"
+    RETRY_TIMES=$((RETRY_TIMES + 1))
+    if [ "$RETRY_TIMES" -eq 30 ]; then
+        echo "::error::Time out while waiting for mapper to compete the work"
```
aaperis, line 89:
```suggestion
        echo "::error::Time out while waiting for mapper to complete the work"
```

.github/integration/scripts/large_submission/verify.sh
```diff
@@ -0,0 +1,108 @@
...
+    i=$((i + 1))
+done
+
+sleep 10
+RETRY_TIMES=0
+until [ "$(curl -s -u guest:guest http://rabbitmq:15672/api/queues/sda/archived | jq '.messages')" -eq 0 ]; do
+    echo "waiting for messages to be processed"
+    RETRY_TIMES=$((RETRY_TIMES + 1))
+    if [ "$RETRY_TIMES" -eq 30 ]; then
+        echo "::error::Time out while waiting for verify to compete the work"
```
aaperis, line 92:
```suggestion
        echo "::error::Time out while waiting for verify to complete the work"
```

.github/integration/scripts/large_submission/finalize.sh
```diff
@@ -0,0 +1,111 @@
...
+        echo "This is currently expected"
+        exit 1
```
aaperis, line 95: why is this currently expected, wouldn't this produce a failing action?
 
probably I am missing something that has been discussed already

aaperis (commented):
Looks good but I can't seem to remember why it's ok to have the finalize test failing.

.github/integration/scripts/large_submission/finalize.sh
jbygdell: Finalize can't handle the case and keeps nacking the message.

.github/integration/scripts/large_submission/mapper.sh
```diff
@@ -0,0 +1,86 @@
...
+        --arg dataset_id EGAD74900000101 \
+        --slurpfile accession_ids /shared/accessions.txt \
+        '$ARGS.named'
+)
+echo "$mapping_payload" >/shared/payload
+
+token="$(cat /shared/token)"
+
+set +e
+resp=$(curl -s -d @/shared/payload \
```
nanjiangshu, line 64: I'm a bit confused there, since you have directed the output to `/dev/null`, do you want `resp` to capture the exit code for the curl command? If so, then it should be something like this 
```sh
responseBody=$(curl -s -d @/shared/payload \
    -H "Content-Type: application/json" \
    -H "Authorization: Bearer $token" \
    -X POST http://api:8080/dataset/create)
resp=$?
```

nanjiangshu (commented):
Looks good except for a couple of minor comments.

Issue #1037, opened 2024-09-13
ingesting an already ingested file leads never ending logs


MalinAhlberg:
**Describe the bug**
Starting ingestion of a file that is already ingested leads no never ending logs, but no error message or warning fr om`sda-admin`.

**Steps to reproduce**
Using the [setup described here](https://github.com/neicnordic/sensitive-data-archive/pull/1026) (but with new versions of the code), run
`./sda-admin file ingest --filepath requester_demo.org/data/file1.c4gh --user requester@demo.org` twice.
Run `docker compose -f .github/integration/sda-s3-integration.yml logs -f` to see the logs being filled with for example:
```
postgres  | 2024-09-12 13:36:22.248 UTC [80] ERROR:  duplicate key value violates unique constraint "unique_checksum"
...
verify    | {"level":"error","msg":"SetVerified failed, reason: (pq: duplicate key value violates unique constraint \"unique_checksum\")","time":"2024-09-12T13:36:22Z"}
```

**Expected behavior**
- [ ] ...

**Additional context**
From #1026. See also #1027.

**Estimation of size**: medium

**Estimation of priority**: medium

PR #1526, opened 2025-03-31
Refactor ingest main work loop

### LINKED ISSUE #1519: ingest loops when re-queueing messages
When trying to reingest a file through restoring a message, ingest does not check if the file is ingested so it is returned back to the queue and we end up with an stuck message in the archived.

We need some extra checks before the file event creation
Ingest should check the last status from the `file_event_log` and if the work that the service is supposed to do has happened, then the file should be moved along to the next service.

**How to reproduce:**
Send an ingest message to an already Ingested file.
We should not end up with messages stuck in a loop.

**Acceptance criteria:**

- [ ] If the message is re-queued, we should log the event/error
- [ ] Tests, if possible

**Additional context**
Created from #1463. Also see #1377, #1505 and #1082.
Also check if #1037 can be closed when this is fixed.

### LINKED ISSUE #1561: [ingest] break apart the main work loop

As a developer, I want all the code to statically testable so that It is easier to create a full test suite even for extreme edge cases.

## Acceptance criteria

- [ ] The main function send each message to a worker function that returns ack or nack with regards to the message.
- [ ] Testsuite is created
- [ ] All tests pass

## Estimation of size

small/medium/big

## Estimation of priority

low/medium/high

jbygdell:
## Related issue(s) and PR(s)
This PR closes #1561 

This PR also closes #1519 and #1037.

## Description
The code for the main pipeline components are all cluttered i n the main function. That makes it all but impossible to run any tests other than full integration tests using the final container.

This PR breaks apart the main work loop in `ingest` into parts that can be tested and adds test for the following conditions:
* TestCancelFile
* TestCancelFile_wrongCorrelationID
* TestIngestFile
* TestIngestFile_missingFile
* TestIngestFile_reingestCancelledFile
* TestIngestFile_reingestCancelledFileNewChecksum
* TestIngestFile_reingestVerifiedCancelledFile
* TestIngestFile_reingestVerifiedCancelledFileNewChecksum
* TestIngestFile_reingestVerifiedFile
* TestIngestFile_secondTime
* TestIngestFile_unknownInboxType

## How to test
While in the `sda` folder run `go test -v cmd/ingest/*.go` to only test the ingest part. In order to test all of the GO code just run `go test`.

### DIFF
--- a/sda/cmd/ingest/ingest.go
+++ b/sda/cmd/ingest/ingest.go
@@ -186,7 +186,6 @@ func main() {
 					log.Errorf("Failed to Nack message, reason: (%s)", err.Error())
 				}
 			default:
-				log.Errorln("better message needed")
 				if err := delivered.Reject(false); err != nil {
 					log.Errorf("failed to reject message for reason: (%s)", err.Error())
 				}
@@ -200,7 +199,7 @@ func main() {
 func (app *Ingest) cancelFile(correlationID string, message schema.IngestionTrigger) string {
 	fileUUID, err := app.DB.GetFileID(correlationID)
 	if err != nil {
-		log.Errorf("failed to get ID for file from message: %s, reason: %s", correlationID, err.Error())
+		log.Errorf("failed to get ID for file from message (correlationID: %s), reason: %s", correlationID, err.Error())
 		if strings.Contains(err.Error(), "sql: no rows in result set") {
 			return "reject"
 		}
--- a/sda/cmd/ingest/ingest.go
+++ b/sda/cmd/ingest/ingest.go
@@ -501,7 +501,7 @@ func (app *Ingest) ingestFile(correlationID string, message schema.IngestionTrig
 		return "ack"
 	}
 
-	if err := app.DB.SetArchived(fileInfo, fileID, correlationID); err != nil {
+	if err := app.DB.SetArchived(fileInfo, fileID); err != nil {
 		log.Errorf("SetArchived failed, reason: (%s)", err.Error())
 
 		return "nack"
--- a/sda/cmd/ingest/ingest.go
+++ b/sda/cmd/ingest/ingest.go
@@ -28,7 +28,17 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-func main() {
+type Ingest struct {
+	Archive        storage.Backend
+	ArchiveKeyList []*[32]byte
+	Conf           *config.Config
+	DB             *database.SDAdb
+	Inbox          storage.Backend
+	MQ             *broker.AMQPBroker
+}
+
+func (app *Ingest) main() {
+	var err error
 	sigc := make(chan os.Signal, 5)
 	signal.Notify(sigc, os.Interrupt, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
 	// Create a function to handle panic and exit gracefully
@@ -39,71 +49,71 @@ func main() {
 	}()
 
 	forever := make(chan bool)
-	conf, err := config.NewConfig("ingest")
+
+	app.Conf, err = config.NewConfig("ingest")
 	if err != nil {
 		log.Error(err)
 		sigc <- syscall.SIGINT
 		panic(err)
 	}
-	mq, err := broker.NewMQ(conf.Broker)
+	app.MQ, err = broker.NewMQ(app.Conf.Broker)
 	if err != nil {
 		log.Error(err)
 		sigc <- syscall.SIGINT
 		panic(err)
 	}
-	db, err := database.NewSDAdb(conf.Database)
+	app.DB, err = database.NewSDAdb(app.Conf.Database)
 	if err != nil {
 		log.Error(err)
 		sigc <- syscall.SIGINT
 		panic(err)
 	}
-	if db.Version < 8 {
+	if app.DB.Version < 8 {
 		log.Error("database schema v8 is required")
 		sigc <- syscall.SIGINT
 		panic(err)
 	}
-	archiveKeyList, err := config.GetC4GHprivateKeys()
+	app.ArchiveKeyList, err = config.GetC4GHprivateKeys()
 	if err != nil {
 		log.Error(err)
 		sigc <- syscall.SIGINT
 		panic(err)
 	}
-	archive, err := storage.NewBackend(conf.Archive)
+	app.Archive, err = storage.NewBackend(app.Conf.Archive)
 	if err != nil {
 		log.Error(err)
 		sigc <- syscall.SIGINT
 		panic(err)
 	}
-	inbox, err := storage.NewBackend(conf.Inbox)
+	app.Inbox, err = storage.NewBackend(app.Conf.Inbox)
 	if err != nil {
 		log.Error(err)
 		sigc <- syscall.SIGINT
 		panic(err)
 	}
 
-	defer mq.Channel.Close()
-	defer mq.Connection.Close()
-	defer db.Close()
+	defer app.MQ.Channel.Close()
+	defer app.MQ.Connection.Close()
+	defer app.DB.Close()
 
 	go func() {
-		connError := mq.ConnectionWatcher()
+		connError := app.MQ.ConnectionWatcher()
 		log.Error(connError)
 		forever <- false
 	}()
 
 	go func() {
-		connError := mq.ChannelWatcher()
+		connError := app.MQ.ChannelWatcher()
 		log.Error(connError)
 		forever <- false
 	}()
 
 	log.Info("starting ingest service")
-	var message schema.IngestionTrigger
 
 	go func() {
 		start := time.Now()
 		for i := 1; i > 0; i++ {
-			h, err := db.ListKeyHashes()
+			h, err := app.DB.ListKeyHashes()
 			if err != nil {
 				log.Errorln(err.Error())
 			}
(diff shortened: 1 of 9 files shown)


sda/cmd/verify/verify.go
```diff
@@ -392,6 +392,15 @@ func main() {
 					continue
 				}
 
+				if err := db.UpdateFileEventLog(message.FileID, "verified", delivered.CorrelationId, "ingest", "{}", string(verifiedMessage)); err != nil {
```
MalinAhlberg, line 395: why is this added..?

sda/internal/database/db_functions.go
```diff
@@ -202,34 +202,37 @@ func (dbs *SDAdb) storeHeader(header []byte, id string) error {
...
+		if err == nil {
+			break
+		}
+		time.Sleep(time.Duration(math.Pow(2, float64(count))) * time.Second)
 	}
 
 	return err
 }
-func (dbs *SDAdb) setArchived(file FileInfo, fileID, corrID string) error {
+func (dbs *SDAdb) setArchived(file FileInfo, fileID string) error {
```
MalinAhlberg, line 218: no longer updates the `file_event_log`?

sda/cmd/ingest/ingest.go
```diff
@@ -137,440 +149,445 @@ func main() {
...
-
-					// Restart on new message
-					continue
-				}
+func (app *Ingest) cancelFile(correlationID string, message schema.IngestionTrigger) string {
+	fileUUID, err := app.DB.GetFileID(correlationID)
+	if err != nil {
+		log.Errorf("failed to get ID for file from message: %s, reason: %s", correlationID, err.Error())
+		if strings.Contains(err.Error(), "sql: no rows in result set") {
+			return "reject"
```
MalinAhlberg, line 205: I don't follow the logic of `reject`s... Could you maybe add something in the PR description or in comments, or in a stand-up?

MalinAhlberg (commented):
I think the ingest loop refactoring is very nice :star:!

The whole refactoring is a bit hard to follow though, so sorry for the confused questions. There seems to be a bit more than just refactoring going on, and I don't really follow. So I put my initial questions down in the comments.

sda/cmd/ingest/ingest.go
```diff
@@ -137,440 +149,445 @@ func main() {
...
-					if err != nil {
-						log.Errorf("failed to get ID for file, reason: %s", err.Error())
-						if err := delivered.Nack(false, true); err != nil {
-							log.Errorf("failed to Nack message, reason: (%s)", err.Error())
-						}
-
-						continue
-					}
+			default:
+				log.Errorln("better message needed")
```
nanjiangshu, line 189: Will it be better to print the content of the message? Otherwise it is not easy to debug why it is failed. Also, it is not very clear what does `better message` mean.

sda/cmd/ingest/ingest.go
```diff
@@ -137,440 +149,445 @@ func main() {
...
-						log.Errorf("failed to publish message, reason: (%s)", err.Error())
-					}
-
-					// Restart on new message
-					continue
-				}
+func (app *Ingest) cancelFile(correlationID string, message schema.IngestionTrigger) string {
+	fileUUID, err := app.DB.GetFileID(correlationID)
+	if err != nil {
+		log.Errorf("failed to get ID for file from message: %s, reason: %s", correlationID, err.Error())
```
nanjiangshu, line 203:
```suggestion
		log.Errorf("failed to get ID for file from message (correlationID: %s), reason: %v", correlationID, err)
```

nanjiangshu: Otherwise, it is hard to know the value printed is correlationID without looking at the code.

sda/cmd/ingest/ingest.go
nanjiangshu: I'm also not super clear about the new logic, but it looks quite neat.

It seems each action will return one of the three status 
- `ack`
- `nack`
- `reject`

It it works well, maybe we can apply this logic to other services as well, or?

jbygdell: There are 3 responses to a message:
Ack - All ok
Nack - Failure, re queue message and try again
Reject - failure, don't re queue the message. Could as well has been Ack but used to be able to tell good  from bad results in the tests among other things. For example an error where the message is not to be re queued.

sda/cmd/ingest/ingest.go
jbygdell: Did I forget to change that? It should hold a relevant message.

sda/internal/database/db_functions.go
jbygdell: If you look in ingest.go where this is called from you will see that the file_event_log update is done separately.

sda/cmd/ingest/ingest.go
```diff
@@ -137,440 +148,403 @@ func main() {
...
-					body, _ := json.Marshal(fileError)
-					if err = mq.SendMessage(delivered.CorrelationId, conf.Broker.Exchange, "error", body); err != nil {
-						log.Errorf("failed to publish message, reason: (%s)", err.Error())
-					}
-
-					// Restart on new message
-					continue
-				}
+func (app *Ingest) cancelFile(correlationID string, message schema.IngestionTrigger) string {
+	fileUUID, err := app.DB.GetFileID(correlationID)
```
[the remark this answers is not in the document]
jbygdell: That should not be possible without an DB error.

sda/internal/database/db_functions.go
```diff
@@ -295,35 +298,42 @@ func (dbs *SDAdb) getHeader(fileID string) ([]byte, error) {
...
-		file.DecryptedChecksum,
-		"SHA256",
-	)
+	const verified = "UPDATE sda.files SET decrypted_file_size = $1 WHERE id = $2;"
+	if _, err := dbs.DB.Exec(verified, file.DecryptedSize, fileID); err != nil {
+		return fmt.Errorf("setVerified error: %s", err.Error())
+	}
 
-	return err
+	const addArchiveChecksum = "INSERT INTO sda.checksums(file_id, checksum, type, source)" +
```
nanjiangshu, line 322: I find long SQL code like this is not easy to read when using string concatenation like this. I think using raw string is easier to read and maintain and also easier to copy to SQL shell for manual debugging, for example

```golang
const addArchiveChecksum = `
INSERT INTO sda.checksums(file_id, checksum, type, source)
VALUES ($1, $2, upper($3)::sda.checksum_algorithm, upper('ARCHIVED')::sda.checksum_source)
ON CONFLICT ON CONSTRAINT unique_checksum
DO UPDATE SET checksum = EXCLUDED.checksum;
`
```

What do you think? Should we fixed in this refactoring PR or should I create a new issue for that?

nanjiangshu (commented):
I tested manually with repeated ingestion and the errors described in the issues are gone. Great work! Apart from that, I added a few minor comments.

sda/cmd/verify/verify.go
jbygdell: Because `setVerified` is moved to GO and to keep all calls identical in `s3inbox`, `ingest` and `verify`

This change is brought by the fact that ingest now can handle cases correctly, re-ingesting a file where a new version is uploaded to the inbox, with new checksum.

sda/internal/database/db_functions.go
jbygdell: In that case we probably should do a dedicated rework of all SQL statements as a separate PR since the current style is used elsewhere.

sda/cmd/ingest/ingest.go
jbygdell, line 189:
```suggestion
```

sda/cmd/ingest/ingest.go
```diff
@@ -137,440 +149,445 @@ func main() {
...
+	}
+	readBuffer := make([]byte, bufSize)
+	hash := sha256.New()
+	var bytesRead int64
+	var byteBuf bytes.Buffer
+
+	for bytesRead < fileSize {
+		i, _ := io.ReadFull(file, readBuffer)
+		if i == 0 {
+			log.Debugln("this should not happen")
```
MalinAhlberg, line 404: An error log would be useful here..?

sda/cmd/ingest/ingest.go
MalinAhlberg: Ok, I see, it sounds reasonable to me. I added a suggestion of a comment to the code above, for clarity for other confused developers.

sda/internal/database/db_functions.go
nanjiangshu: I created a new issue https://github.com/neicnordic/sensitive-data-archive/issues/1584 for it.

sda/cmd/ingest/ingest.go
```diff
@@ -137,440 +148,403 @@ func main() {
...
+		}
+	case "":
+		// Catch all for inboxes that doesn't update the DB
+		fileID, err = app.DB.RegisterFile(message.FilePath, message.User)
+		if err != nil {
+			log.Errorf("InsertFile failed, reason: (%s)", err.Error())
+
+			return "nack"
+		}
+	case "uploaded":
```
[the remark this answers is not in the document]
MalinAhlberg: I see, thanks.
But there is no case for `enabled` now?

sda/cmd/ingest/ingest.go
```diff
@@ -137,440 +149,444 @@ func main() {
...
+	}
+	readBuffer := make([]byte, bufSize)
+	hash := sha256.New()
+	var bytesRead int64
+	var byteBuf bytes.Buffer
+
+	for bytesRead < fileSize {
+		i, _ := io.ReadFull(file, readBuffer)
+		if i == 0 {
+			log.Debugln("this should not happen")
```
jbygdell, line 404:
```suggestion
			log.Errorln("readBuffer returned 0 bytes, this should not happen")
```

MalinAhlberg (approved):
I also tested to re-ingest, no loops occur anymore. Great work!

nanjiangshu (approved):
Looks good.

PR #1543, opened 2025-04-01
[s3inbox] Send remove message on overwrite

### LINKED ISSUE #1480: [s3inbox] send remove message if a file is overwritten.

If a submitters uploads a file that overwrites an existing file in the inbox a remove message needs to be sent before the new upload message can be sent. Otherwise the FEGA portal doesn't know that the new version of the file exists.

We know the path that we want to upload to. We should do a HEAD request and figure out if we will be overwriting. Then we send the messages from the acceptance criteria

## Acceptance criteria

- [ ] remove message is sent followed by the new upload message

## Estimation of size

small/medium/big

## Estimation of priority

low/medium/high

MalinAhlberg:
## Related issue(s) and PR(s)
This PR closes #1480 .

## Description
When uploading a file to s3inbox, a check is done to see if the file is already there. If it is, a delete message is sent before the upload message, so that the FEGA portal will know that the new version of the file exists.

~**Note**: The `remove` message is formatted according to [this spec](https://neic-sda.readthedocs.io/en/latest/connection/#message-format). However, our messages will have two extra fields, `"encrypted_checksums":null` and `"filesize":0`, since it shares the go struct with upload messages. Hopefully, that won't be any problem. Please let me know if you think it will.~

## How to test
Run the integration tests, and check that after the tests in `.github/integration/tests/sda/10_upload_test.sh` are run, there are 6 messages for `upload` and one for `remove` in  rabbitmq messages in the `inbox` queue.

You can also manually upload a file twice, and check the queue and the logs to make sure the `remove` message is only sent the second time:
```s3cmd -c /tmp/shared/s3cfg put dummyfile.c4gh s3://test_dummy.org/```

### DIFF
--- a/sda/cmd/s3inbox/proxy.go
+++ b/sda/cmd/s3inbox/proxy.go
@@ -21,6 +21,7 @@ import (
 	"github.com/minio/minio-go/v6/pkg/signer"
 	"github.com/neicnordic/sensitive-data-archive/internal/broker"
 	"github.com/neicnordic/sensitive-data-archive/internal/database"
+	"github.com/neicnordic/sensitive-data-archive/internal/schema"
 	"github.com/neicnordic/sensitive-data-archive/internal/storage"
 	"github.com/neicnordic/sensitive-data-archive/internal/userauth"
 	log "github.com/sirupsen/logrus"
@@ -156,8 +157,9 @@ func (p *Proxy) allowedResponse(w http.ResponseWriter, r *http.Request, token jw
 		return
 	}
 
-	// register file in database if it's the start of an upload
+	// if this is an upload request
 	if p.detectRequestType(r) == Put && p.fileIds[r.URL.Path] == "" {
+		// register file in database
 		log.Debugf("registering file %v in the database", r.URL.Path)
 		p.fileIds[r.URL.Path], err = p.database.RegisterFile(filepath, username)
 		log.Debugf("fileId: %v", p.fileIds[r.URL.Path])
@@ -166,6 +168,16 @@ func (p *Proxy) allowedResponse(w http.ResponseWriter, r *http.Request, token jw
 
 			return
 		}
+
+		// check if the file already exists, in that case send an overwrite message,
+		// so that the FEGA portal is informed that a new version of the file exists.
+		err = p.sendMessageOnOverwrite(r, rawFilepath, token)
+		if err != nil {
+			p.internalServerError(w, r, err.Error())
+
+			return
+		}
+
 	}
 
 	log.Debug("Forwarding to backend")
@@ -285,18 +297,21 @@ func (p *Proxy) checkAndSendMessage(jsonMessage []byte, r *http.Request) error {
 
 func (p *Proxy) uploadFinishedSuccessfully(req *http.Request, response *http.Response) bool {
 	if response.StatusCode != 200 {
+
 		return false
 	}
 
 	switch req.Method {
 	case http.MethodPut:
 		if !strings.Contains(req.URL.String(), "partNumber") {
+
 			return true
 		}
 
 		return false
 	case http.MethodPost:
 		if strings.Contains(req.URL.String(), "uploadId") {
+
 			return true
 		}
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: sda/cmd/s3inbox/proxy.go [hunk 1]
```diff
@@ -168,6 +168,35 @@ func (p *Proxy) allowedResponse(w http.ResponseWriter, r *http.Request, token jw
...
+	// check if the file already exists, in that case send an overwrite message,
+	// so that the FEGA portal is informed that a new version of the file exists.
+	exist := p.checkFileExists(r.URL.Path)
+	if exist {
+		log.Debug("create rewrite message")
+		event := Event{}
+		var err error
+		event.Operation = "remove"
+		event.Filepath = rawFilepath
+		event.Username = username
```
jbygdell, line 180: If you import `github.com/neicnordic/sensitive-data-archive/internal/schema` you will get access to the JSON schemas and the structs so you can validate the outgoing payload.

sda/cmd/s3inbox/proxy.go
```diff
@@ -168,6 +168,35 @@ func (p *Proxy) allowedResponse(w http.ResponseWriter, r *http.Request, token jw
 		}
 	}
 
+	// check if the file already exists, in that case send an overwrite message,
+	// so that the FEGA portal is informed that a new version of the file exists.
+	exist := p.checkFileExists(r.URL.Path)
```
jbygdell, line 173: An error here will assume that the file doesn't exist.

jbygdell (changes requested):
We have all the JSON schemas available so validation of the outgoing message can be done.

We should also not only check for the existence of the file but also handle any errors in a way so that the user gets the proper notification.

sda/cmd/s3inbox/proxy.go
```diff
@@ -511,6 +550,31 @@ func (p *Proxy) requestInfo(fullPath string) (string, int64, error) {
...
+		Key:    &filePath,
+	})
+	if err != nil {
+		// typically 404, if file is not found
+		log.Debugf("could not find file: %v", err)
+
+		return false, nil
+	}
+
+	return result != nil, nil
```
jbygdell, line 575:
```suggestion

	return result != nil, err
```

This can be shortened to this since the check for error comes first.

jbygdell (changes requested):
The `checkFileExists` should have a static test so we can validate the two expected results.

sda/cmd/s3inbox/proxy.go
MalinAhlberg: But that way we would forward error messages like "file not found". I would expect the function to return `false, nil` in case the file is not found, not `false, <an error including 404>`. Would you agree?
(It is also worth mentioning that those errors from s3 are hard to parse, as they don't have `err.Error()`.)

jbygdell: This is a bit of semantics, where do you parse the error, in the function or at the receiving end?

IF this function returns the raw error we can easily see them in the static code tests and implement the handling where the function is being called from.

sda/cmd/s3inbox/proxy.go
```diff
@@ -511,6 +526,66 @@ func (p *Proxy) requestInfo(fullPath string) (string, int64, error) {
...
+			FilePath:  rawFilepath,
+			Operation: "remove",
+		}
+
+		privateClaims := token.PrivateClaims()
+		log.Info("user ", msg.User, " with pilot ", privateClaims["pilot"], " will overwrite file ", msg.FilePath, " at ", time.Now())
+
+		jsonMessage, err := json.Marshal(msg)
+		if err != nil {
+
```
jbygdell, line 574:
```suggestion
```

sda/cmd/s3inbox/proxy.go
```diff
@@ -511,6 +526,66 @@ func (p *Proxy) requestInfo(fullPath string) (string, int64, error) {
...
+			return err
+		}
+
+		err = p.checkAndSendMessage(jsonMessage, r)
+		if err != nil {
+
```
jbygdell, line 580:
```suggestion
```

sda/cmd/s3inbox/proxy_test.go
```diff
@@ -451,11 +503,15 @@ func (suite *ProxyTests) TestDatabaseConnection() {
...
+	proxy := NewProxy(suite.S3Fakeconf, helper.NewAlwaysAllow(), messenger, database, new(tls.Config))
 
 	// PUT a file into the system
 	filename := "/dummy/db-test-file"
-	r, _ := http.NewRequest("PUT", filename, nil)
+
+	myString := "Hello, world!"
+	stringReader := strings.NewReader(myString)
+
+	r, _ := http.NewRequest("PUT", filename, stringReader)
```
jbygdell, line 514:
```suggestion
	stringReader := strings.NewReader("Hello, world!")
	r, _ := http.NewRequest("PUT", filename, stringReader)
```

Since the `myString` variable isn't going to be used it doesn't have to be declared, the string can be placed directly into the reader.

sda/cmd/s3inbox/proxy_test.go
```diff
@@ -10,8 +11,14 @@ import (
...
 	"net/url"
 	"os"
+	"strings"
 	"testing"
 
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/credentials"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+
```
jbygdell, line 21:
```suggestion
```

sda/cmd/s3inbox/proxy.go
```diff
@@ -511,6 +526,66 @@ func (p *Proxy) requestInfo(fullPath string) (string, int64, error) {
...
+		username := token.Subject()
+		msg := schema.InboxRemove{
+			User:      username,
+			FilePath:  rawFilepath,
+			Operation: "remove",
+		}
+
+		privateClaims := token.PrivateClaims()
+		log.Info("user ", msg.User, " with pilot ", privateClaims["pilot"], " will overwrite file ", msg.FilePath, " at ", time.Now())
```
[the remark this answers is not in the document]
MalinAhlberg: The error log is removed. It was not intended to be there.

MalinAhlberg:
I have fixed the things you comment on @jbygdell . Please have a look and resolve the comments.

Personally, I think our linters should be responsible for things like newlines. They should complain on everything we want to enforce. If they do not enforce a standard, it should be optional :) .

jbygdell:
> Personally, I think our linters should be responsible for things like newlines. They should complain on everything we want to enforce. If they do not enforce a standard, it should be optional :) .

The new version of the linter (that caused all the failures) will catch the empty lines among many other things in the codbase. So I'm just being preemptive here.

jbygdell (approved):

nanjiangshu (approved):
Works well when I tested. Great job!

PR #1594, opened 2025-04-10
[postgres] update constraint to also include the submission user

### LINKED ISSUE #1230: [Postgres] `CONSTRAINT unique_ingested` only looks at the `submission_file_path` and the `archive_file_path`
## Describe the bug

Since we prefer not to leak the submitter ID from the file path of the uploaded file we need to make sure that the constraint in the SQL schema doesn't only look at the submission file path and whether or not the file has been ingested already before giving it an ID.

Related to #1097 

## Steps to reproduce

Start the postgres container in any way you like.
Exec into the container and run the following queries.

```sql
SELECT sda.register_file('path/to/testfile.ex', 'testuser');
SELECT sda.register_file('path/to/testfile.ex', 'another-user');
```

You will see that the returned IDs are identical since the `CONSTRAINT unique_ingested` only looks at the `submission_file_path` and the `archive_file_path`

## Expected behavior
- [ ] Multiple users should be able to submit files with the exact same file path at the same time.

## out of scope
- [ ] A user should be able to submit files with the same name several times, given that the previous file has been assigned to a dataset. 

## Estimation of size
small

## Estimation of priority
low

jbygdell:
## Related issue(s) and PR(s)
This PR closes #1230 .

### DIFF
--- a/.github/integration/tests/postgres/20_inbox_queries.sh
+++ b/.github/integration/tests/postgres/20_inbox_queries.sh
@@ -10,6 +10,17 @@ for host in migrate postgres; do
         exit 1
     fi
 
+    newFileID=$(psql -U inbox -h "$host" -d sda -At -c "SELECT sda.register_file('inbox/test-file.c4gh', 'other-user');")
+    if [ -z "$newFileID" ]; then
+        echo "register_file failed"
+        exit 1
+    fi
+
+    if [ "$fileID" = "$newFileID" ]; then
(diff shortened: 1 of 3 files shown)


MalinAhlberg (approved):
Looks good to me!

Should it be merged to the feature branch (`feature/do-not-leak-user-ids`) instead of main..?

nanjiangshu (approved):

PR #1600, opened 2025-04-14
Tests/make verify fail

### LINKED ISSUE #1505: verify loops on double verification
## Describe the bug
When calling `verify` twice, it sometimes loops and start printing endless logs.

## Steps to reproduce
Checkout branch [`tests/make-verify-fail`]( https://github.com/neicnordic/sensitive-data-archive/commit/1a6efbdeb0e7fe93bf9fffdf483ae5ec3c8e9a01) and run `make integrationtest-sda-s3-run`. Keep an eye on the logs from `verify`.

Note that `verify` sometimes handles verifications fine, and then logs:
` verify  | {"level":"debug","msg":"file already verified","time":"2025-03-20T09:01:18Z"}`

note: the verification case comes from disabling-enabling the file.
the loop with logs case comes from double-verification.

## Expected behavior
- [ ] Verify does not loop 
- [ ] Tests verifying the fix are added (?)
- [ ] Should still log the above message but on warn level
- [ ] ack the message for verification

possible solution:
Remove "enabled" case from https://github.com/neicnordic/sensitive-data-archive/blob/964539f66ad3b4f3141287c1521e875a696e1516/sda/cmd/verify/verify.go#L356
and move out the check that we do in that. 

## Additional context
Created from #1463. Also see #1377 and #1082.

## Estimation of size
medium

## Estimation of priority
medium

aaperis:
## Related issue(s) and PR(s)
This PR closes #1505 .

### DIFF
--- a/sda/cmd/verify/verify.go
+++ b/sda/cmd/verify/verify.go
@@ -345,18 +345,30 @@ func main() {
 
 					continue
 				}
-				switch status {
-				case "disabled":
+
+				if status == "disabled" {
 					log.Infof("file with correlation ID: %s is disabled, stopping verification", delivered.CorrelationId)
 					if err := delivered.Ack(false); err != nil {
 						log.Errorf("Failed acking canceled work, reason: (%s)", err.Error())
 					}
 
 					continue
-				case "enabled":
-					fileInfo, err := db.GetFileInfo(message.FileID)
-					if err != nil {
-						log.Errorf("failed to get info for file: %s", message.FileID)
+				}
+
+				fileInfo, err := db.GetFileInfo(message.FileID)
+				if err != nil {
+					log.Errorf("failed to get info for file: %s", message.FileID)
+					if err := delivered.Nack(false, true); err != nil {
+						log.Errorf("failed to Nack message, reason: (%s)", err.Error())
+					}
+
+					continue
+				}
+
+				if fileInfo.DecryptedChecksum != "" {
+					log.Warnln("file already verified")
+					if err := mq.SendMessage(delivered.CorrelationId, conf.Broker.Exchange, conf.Broker.RoutingKey, verifiedMessage); err != nil {
+						log.Errorf("failed to publish message, reason: (%s)", err.Error())
 						if err := delivered.Nack(false, true); err != nil {
 							log.Errorf("failed to Nack message, reason: (%s)", err.Error())
 						}
@@ -364,23 +376,11 @@ func main() {
 						continue
 					}
 
-					if fileInfo.DecryptedChecksum != "" {
-						log.Debugln("file already verified")
-						if err := mq.SendMessage(delivered.CorrelationId, conf.Broker.Exchange, conf.Broker.RoutingKey, verifiedMessage); err != nil {
-							log.Errorf("failed to publish message, reason: (%s)", err.Error())
-							if err := delivered.Nack(false, true); err != nil {
-								log.Errorf("failed to Nack message, reason: (%s)", err.Error())
-							}
-
-							continue
-						}
-
-						if err := delivered.Ack(false); err != nil {
-							log.Errorf("failed to Ack message, reason: (%s)", err.Error())
-						}
-
-						continue
+					if err := delivered.Ack(false); err != nil {
+						log.Errorf("failed to Ack message, reason: (%s)", err.Error())
 					}
+
+					continue
 				}
 
 				if err := db.SetVerified(file, message.FileID); err != nil {
--- a/.github/integration/sda/rbac.json
+++ b/.github/integration/sda/rbac.json
@@ -15,6 +15,11 @@
          "path": "/dataset/*",
          "action": "(POST)|(PUT)"
       },
+      {
+         "role": "submission",
+         "path": "/dataset/*",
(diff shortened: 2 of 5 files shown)


sda/cmd/verify/verify.go
```diff
@@ -345,60 +345,56 @@ func main() {
...
-							if err := delivered.Nack(false, true); err != nil {
-								log.Errorf("failed to Nack message, reason: (%s)", err.Error())
-							}
+					continue
+				}
 
-							continue
-						}
+				switch {
+				case fileInfo.DecryptedChecksum != "" && fileInfo.DecryptedChecksum == fmt.Sprintf("%x", sha256hash.Sum(nil)):
```
jbygdell, line 369:
```suggestion
				case fileInfo.DecryptedChecksum == fmt.Sprintf("%x", sha256hash.Sum(nil)):
```

The second part can handle all three possible combinations alone.

sda/cmd/verify/verify.go
```diff
@@ -345,60 +345,56 @@ func main() {
...
-							log.Errorf("failed to publish message, reason: (%s)", err.Error())
-							if err := delivered.Nack(false, true); err != nil {
-								log.Errorf("failed to Nack message, reason: (%s)", err.Error())
-							}
+					continue
+				}
 
-							continue
-						}
+				switch {
```
jbygdell, line 368: Technically this switch is not needed. If the `db.UpdateFileEventLog` is moved outside it can be simplified to:
```
if  fmt.Sprintf("%x", sha256hash.Sum(nil)) != fileInfo.DecryptedChecksum {
	if err := db.SetVerified(file, message.FileID); err != nil {
		log.Errorf("SetVerified failed, reason: (%s)", err.Error())
		if err := delivered.Nack(false, true); err != nil {
			log.Errorf("failed to Nack message, reason: (%s)", err.Error())
		}

		continue
	}
}			
```

jbygdell (changes requested):
The changes to the api is not related to the scope of the issue

sda/cmd/verify/verify.go
```diff
@@ -345,60 +345,56 @@ func main() {
...
+				case fileInfo.DecryptedChecksum != "" && fileInfo.DecryptedChecksum == fmt.Sprintf("%x", sha256hash.Sum(nil)):
+					log.Warnln("file already verified")
 
-						if err := delivered.Ack(false); err != nil {
-							log.Errorf("failed to Ack message, reason: (%s)", err.Error())
+					if err := db.UpdateFileEventLog(message.FileID, "verified", delivered.CorrelationId, "ingest", "{}", string(verifiedMessage)); err != nil {
```
MalinAhlberg, line 372: What is the reasoning for updating the event log here..?

MalinAhlberg (approved):
Good work!
If one wants to keep the extra test I added (2e48725e157340d21010c1ce2e557977e4faf07e) one could update `5` to `6` [here](https://github.com/neicnordic/sensitive-data-archive/blob/2684564e4ad914a218f2792a405c7b8d4aa6ad0b/.github/integration/tests/sda/21_cancel_test.sh#L89). Then the tests pass (on my machine at least :stuck_out_tongue_winking_eye: ).
Anyway, I approve.

sda/cmd/verify/verify.go
aaperis: This came about because of [this addition](https://github.com/neicnordic/sensitive-data-archive/pull/1526/files#r2031455478) which changed the behavior of verify (and also basically the current issue).

aaperis:
> Good work! If one wants to keep the extra test I added ([2e48725](https://github.com/neicnordic/sensitive-data-archive/commit/2e48725e157340d21010c1ce2e557977e4faf07e)) one could update `5` to `6` [here](https://github.com/neicnordic/sensitive-data-archive/blob/2684564e4ad914a218f2792a405c7b8d4aa6ad0b/.github/integration/tests/sda/21_cancel_test.sh#L89). Then the tests pass (on my machine at least 😜 ). Anyway, I approve.

Since rebasing to the merged code of #1526, the error that the code you had added changed. Hence it is only meaningful to check if the file is already verified, which is implicitly done in subsequent tests of the workflow. That is why I removed this part :sunglasses: .

sda/cmd/verify/verify.go
```diff
@@ -344,51 +344,37 @@ func main() {
...
-
-				if err := db.SetVerified(file, message.FileID); err != nil {
-					log.Errorf("SetVerified failed, reason: (%s)", err.Error())
-					if err := delivered.Nack(false, true); err != nil {
-						log.Errorf("failed to Nack message, reason: (%s)", err.Error())
-					}
-
-					continue
+				} else {
+					log.Warnln("file already verified")
```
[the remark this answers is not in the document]
aaperis: There needs to be a way to somehow follow what happens during ingestion. I think this is not superfluous as it is.

jbygdell, line 377:
```suggestion
					log.Infof("file with correlation id: %s is already verified", delivered.CorrelationId)
```

jbygdell (approved):

nanjiangshu (approved):
Looks great!

Issue #1589, opened 2025-04-09
Test if the user id migration script breaks anything in our stack


pahatz:
From the migration script for removing user ids issue #1123
Run the script and check if anything breaks in the services. This should be done in the branch `remove-user-ids-from-filepaths`
things to test: run the integration test suite.

jbygdell:
```sql
UPDATE sda.files SET submission_file_path = REPLACE(submission_file_path, CONCAT(REPLACE(submission_user, '@', '_'),'/'), '');
```

PR #1601, opened 2025-04-14
Do not leak user IDs

jbygdell:
## Related issue(s) and PR(s)

## Description
Used for validation of #1589


charts/sda-svc/README.md
```diff
@@ -18,8 +18,13 @@ A major chart version change (like v1.2.3 -> v2.0.0) indicates that there is an
 
 ### To 3.0.0
 
-This version adds Jobs that migrates the database schema from a V1 release and sets the first registered crypt4gh key to all ingested files that do not have a c4gh key referenced.  
-When upgrading from a V1 release *both* `upgradeFromV1` and `setKeyHash` should be run.
+This version adds Jobs that performes the following actions:
```
nanjiangshu, line 21:
```suggestion
This version adds jobs that perform the following actions:
```

nanjiangshu (commented):
Looks good except for some minor comments.

.github/integration/tests/sda/20_ingest-verify_test.sh
```diff
@@ -32,6 +32,10 @@ for file in NA12878.bam NA12878_20k_b37.bam NA12878.bai NA12878_20k_b37.bai; do
             '$ARGS.named|to_entries|map(with_entries(select(.key=="key").key="type"))'
     )
 
+    if [ "$file" = "NA12878.bam" ]; then
+        file="test_dummy.org/NA12878.bam"
```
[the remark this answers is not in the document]
jbygdell: This is a sanity check that if there are old messages hanging around we should still be able to process them correctly.

sda/cmd/s3inbox/proxy.go
```diff
@@ -192,7 +194,7 @@ func (p *Proxy) allowedResponse(w http.ResponseWriter, r *http.Request, token jw
 	// nolint: nestif // We need a nested if statement for checking whether fileId is persisted during possible reconnections
 	if p.uploadFinishedSuccessfully(r, s3response) {
 		log.Debug("create message")
-		message, err := p.CreateMessageFromRequest(r, token)
+		message, err := p.CreateMessageFromRequest(r, token, anonymizedFilepath)
```
KarlG-nbis, line 197: Here we are passing `anonymizedFilepath` to `CreateMessageFromRequest` but the new arg we added is `user string`, so this doesnt seem correct, and we should rename the variable if we dont expect to pass the user (but from func implementation we do expect user), but also since in `CreateMessageFromRequest` we also construct the anonymizedFilepath, so possibly we just just update the `CreateMessageFromRequest` to take the `anonymizedFilepath` so we do not need construct it again

KarlG-nbis (approved):

nanjiangshu (approved):
Looks good!

PR #1640, opened 2025-04-30
Bugfix/OIDC aud field check

FracassandoCasualmente:
## Related issue(s) and PR(s)
This PR closes [1639].

## Description
Resolves a vulnerability where a rogue server could impersonate any user who logs into it, enabling unauthorized access to any SDA instance using LS AAI.

## How to test
- Run `make sda-s3-up`
- Log into the rogue service at http://localhost:8802/ as a dummy user
- Get the access token from the rogue server and put it into a "token" variable in the terminal
- Run `curl -H "Authorization: Bearer $token" localhost:8090/files`
- Check that the api returns an error response
- Download the submission credentials file from the rogue service
- Change the "host_base" in the configuration to "http://localhost:18000"
- Run `sda-cli -config s3cmd.conf list` and check that the inbox returns an error response
- Repeat the same steps but by using the credentials from the real auth service at http://localhost:8801 and check that they both return a non-error response

### DIFF
--- a/sda/internal/userauth/userauth.go
+++ b/sda/internal/userauth/userauth.go
@@ -8,13 +8,26 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 
 	"github.com/lestrrat-go/jwx/v2/jwk"
 	"github.com/lestrrat-go/jwx/v2/jws"
 	"github.com/lestrrat-go/jwx/v2/jwt"
+	"github.com/neicnordic/sensitive-data-archive/internal/config"
 	log "github.com/sirupsen/logrus"
 )
+
+var configAuth *config.Config
+
+func init() {
+	// initialize config
+	var err error
+	configAuth, err = config.NewConfig("auth")
+	if err != nil {
+		log.Debugf("failed to get auth config, reason: %v", err)
+	}
+}
 
 // Authenticator is an interface that takes care of authenticating users to the
 // S3 proxy. It contains only one method, Authenticate.
@@ -58,6 +71,11 @@ func (u *ValidateFromToken) Authenticate(r *http.Request) (jwt.Token, error) {
 			return nil, fmt.Errorf("failed to get issuer from token (%v)", iss)
 		}
 
+		err = checkAudience(token)
+		if err != nil {
+			return nil, fmt.Errorf("failed to check audience (%v)", err)
+		}
+
 		return token, nil
 
 	case r.Header.Get("Authorization") != "":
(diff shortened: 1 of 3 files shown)


sda/internal/userauth/userauth.go
```diff
@@ -8,14 +8,27 @@ import (
...
+
+func init() {
+	// initialize config
+	var err error
+	configAuth, err = config.NewConfig("auth")
+	if err != nil {
+		log.Debugf("failed to get auth config, reason: %v", err)
+	}
+}
+
```
jbygdell, line 31:
```suggestion
```

This is an internal component used by several applications, as such distinct application config should no take place here.

sda/internal/userauth/userauth.go
```diff
@@ -58,6 +71,11 @@ func (u *ValidateFromToken) Authenticate(r *http.Request) (jwt.Token, error) {
 			return nil, fmt.Errorf("failed to get issuer from token (%v)", iss)
 		}
 
+		err = checkAudience(token)
+		if err != nil {
+			return nil, fmt.Errorf("failed to check audience (%v)", err)
+		}
+
```
jbygdell, line 78:
```suggestion
```

Can be handled by the `jwt.Parse` function.

sda/internal/userauth/userauth.go
```diff
@@ -76,13 +94,37 @@ func (u *ValidateFromToken) Authenticate(r *http.Request) (jwt.Token, error) {
...
```
jbygdell, line 101:
```suggestion
```

Can be handled by the `jwt.Parse` function.

sda/internal/userauth/userauth.go
```diff
@@ -76,13 +94,37 @@ func (u *ValidateFromToken) Authenticate(r *http.Request) (jwt.Token, error) {
...
+	}
+
+	if !slices.Contains(audList, correctAud) {
+		return fmt.Errorf("access token from wrong OIDC client. Expected: %v Received: %v",
+			correctAud, audList)
+	}
+
+	return nil
+}
+
```
jbygdell, line 127:
```suggestion
```

The `jwt.Parse` function has this functionality built in so there is no need to do this manually.

sda/internal/userauth/userauth.go
```diff
@@ -8,14 +8,27 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"slices"
```
jbygdell, line 11:
```suggestion
```

sda/internal/userauth/userauth.go
```diff
@@ -8,14 +8,27 @@ import (
...
 	"strings"
 
 	"github.com/lestrrat-go/jwx/v2/jwk"
 	"github.com/lestrrat-go/jwx/v2/jws"
 	"github.com/lestrrat-go/jwx/v2/jwt"
+	"github.com/neicnordic/sensitive-data-archive/internal/config"
```
jbygdell, line 17:
```suggestion
```

jbygdell (changes requested):
Hi,

Thank you for your contribution.

We agree that this is something that needs addressing. Since the token authentication is used by several applications this will require changes to other parts of the code as well, mainly in the authenticator initiation and configurations. Which  in turn requires changes to the helm chart and the test suite.

FracassandoCasualmente:
> Hi,
> 
> Thank you for your contribution.
> 
> We agree that this is something that needs addressing. Since the token authentication is used by several applications this will require changes to other parts of the code as well, mainly in the authenticator initiation and configurations. Which in turn requires changes to the helm chart and the test suite.

Hello! 
Thanks for reply and for the corrections.

Regarding the changes that are missing, I don't really have experience with helm charts nor have broad knowledge of the code base, so it would take me some time (1+ month?) to fix.
I'm happy to take the time to handle this, especially if your team feels short-staffed or overwheled. However, if you think it would take too long for the change to reach production, feel free to close this PR and handle it internally. :)

PR #1652, opened 2025-05-06
[Release V1] backport PR#1521

jbygdell:
## Description
Fixes failing tests due to java issues.

### DIFF
--- a/sda-sftp-inbox/pom.xml
+++ b/sda-sftp-inbox/pom.xml
@@ -24,6 +24,18 @@
         <java.version>19</java.version>
     </properties>
 
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.junit</groupId>
+                <artifactId>junit-bom</artifactId>
+                <version>5.12.1</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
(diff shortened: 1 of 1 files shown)


jbygdell:
@Parisa68 could you look into what causes the DOA container build to fail?

Parisa68:
> @Parisa68 could you look into what causes the DOA container build to fail?

the settings.xml being made in this action is for the newer Norwegian repository, and the sda-doa pom file in v1 is for the older Norwegian repo. so the authorization fails. I will update the pom file in an hour.

kjellp (commented):
@jbygdell Now the sda-doa builds correctly. The 2 sda-doa integration tests (posix and S3) will not succeed before we backport more from main in the later planned PR.

But there is a 3rd failing test to look into, is it related to what this PR should fix and thus need more work?

kjellp (approved):
The previously failing test succeeded in second attempt. Approving now as the sda-doa integration tests are not expected to succeed yet.

jbygdell:
One question before we merge this one.
How much work would it take to make the failing tests pass?

Parisa68:
> One question before we merge this one. How much work would it take to make the failing tests pass?

It can be done just by backporting sda-doa. the major difference between current doa(in main) and the doa present in release_v1 is related to the integration test (I think 1 or 2 features have also been added).

PR #1654, opened 2025-05-06
[release_v1] Run workflows on release_v1

Parisa68:
**DO NOT MERGE until just before the release v1.0.0. is ready**

PR #1657, opened 2025-05-07
[release_v1] Update GO to 1.24

jbygdell:
## Description
Update the `release_v1` branch to use GO 1.24 with the latest library versions


Parisa68:
Based on the plan, this is the PR that should create v1.0.0. So before this goes in, we should first merge pr #1654, which prepares the workflow, and then merge this one.

jbygdell:
> Based on the plan, this is the PR that should create v1.0.0. So before this goes in, we should first merge pr #1654, which prepares the workflow, and then merge this one.

As per the comment in the other PR "You do not want DOA in v1.0.0 but rather in v1.1.0"?

Parisa68:
> > Based on the plan, this is the PR that should create v1.0.0. So before this goes in, we should first merge pr #1654, which prepares the workflow, and then merge this one.
> 
> As per the comment in the other PR "You do not want DOA in v1.0.0 but rather in v1.1.0"?

Correct. we still think it is best to have doa merged in a later pr that creates v1.1.0.

jbygdell:
Converting to draft since the worflow releaser needs to be merged first.

kjellp:
@jbygdell We had a look at both the code changes and the test outputs and think we are mostly ready to merge soon. We also noticed that you have already put the #major string ready in one commit message :) But we can formally approve this PR when it's still a Draft.

The #1654 PR is also ready, awaiting your approval. They should be merged quickly on after eachother (to avoid someone accidently approving a release_v1 dependabot PR in between, that would cause a tag version conflict we dont want).

So maybe we need a quick work session together to do the last clicks in the right order? Will propose a meeting tomorrow morning is a calendar invite.

Parisa68 (approved):

kjellp (approved):

PR #1658, opened 2025-05-09
[charts] require a pubkey for s3inbox when resigning jwt

### LINKED ISSUE #722: [charts][sda-svc] add assertions when deploying with `jwtResign = true`
**Describe the bug**
When jwtResign is set to true in sda-auth, s3inbox needs to have access to the public key with which the jwt is resigned in sda-auth. Currently t is possible to configure Helm values for the sda-svc charts with jwtResign=true and without configuring the pub key for s3inbox. This will lead to s3inbox malfunctioning after it is deployed.

**Expected behavior**
- The chart needs to verify that the jwt public key has been defined for S3inbox when jwtResign is true for auth.
- Helm should fail if resignJwt is not  configured properly 

**Additional context**
In addition, `sda-auth` needs to check and fail early if the jwt private key is not specified (maybe this is fixed for >v0.2.56)
- EDIT: both jwtPrivateKey and c4gh pub key are checked and auth fails early in case of error so this holds true. 

**Estimation of size**: small/medium

aaperis:
## Related issue(s) and PR(s)
This PR closes #722 .

## How to test
Deploy charts locally using make commands. Then `helm uninstall pipeline` and `vi /tmp/values.yaml` and comment out the `jwtPub` line and redeploy the pipeline stack. You should see an explanatory error message.

### DIFF
--- a/charts/sda-svc/Chart.yaml
+++ b/charts/sda-svc/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: sda-svc
-version: 0.30.14
+version: 0.30.15
 appVersion: v0.5.14
 kubeVersion: '>= 1.26.0'
 description: Components for Sensitive Data Archive (SDA) installation
--- a/charts/sda-svc/templates/s3-inbox-deploy.yaml
+++ b/charts/sda-svc/templates/s3-inbox-deploy.yaml
@@ -190,14 +190,16 @@ spec:
         - name: SERVER_KEY
           value: {{ include "tlsPath" . }}/tls.key
       {{- end }}
-        {{- if.Values.global.auth.jwtPub }}
+      {{- if .Values.global.auth.jwtPub }}
         - name: SERVER_JWTPUBKEYPATH
           value: {{ include "jwtPath" . }}
-        {{- end }}
-        {{- if not .Values.global.auth.resignJwt }}
+      {{- end }}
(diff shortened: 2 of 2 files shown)


jbygdell (changes requested):
The Chart.yaml needs to have a version bump in place other wise this change will not get published.

aaperis:
> The Chart.yaml needs to have a version bump in place other wise this change will not get published.

indeed, I forgot to update it!

jbygdell (approved):

nanjiangshu (commented):
Looks good.

nanjiangshu (approved):

nanjiangshu:
> > The Chart.yaml needs to have a version bump in place other wise this change will not get published.
> 
> indeed, I forgot to update it!

I don't quite understand the version bump here. Should we let the `publish_container` workflow handle version bumping, or do it manually when we change the Helm charts? @aaperis @jbygdell

jbygdell:
> > > The Chart.yaml needs to have a version bump in place other wise this change will not get published.
> > 
> > 
> > indeed, I forgot to update it!
> 
> I don't quite understand the version bump here. Should we let the `publish_container` workflow handle version bumping, or do it manually when we change the Helm charts? @aaperis @jbygdell

When there are only changes in the chart folders the auto bumping doesn't happen since the containers are not being built.
Larger changes to only the charts would warrant a minor or even major bump and those needs to be done manually.

PR #1627, opened 2025-04-25
Download re-encrypted file from the inbox through API

nanjiangshu:
## Related issue(s) and PR(s)
This PR closes #1346 

#### If you want to test manually

1. Start the docker setup of the integration test
2. Upload two files, ingest one of the file using admin-API
3. Try to download and decrypt both of them by 

```sh
pubkey=$(base64 -w0 <your-pub-key-file>)
curl -v -H "Authorization: Bearer $ACCESS_TOKEN"  -H "C4GH-Public-Key: $pubkey" -X GET ${API_HOST}/users/${user}/file/${fileid}  -o  t1.txt.c4gh

sda-cli decrypt -key <your-private-key-file> t1.txt.c4gh
```

PR #1661, opened 2025-05-09
Fix linting function in Makefile

### LINKED ISSUE #1659: Linting command in Makefile is broken
## Describe the bug

When running 
```sh
make lint-sda
```

It returns the  error 

```
Running golangci-lint in the `sda` folder
Error: unknown shorthand flag: 'e' in -e
Failed executing command with error: unknown shorthand flag: 'e' in -e
make: *** [lint-sda] Error 3
```

This is because the current setup use version 2 of `golangci-lint`.

## Expected behavior
- [ ] 
- [ ] linting commands in Makefile should work

## Additional context
Related comment: https://github.com/neicnordic/sensitive-data-archive/pull/1627#discussion_r2081488037
## Estimation of size
small/medium/big

## Estimation of priority
low/medium/high

nanjiangshu:
## Related issue(s) and PR(s)
This PR closes #1659 

## Description
This PR was separated from the commit previously pushed to the PR https://github.com/neicnordic/sensitive-data-archive/pull/1627. See the comment https://github.com/neicnordic/sensitive-data-archive/pull/1627#discussion_r2081488037 for more details. 

## How to test

`make lint-all` should work and not report any error.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -58,3 +58,5 @@ linters:
         disabled: true
       - name: unused-receiver # disable for now
         disabled: true
+      - name: unhandled-error # disable for now
+        disabled: true
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,3 @@
-# linter modules to be included/excluded
-LINT_INCLUDE=-E bodyclose,gocritic,gofmt,gosec,govet,nestif,nlreturn,revive,rowserrcheck
-LINT_EXCLUDE=-e G401,G501,G107
-
 help:
 	@echo 'Welcome!'
 	@echo ''


.golangci.yml
```diff
@@ -58,3 +58,5 @@ linters:
         disabled: true
       - name: unused-receiver # disable for now
         disabled: true
+      - name: unhandled-error # disable for now
+        disabled: true
```
aaperis, line 62:
```suggestion
```
I think this is not needed.

aaperis:
I think the only thing you need to do is replace the three occurrences of: 
```golangci-lint run $(LINT_INCLUDE) $(LINT_EXCLUDE)```
with 
```golangci-lint run -c ../.golangci.yml```
and you're good to go.

**Review by MalinAhlberg — COMMENTED** [review]
Nice that this is fixed!
Like Alex, I'm not sure why `unhandled-error` is removed..?

Apart from that, I think it looks good :rocket:

File: .golangci.yml (hunk 1 above)
Comment (nanjiangshu) [reply .golangci.yml:62]: Without this, one gets lots of unhandled-errors such as 
```
sda/cmd/api/api_test.go:135:3: unhandled-error: Unhandled error in call to function io.Closer.Close (revive)
		res.Body.Close()
		^
sda/cmd/api/api_test.go:196:3: unhandled-error: Unhandled error in call to function io.Closer.Close (revive)
		res.Body.Close()
		^
sda/cmd/notify/notify_test.go:127:3: unhandled-error: Unhandled error in call to function fmt.Println (revive)
		fmt.Println(err)
```

nanjiangshu:
> I think you just need to replace the three occurrences of: `golangci-lint run $(LINT_INCLUDE) $(LINT_EXCLUDE)` with `golangci-lint run -c ../.golangci.yml` and you're good to go.

Good point! I will remove `$(LINT_INCLUDE) $(LINT_EXCLUDE)` since they are basically just empty string after remove the ENVs.

However, it seems the argument `-c ../.golangci.yml` is not needed since by default the config file `.golangci.yml` will be loaded.

.golangci.yml
jbygdell: There is a specific check for body-close and similar. Do not blatantly ignore all error returns.

jbygdell: But there should not be any need for changes to the linter config.

nanjiangshu: That makes sense. Maybe we should fix them by adding `//nolint:errcheck` to these lines of code? 
I will remove the setting for `unhandled-error`.

aaperis (approved):

jbygdell (approved):

PR #1663, opened 2025-05-12
[HotFix] Ensure cryp4gh keys are loaded on startup.

### LINKED ISSUE #1597: [Ingest] don't fail startup if no c4gh key is configured
## Describe the bug
If the `c4gh.privateKeys` list is empty, Ingest will happily start anyway.

## Expected behavior
- [ ] Ingest should fail hard and loud if no crypt4gh key has been configured

## Additional context
Affected versions >= v0.3.180

## Estimation of size
small

## Estimation of priority
high

### LINKED ISSUE #1662: Applications start without configured c4gh keys
## Describe the bug

Both ingest and verify will happily start even if no crypt4gh keys have been configured.
No errors will be reported until the apps try to decrypt a file during the work process.

## Steps to reproduce

Remove the `c4gh.privateKeys` block from the config file and run `make integrationtest-sda-s3-run`

## Expected behavior
- [ ] Ingest and verify should fail fast and hard if no crypt4gh keys are configured.

## Estimation of size
small

## Estimation of priority
high

jbygdell:
## Related issue(s) and PR(s)
This PR closes #1597  
This PR closes #1662.

## Description
This PR fixes a bug in the code where `Ingest` and `Verify` would happily start even when no crypt4gh keys have been configured.

## How to test
Remove the `c4gh.privateKeys` block from the config file and run `make sda-s3-up`

## Extra info

Chart tests will fail since the charts are not updated to handle multiple keys yet. (WIP to be completed once this is merged.)

### DIFF
--- a/sda/cmd/ingest/ingest.go
+++ b/sda/cmd/ingest/ingest.go
@@ -8,6 +8,7 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -75,10 +76,9 @@ func main() {
 		panic(err)
 	}
 	app.ArchiveKeyList, err = config.GetC4GHprivateKeys()
-	if err != nil {
-		log.Error(err)
+	if err != nil || len(app.ArchiveKeyList) == 0 {
 		sigc <- syscall.SIGINT
-		panic(err)
+		panic(errors.New("no C4GH private keys configured"))
 	}
 	app.Archive, err = storage.NewBackend(app.Conf.Archive)
 	if err != nil {
--- a/sda/cmd/ingest/ingest_test.go
+++ b/sda/cmd/ingest/ingest_test.go
@@ -169,13 +169,18 @@ type TestSuite struct {
 	filePath   string
 	pubKeyList [][32]byte
 	ingest     Ingest
+	tempDir    string
 }
 
 func (ts *TestSuite) SetupSuite() {
+	var err error
 	viper.Set("log.level", "debug")
-	tempDir := ts.T().TempDir()
-	keyFile1 := fmt.Sprintf("%s/c4gh1.key", tempDir)
-	keyFile2 := fmt.Sprintf("%s/c4gh2.key", tempDir)
+	ts.tempDir, err = os.MkdirTemp("", "c4gh-keys")
+	if err != nil {
+		ts.FailNow("Failed to create temp directory")
+	}
+	keyFile1 := fmt.Sprintf("%s/c4gh1.key", ts.tempDir)
+	keyFile2 := fmt.Sprintf("%s/c4gh2.key", ts.tempDir)
 
 	publicKey, err := helper.CreatePrivateKeyFile(keyFile1, "test")
 	if err != nil {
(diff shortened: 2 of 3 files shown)


sda/cmd/verify/verify.go
```diff
@@ -41,8 +41,8 @@ func main() {
 		log.Fatal(err)
 	}
 	archiveKeyList, err := config.GetC4GHprivateKeys()
-	if err != nil {
-		log.Fatal(err)
+	if err != nil || len(archiveKeyList) == 0 {
+		log.Fatal("no C4GH private keys configured")
```
pontus, line 45: Would be nice to provide error if this is because of one?

sda/cmd/ingest/ingest.go
```diff
@@ -75,10 +76,9 @@ func main() {
 		panic(err)
 	}
 	app.ArchiveKeyList, err = config.GetC4GHprivateKeys()
-	if err != nil {
-		log.Error(err)
+	if err != nil || len(app.ArchiveKeyList) == 0 {
 		sigc <- syscall.SIGINT
-		panic(err)
+		panic(errors.New("no C4GH private keys configured"))
```
pontus, line 81: If we're failing because of an error, it's helpful showing what.

pontus (approved):

sda/cmd/verify/verify.go
nanjiangshu: I agree. Maybe something like this:

```suggestion
if err != nil {
	log.Fatalf("encountered an error: %v", err)
}
if len(archiveKeyList) == 0 {
	log.Fatal("no C4GH private keys configured")
}
```

nanjiangshu (commented):
Looks good to me except for the comment about the error message handling. In addition, the deployment of helm chart (ingest and verify services) is failing because of missing c4gh keys.

sda/cmd/verify/verify.go
jbygdell: The errors are printed in the`GetC4GHprivateKeys` function

sda/cmd/ingest/ingest.go
jbygdell: The errors are printed in the`GetC4GHprivateKeys` function

pontus: It's seldom I feel there's too much information about why something fails and I would prefer not to rely on some other part of code to do things that aren't obviously part of the contract.

But if you feel it's enough it's approved.

jbygdell:
> In addition, the deployment of helm chart (ingest and verify services) is failing because of missing c4gh keys.

I know, it is in the PR description..
But it is a catch 22 thing, Merging this PR will make fixing the charts much easier since these two services will fail if not configured properly.

nanjiangshu (approved):

sda/cmd/ingest/ingest_test.go
```diff
@@ -723,3 +735,9 @@ func (ts *TestSuite) TestIngestFile_missingFile() {
 
 	assert.Equal(ts.T(), "ack", ts.ingest.ingestFile(corrID, message))
 }
+func (ts *TestSuite) TestDetectMisingC4GHKeys() {
+	viper.Set("c4gh.privateKeys", "")
+	privateKeys, err := config.GetC4GHprivateKeys()
+	assert.NoError(ts.T(), err)
+	assert.Equal(ts.T(), 0, len(privateKeys))
+}
```
aaperis, line 743: What does this part check? Does it check that `config.GetC4GHprivateKeys` translates an empty string to an empty array?

jbygdell: Sanity check for the current functionality.
It should be possible to have the function to return an error if the marshaled struct is empty. It was a bit to time consuming to get it to work for this PR.

PR #1764, opened 2025-06-17
[config] Adjust log levels

### LINKED ISSUE #1585: [sync] Reading config file logs should be on debug level.

I believe that all of these (except one) should be Debug logs:

https://github.com/neicnordic/sensitive-data-archive/blob/2a81441691eccbe44d286ea678752ece80061bc1/sda/internal/config/config.go#L169-L207

Except for when failing to parse, shouldn't that log an Error (or at least Warning):

https://github.com/neicnordic/sensitive-data-archive/blob/2a81441691eccbe44d286ea678752ece80061bc1/sda/internal/config/config.go#L186

## Acceptance criteria

- [ ] Correct logging levels

## Estimation of size

small

## Estimation of priority

high

zeidlitz:
## Related issue(s) and PR(s)
This PR closes #1585 

## Description
Adjusting log levels from info to debug except for one case that is changed to warn


sda/internal/config/config.go
```diff
@@ -192,29 +192,29 @@ func NewConfig(app string) (*Config, error) {
...
-		log.Infoln(err.Error())
+		log.Debugf("Error reading config : %v", err.Error())
 		if _, ok := err.(viper.ConfigFileNotFoundError); !ok {
-			log.Infoln("ReadInConfig Error")
+			log.Warnf("ReadInConfig Error : %v", err.Error())
 
 			return nil, err
 		}
-		log.Infoln("No config file found, using ENVs only")
+		log.Debugf("No config file found, using ENVs only")
```
MalinAhlberg, line 203: I imagine that this could be useful to have as info..?

sda/internal/config/config.go
```diff
@@ -192,29 +192,29 @@ func NewConfig(app string) (*Config, error) {
 	if viper.IsSet("configFile") {
 		viper.SetConfigFile(viper.GetString("configFile"))
 	}
-	log.Infoln("reading config")
+	log.Debugf("Attempting to read configuration...")
 	if err := viper.ReadInConfig(); err != nil {
-		log.Infoln(err.Error())
+		log.Debugf("Error reading config : %v", err.Error())
```
jbygdell, line 197:
```suggestion
		log.Debugf("Error reading config : %s", err.Error())
```

`err.Error()` is by definition a string.

The alternative would be:

```
		log.Debugf("Error reading config : %v", err)
```

sda/internal/config/config.go
```diff
@@ -192,29 +192,29 @@ func NewConfig(app string) (*Config, error) {
...
 		if _, ok := err.(viper.ConfigFileNotFoundError); !ok {
-			log.Infoln("ReadInConfig Error")
+			log.Warnf("ReadInConfig Error : %v", err.Error())
```
jbygdell, line 199: See comment above.

jbygdell (changes requested):

sda/internal/config/config.go
nanjiangshu: I think this is debug info. The problem is that should we debug the production server when there is a problem? In principle, debugging should be done with the staging server which has log level set to debug, but in reality...

nanjiangshu: I think this is debug info. The problem is that should we debug the production server when there is a problem? In principle, debugging should be done with the staging server which has log level set to debug, but in reality...

MalinAhlberg (approved):
:+1:

PR #1792, opened 2025-07-01
[api] Optimize db queries for /users and /files

aaperis:
## Description
The database has grown and the `sda-api` struggles to satisfy requests because of inefficient queries to the db.

The fix works for TAG PR1188 but it should work on the latest (here) as well without any difference since the queries are almost unchanged.

## Background
Basically the problem in both `/users` and `/files` endpoints was that the last subquery that filtered out "inactive" users or files, i.e. that were associated to objects already belonging to a dataset, was going through all the result list. Now, it stops scanning this list as soon as it has a hit.

## How to test
Tests pass, the [deployed hotfix](https://github.com/neicnordic/sensitive-data-archive/pull/1781) works as it should.

### DIFF
--- a/sda/internal/database/db_functions.go
+++ b/sda/internal/database/db_functions.go
@@ -731,7 +731,7 @@ func (dbs *SDAdb) getUserFiles(userID string, allData bool) ([]*SubmissionFileIn
 	// select all files (that are not part of a dataset) of the user, each one annotated with its latest event
 	const query = "SELECT f.id, f.submission_file_path, f.stable_id, e.event, f.created_at FROM sda.files f " +
 		"LEFT JOIN (SELECT DISTINCT ON (file_id) file_id, started_at, event FROM sda.file_event_log ORDER BY file_id, started_at DESC) e ON f.id = e.file_id WHERE f.submission_user = $1 " +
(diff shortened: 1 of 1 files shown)


MalinAhlberg (dismissed):
:star:

sda/internal/database/db_functions.go
```diff
@@ -815,7 +815,7 @@ func (dbs *SDAdb) ListActiveUsers() ([]string, error) {
 	db := dbs.DB
 
 	var users []string
-	rows, err := db.Query("SELECT DISTINCT submission_user FROM sda.files WHERE id NOT IN (SELECT f.id FROM sda.files f RIGHT JOIN sda.file_dataset d ON f.id = d.file_id) ORDER BY submission_user ASC;")
+	rows, err := db.Query("SELECT DISTINCT submission_user FROM sda.files f1 WHERE NOT EXISTS (SELECT 1 FROM sda.file_dataset d WHERE f1.id = d.file_id) ORDER BY submission_user ASC;")
```
jbygdell, line 818:
```suggestion
	rows, err := db.Query("SELECT DISTINCT submission_user FROM sda.files f WHERE NOT EXISTS (SELECT 1 FROM sda.file_dataset d WHERE f.id = d.file_id) ORDER BY submission_user ASC;")
```

jbygdell (changes requested):

jbygdell (approved):

PR #1794, opened 2025-07-08
[ingest] register configured C4GH keys on startup

### LINKED ISSUE #1556: [ingest] Register cryp4gh key in the database on startup

As a repository admin, I want the bootstrapping of a new system to be as simple as possible so that I don't forget to do important steps.

## Acceptance criteria

- [ ] IF the encryption_keys table is empty, insert the configured key

## Estimation of size

small/medium/big

## Estimation of priority

low/medium/high

jbygdell:
## Related issue(s) and PR(s)
This PR closes #1556 .

## Description
If the `encryption_keys` table in the database is empty when ingest starts it will add all configured keys.

### DIFF
--- a/sda/cmd/ingest/ingest.go
+++ b/sda/cmd/ingest/ingest.go
@@ -15,7 +15,6 @@ import (
 	"os/signal"
 	"strings"
 	"syscall"
-	"time"
 
 	"github.com/neicnordic/crypt4gh/keys"
 	"github.com/neicnordic/crypt4gh/model/headers"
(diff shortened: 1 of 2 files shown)


sda/cmd/ingest/ingest.go
```diff
@@ -197,6 +183,23 @@ func main() {
 	<-forever
 }
 
+func (app *Ingest) registerC4GHKey() error {
+	h, err := app.DB.ListKeyHashes()
+	if err != nil {
+		return err
+	}
+	if len(h) == 0 {
```
aaperis, line 191: What if `len(h) != 0` but none of the pukeyhashes in the db corresponds to a key in the `ArchiveKeyList`?

This could be a scenario where we add a key to ingest but forget to register its hash to the db and the db keyhash list is not empty.
Wouldn't it then be possible  to ingesting files with an unknown hashkey?

jbygdell: First off, that is outside the scope of this issue. And secondly there is no difference with regard to that case compared to the current handling.

But, if we want to safeguard against that kind of misconfiguration we first need to implement liveness checks for the non API components. Since we would need some way to ensure that the deployment will fail visibly instead of just having a pod restart continuously like we have now.

aaperis (approved):

PR #1796, opened 2025-07-09
[charts] New major release for sda-svc.

jbygdell:
## Related issue(s) and PR(s)
This PR is related to #1794 In the sense that it simplifies bootstrapping a deployment with all components generated by the charts.

It also allows for an upgrade from a V1 deployment by performing the necessary database schema migrations.

## Description
* Allows for the creation of the secret containing the crypt4gh keys directly from the values file.
* Adds a job that will perform the necessary schema migration from a V1 deployment.
* Adds a job that will set the c4gh key hash in the files table when upgrading from an previous version of the stack.
* Removes outdated parts from the values file and the corresponding sections in the deployment manifests.
* Adds a note to the readme that while it is possible to create the c4gh secret from the values file it is advisable to do it separately.
* Updates the readme with a section about upgrading to this version from a previous one.


pahatz (approved):
Looks good overall, only some cosmetic remarks.

PR #1849, opened 2025-08-13
[api] audit trail information in the logs

zeidlitz:
## Related issue(s) and PR(s)

### DIFF
--- a/sda/cmd/api/api.go
+++ b/sda/cmd/api/api.go
@@ -45,9 +45,10 @@ type dataset struct {
 }
 
 var (
-	Conf *config.Config
-	err  error
-	auth *userauth.ValidateFromToken
+	Conf        *config.Config
+	err         error
+	auth        *userauth.ValidateFromToken
+	auditLogger *log.Logger
 )
 
 func main() {
@@ -234,11 +235,21 @@ func checkDB(db *database.SDAdb, timeout time.Duration) error {
 	return db.DB.PingContext(ctx)
 }
 
+func auditLog(auditFields log.Fields) {
+	auditFields["audit"] = true
+	Conf.API.AuditLogger.WithFields(auditFields).Info("Incoming audit event")
+}
+
 func rbac(e *casbin.Enforcer) gin.HandlerFunc {
 	return func(c *gin.Context) {
 		token, err := auth.Authenticate(c.Request)
 		if err != nil {
-			log.Debugln("bad token")
+			if Conf.API.AuditLogger != nil {
+				auditLog(log.Fields{
+					"authentication error": err.Error(),
+					"path":                 c.Request.URL.Path,
+				})
+			}
 			c.AbortWithStatusJSON(http.StatusUnauthorized, gin.H{"error": err.Error()})
 
 			return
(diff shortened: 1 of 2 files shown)


sda/cmd/api/api.go
```diff
@@ -202,10 +203,26 @@ func checkDB(db *database.SDAdb, timeout time.Duration) error {
...
+	log.WithFields(log.Fields{
+		"audit":  true,
+		"user":   token.Subject(),
+		"method": c.Request.Method,
+		"path":   c.Request.URL.Path,
+		"route":  c.FullPath(),
+		"body":   string(bodyBytes),
+	}).Info("incoming request")
+}
+
```
jbygdell, line 220:
```suggestion
func auditLog(token jwt.Token, c *gin.Context) {
	log.WithFields(log.Fields{
		"audit":  true,
		"user":   token.Subject(),
		"path":   c.Request.URL.Path,
	}).Info("incoming request")
}

```

This should probably be enough

zeidlitz:
**Two questions for review:**

1. Should this be in the rbac middleware or should we create a sepperate middleware that does only this? I think It's fine to keep it in the rbac middleware but I'm not really sure on the "best practice" in this case.

2. Not super familiar with how audit logging is usually treated. How do we set the log level / or logging output?

jbygdell:
> **Two questions for review:**
> 
> 1. Should this be in the rbac middleware or should we create a sepperate middleware that does only this? I think It's fine to keep it in the rbac middleware but I'm not really sure on the "best practice" in this case.

If can definitively be in the rbac middleware, keep in mind that authorization failures might need special handling since information might be missing.

> 2. Not super familiar with how audit logging is usually treated. How do we set the log level / or logging output?

Audit should be logged at a level that is "always" available and it might mean `Info` with this library.

nanjiangshu (dismissed):
Works well. Great job!

viklund:
I'm worried that if it's just logged at `Info` level, it will be easy in the future to turn it off by mistake by changing loglevel in configuration and this implication won't be obvious. I think there should be a specific setting to turn off audit logging.

zeidlitz:
> I'm worried that if it's just logged at `Info` level, it will be easy in the future to turn it off by mistake by changing loglevel in configuration and this implication won't be obvious. I think there should be a specific setting to turn off audit logging.

Sure! I will add a configuration option

sda/cmd/api/api.go
```diff
@@ -234,10 +241,26 @@ func checkDB(db *database.SDAdb, timeout time.Duration) error {
...
+	}
+	if auditLogger != nil {
+		auditLogger.WithFields(log.Fields{
+			"audit": true,
+			"user":  token.Subject(),
+			"path":  c.Request.URL.Path,
+		}).Info("incoming audit event")
+	}
+}
+
```
jbygdell, line 258: If you make the function more generic what should be logged can be decided at the time of calling and you don't needs any if statements in the function.

```suggestion
func auditLog(auditFields log.Fields) {
	auditFields["audit"] =  true
	auditLogger.WithFields(auditFields).Info("incoming audit event")
}

```

Then when it is called it will be like this:

```
if conf.API.auditLogger != nil {
	auditLog(log.Fields{
			"user":  token.Subject(),
			"path":  c.Request.URL.Path,
		})
}
```

And in the case of a JWT error we could log something like this:

```
if conf.API.auditLogger != nil {
	auditLog(log.Fields{
			"authentication error":  err.Error(),
			"path":  c.Request.URL.Path,
		})
}
```

sda/cmd/api/api.go
```diff
@@ -110,6 +112,11 @@ func setup(conf *config.Config) *http.Server {
 		}))
 	}
 
+	auditLogger = log.New()
+	auditLogger.SetFormatter(&log.JSONFormatter{})
+	auditLogger.SetOutput(os.Stdout)
+	auditLogger.SetLevel(log.InfoLevel)
```
jbygdell, line 118: This should probably be part of the API config section.

aaperis:
> Looks good to me.
> 
> One comment: I don't think it's necessary to merge all changes of a PR to one commit. Sometimes it's easier to review when changes are separated into different commits, especially when changes have made after previous approval.

I agree, too many commits is always better than too little :smile:

aaperis:
Is there anywhere I could see the end result of the changes? Like a test or some logs?

zeidlitz:
> Is there anywhere I could see the end result of the changes? Like a test or some logs?

So here you can see an example of the change. It's essentially just one extra log line with some extra fields, and the "audit" : "true" field so we can filter on audit events if wanted. 
[auditlog.log](https://github.com/user-attachments/files/22090106/auditlog.log)

jbygdell (approved):

aaperis (approved):
Looks great, thanks for posting logs :smile:

PR #1851, opened 2025-08-14
[GH Actions] only run the GO linter if the code has been changed

jbygdell:
## Description
This should remove some false errors when the linter triggers on file paths that is not part of the actual PR.


nanjiangshu (approved):
Looks good to me. Hopefully it solves the problem after merging.

PR #1867, opened 2025-08-20
[charts][sda-svc] Parameterize all probes

jbygdell:
## Description
This allows for better control when dealing with sluggish network connections for external services.
It also allows for the administrator to determine which probes to run for a given service.

### DIFF
--- a/charts/sda-svc/values.yaml
+++ b/charts/sda-svc/values.yaml
@@ -373,6 +373,21 @@ credentials:
 api:
   name: api
   replicaCount: 2
+  livenessProbe:
+    # httpGet:
+    #   port: api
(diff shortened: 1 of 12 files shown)


charts/sda-svc/values.yaml
```diff
@@ -373,6 +373,21 @@ credentials:
 api:
   name: api
   replicaCount: 2
+  livenessProbe:
+    # httpGet:
```
zeidlitz, line 377: Do we want to keep these as comments?

zeidlitz (approved):
LGTM

nanjiangshu (commented):
Looks good.

PR #1877, opened 2025-08-21
Pull mock_lsaai image from nbis harbor

aaperis:


nanjiangshu (approved):
Works well.

aaperis:
Upon request, there's a new release of the mock_lsaai that works well. So, problem solved :smile: .

jbygdell (approved):
Squash all commits into one before merging.

PR #1886, opened 2025-08-25
[API] Extend swagger file

### LINKED ISSUE #1884: [admin-api] Design API and document in Swagger the definition for the list files in a dataset API extension
Extend the swagger with a parameter for only specifying a specific dataset when listing files from a user.

jbygdell:
## Related issue(s) and PR(s)
This PR closes #1884.

## Description
Extend the swagger file with the query parameter for the file listing endpoint. Keeping in mind that  this addition should be project agnostic.

### DIFF
--- a/sda/cmd/api/swagger_v1.yml
+++ b/sda/cmd/api/swagger_v1.yml
@@ -300,6 +300,12 @@ paths:
           schema:
             type: string
           required: true
+        - in: query
+          name: path_prefix
+          schema:
+            type: string
+          required: false
+          description: The inbox path prefix to filter entries on.
       responses:
         "200":
           content:


sda/cmd/api/swagger_v1.yml
```diff
@@ -300,6 +300,12 @@ paths:
           schema:
             type: string
           required: true
+        - in: query
+          name: path_prefix
```
nanjiangshu, line 304: Should it only be the prefix or can be a substring at any position? I don't think there's big difference in the database query when using `LIKE`.

If we stick to the `prefix`, then the userID (still exist in the filepath) should not be included.

jbygdell: That will depend on how you implement the query.
`submission_file_path LIKE '%PREFIX%'` or `submission_file_path LIKE 'PREFIX%'`
The first variant allows you to select *any* substring while that latter is strict for the first part of the path.

nanjiangshu: Sure, you can implement like `submission_file_path LIKE '%PREFIX%'` and in this case `PREFIX` works as a substring. However, if one just read the API specification in Swagger, the name `path_prefix` naturally indicate matching the beginning of the file path, which is misleading. I would suggest change 
`path_prefix` to a less misleading name, e.g. `path_substring` or `path_part`.

nanjiangshu (approved):

PR #1918, opened 2025-09-01
Rotate key service

### LINKED ISSUE #1673: [reencrypt] Service for rotating key of one file
`rotate-key` service receives a message to re-encrypt one file.

It is configured with the public key of the target of the rotation.

 1. Check that the file is not already re-encrypted with the target key (key-hash field).
 2. Re-encrypt header with the help of re-encrypt service
 3. Store the header in the files table together with the key-hash of the new key
 4. Note "modified by" (or something in the file status table) that there was an update by the re-encrypt service.
 5. Send a message to verify so the file is re-verified.

aaperis:
## Related issue(s) and PR(s)
This PR closes #1673 .

## Description
WIP:
- use reencrypt service
- unittests
- integration tests


sda/cmd/rotatekey/rotatekey.md
```diff
@@ -0,0 +1,111 @@
+# rotatekey Service
+
+Rotates the crypt4gh encryption key of a file that is mapped to a dataset and stored in the SDA.
+
+## Service Description
+
+The rotatekey service re-encrypts the header of a file with the configured target key, and updates the database with the new header and encryption key hash.
+
+When running, rotatekey reads messages from the `mapping_stream` RabbitMQ queue.
```
KarlG-nbis, line 9: is it the `mapping_stream` or the `rotatekey_stream` https://github.com/neicnordic/sensitive-data-archive/pull/1918/files#diff-19c8132437a099a64900a4e4e0f5a2f7aef5a8b7c6d43b027025d8c29c2c1698R25

sda/cmd/rotatekey/rotatekey.md
```diff
@@ -0,0 +1,111 @@
...
+For each message, these steps are taken (errors halts progress, the message is Nack'ed, an info-error message is sent and the service moves on to the next message):
+
+1. The message is validated as valid JSON that matches the "dataset-mapping" schema.
+2. A database look-up is performed for the configured target public key hash. If the look-up fails or the key has been deprecated, an error is raised.
+3. If the massage contains more than one accession IDs, an error is raised and the message is discarded as the service works on a one file per message basis.
+4. The file ID is fetched from the database.
+5. The key hash of the c4gh key with which the file is currently encrypted is fetched from the database and compared with the configured target key.
+6. If these key hashes differ, the reencrypt service is called to re-encrypt the file header with the target key.
+7. The file header entry in the database is updated with the new one.
```
KarlG-nbis, line 18: Would we not want to capture that a key for a file has been rotated in the file_event_log?

KarlG-nbis: Or would that be inferred by the re verification by the verify service?

aaperis: Good point, I should have probably mentioned in the PR description that  the field `last_modified_by` in the files table will change to `rotatekey` after  the key for a specific file has been rotated.

As the stack is at the moment we have some events driven by file status in the file_event_log so having the last status as reencrypted or similar might require some more thought. It does make sense to me though to capture this at the file_event_log so maybe we should have a discussion about this table at some point.

postgresql/migratedb.d/14.sql
```diff
@@ -41,6 +41,9 @@ BEGIN
     GRANT SELECT, INSERT, UPDATE ON sda.userinfo TO auth;
 
     GRANT base TO auth;
+
+    -- Drop temporary user creation function
+    DROP FUNCTION create_role_if_not_exists;
```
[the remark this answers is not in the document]
KarlG-nbis: Ah ok, yeah didnt look to closely at what was added in version 18, but if we are creating a function with the same name again, and in the env where we run migration the dropping of that function has not been ran, wouldnt it fail? But wouldnt matter much since the function is the same, unless if the execution of 18 stops and the other queries are not executed after the function creation

jbygdell (changes requested):

rabbitmq/definitions.json
```diff
@@ -257,6 +289,14 @@
             "destination": "verified",
             "routing_key": "verified"
         },
+        {
+            "source": "sda",
+            "vhost": "sda",
+            "destination_type": "queue",
+            "arguments": {},
+            "destination": "rotatekey_stream",
```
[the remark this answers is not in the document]
aaperis: I thought we wanted persistence for messages according to the diagrams for the app?

sda/cmd/rotatekey/rotatekey.go
```diff
@@ -0,0 +1,322 @@
...
+				logAndNack(mq, delivered, msg, err)
+
+				continue
+			}
+
+			// Check that the file is not already encrypted with the target key
+			if oldKeyHash == keyhash {
+				msg := fmt.Sprintf("failed to reencrypt file with file-id: %s", fileID)
+				err = errors.New("already encrypted with the given rotation c4gh key")
+				logAndNack(mq, delivered, msg, err)
```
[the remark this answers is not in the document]
aaperis: indeed

sda/internal/config/config.go
```diff
@@ -1197,8 +1234,17 @@ func GetC4GHprivateKeys() ([]*[32]byte, error) {
...
-	keyPath := viper.GetString("c4gh.syncPubKeyPath")
+func GetC4GHPublicKey(app string) (*[32]byte, error) {
+	var keyPath string
+	switch app {
+	case "sync":
+		keyPath = viper.GetString("c4gh.syncPubKeyPath")
+	case "rotatekey":
+		keyPath = viper.GetString("c4gh.rotatePubKeyPath")
+	default:
+		return nil, errors.New("pubKey not set")
```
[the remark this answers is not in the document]
aaperis: > keep in mind her that the public key used by reencrypt will be the same one that `auth` shall use

How does this fit here? From what I see the function `GetC4GHPublicKey` is only called by sync and rotatekey atm.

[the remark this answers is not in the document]
jbygdell: Because of bad naming for the stuff that auth uses and that auth has a similar function for its own use.

In auths case it is called `publicFile` which is a bit to generic and that it is not nested under c4gh part in the config.

.github/integration/scripts/make_sda_credentials.sh
```diff
@@ -121,6 +121,22 @@ if [ ! -f "/shared/sync.sec.pem" ]; then
...
 
+if [ ! -f "/shared/rotatekey.sec.pem" ]; then
+    echo "creating rotatekey crypth4gh key"
+    /shared/crypt4gh generate -n /shared/rotatekey -p rotatekeyPass
+fi
+
+# register the rotation key in the db
+resp=$(psql -U postgres -h postgres -d sda -At -c "SELECT description FROM sda.encryption_keys;")
+if ! echo "$resp" | grep -q 'this is the new key to rotate to'; then
+    rotateKeyHash=$(cat /shared/rotatekey.pub.pem | awk 'NR==2' | base64 -d | xxd -p -c256)
```
nanjiangshu, line 132:
```suggestion
    rotateKeyHash=$(awk 'NR>1 && !/-----/' /shared/rotatekey.pub.pem | base64 -d | xxd -p -c256 | tr -d '\n')
```

More robust parsing of the key
1. In case base64 content spans for multiple lines 
2. In case the output of `xxd -p -c256` may truncated or reformatted unexpectedly.

.github/integration/tests/sda/70_rotate_key_test.sh
```diff
@@ -0,0 +1,239 @@
...
+        '$ARGS.named'
+)
+
+curl -s -u guest:guest "http://rabbitmq:15672/api/exchanges/sda/sda/publish" \
+    -H 'Content-Type: application/json;charset=UTF-8' \
+    -d "$mapping_body" | jq
+
+# check DB for updated key hash in sda.files
+rotatekeyHash=$(psql -U postgres -h postgres -d sda -At -c "select key_hash from sda.encryption_keys where description='this is the rotatekey key';")
+if "$(psql -U postgres -h postgres -d sda -At -c "select key_hash from sda.files where stable_id like 'ROTATE-KEY-0%';" | grep -c "$rotatekeyHash")" -neq 1;
```
nanjiangshu, line 130: `-neq` is not a valid keyword. Maybe you want to do something like this?

```shell
if [ "$(psql -U postgres -h postgres -d sda -At -c \
    "select key_hash from sda.files where stable_id like 'ROTATE-KEY-0%';" \
    | grep -c "$rotatekeyHash")" -ne 1 ]; then
    echo "failed to update the key hash of files"
    exit 1
fi
```

.github/integration/tests/sda/70_rotate_key_test.sh
```diff
@@ -0,0 +1,239 @@
+#!/bin/sh
+set -e
```
nanjiangshu, line 2:
```suggestion
set -eo pipefail
```
With this setting, it can also catch errors in pipelines, e.g. `curl | jq`

postgresql/migratedb.d/14.sql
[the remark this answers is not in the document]
nanjiangshu: I agree with @aaperis

sda/cmd/rotatekey/rotatekey.md
```diff
@@ -0,0 +1,111 @@
+# rotatekey Service
+
+Rotates the crypt4gh encryption key of a file that is mapped to a dataset and stored in the SDA.
+
+## Service Description
+
+The rotatekey service re-encrypts the header of a file with the configured target key, and updates the database with the new header and encryption key hash.
+
+When running, rotatekey reads messages from the `rotatekey_stream` RabbitMQ queue.
+For each message, these steps are taken (errors halts progress, the message is Nack'ed, an info-error message is sent and the service moves on to the next message):
```
nanjiangshu, line 10: Not very clear to me. Do the steps mean those included in the parenthesis or listed in the bullet points below?

sda/cmd/rotatekey/rotatekey.md
```diff
@@ -0,0 +1,111 @@
...
+
+1. The message is validated as valid JSON that matches the "dataset-mapping" schema.
+2. A database look-up is performed for the configured target public key hash. If the look-up fails or the key has been deprecated, an error is raised.
+3. If the massage contains more than one accession IDs, an error is raised and the message is discarded as the service works on a one file per message basis.
+4. The file ID is fetched from the database.
+5. The key hash of the c4gh key with which the file is currently encrypted is fetched from the database and compared with the configured target key.
+6. If these key hashes differ, the reencrypt service is called to re-encrypt the file header with the target key.
+7. The file header entry in the database is updated with the new one.
+8. The key hash entry for the database is updated with the new one (target key).
+9. A re-verify message is compiled, validated and sent to the archived queue so that it is consumed by verify service.
```
nanjiangshu, line 20:
```suggestion
9. A re-verify message is compiled, validated and sent to the archived queue so that it is consumed by the `verify` service.
```

sda/cmd/rotatekey/rotatekey.md
```diff
@@ -0,0 +1,111 @@
...
+  - `verify-ca`
+  - `verify-full`
+
+  More information is available
+  [in the postgresql documentation](https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-PROTECTION)
+
+  Note that if `DB_SSLMODE` is set to anything but `disable`, then `DB_CACERT` needs to be set,
+  and if set to `verify-full`, then `DB_CLIENTCERT`, and `DB_CLIENTKEY` must also be set.
+
+- `DB_CLIENTKEY`: key-file for the database client certificate
```
nanjiangshu, line 88:
```suggestion
- `DB_CLIENTKEY`: key file for the database client certificate
```
`key-file` looks like a special term or variable used in the system

sda/internal/config/config.go
```diff
@@ -389,6 +394,21 @@ func NewConfig(app string) (*Config, error) {
...
 			"c4gh.privateKeys",
 		}
+	case "rotatekey":
+		requiredConfVars = []string{
+			"broker.host",
+			"broker.port",
+			"broker.user",
+			"broker.password",
+			"broker.queue",
+			"c4gh.rotatePubKeyPath",
```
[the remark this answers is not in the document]
nanjiangshu: I also think `activePubKeyPath` is better, as it refers to the key that will be active and remain active until a new key is introduced.

sda/internal/config/config.go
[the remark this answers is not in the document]
jbygdell: This should probably be generalized by having the file path as an input variable. Then it doesn't matter which app is calling it.

.github/integration/tests/sda/70_rotate_key_test.sh
```diff
@@ -0,0 +1,239 @@
...
+	echo "something went wrong with the key rotation"
+	exit 1
+fi
+
+## download file with rotated key, concatenate header and archive body, decrypt and check
+# get rotated header
+psql -U postgres -h postgres -d sda -At -c "select header from sda.files where stable_id='ROTATE-KEY-01';" | xxd -r -p > testfile1_rotated.c4gh
+# get archive file
+archivePath=$(psql -U postgres -h postgres -d sda -At -c "select archive_file_path from sda.files where stable_id='ROTATE-KEY-01';")
+s3cmd --access_key=access --secret_key=secretKey --host=minio:9000 --no-ssl --host-bucket=minio:9000 get s3://archive/"$archivePath" --force
```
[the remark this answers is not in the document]
aaperis: minio server is not tls secured

.github/integration/tests/sda/70_rotate_key_test.sh
aaperis: not posix compliant, see [here](https://www.shellcheck.net/wiki/SC3040)

sda/cmd/rotatekey/rotatekey.go
```diff
@@ -0,0 +1,322 @@
...
+	defer cancel()
+
+	c := reencrypt.NewReencryptClient(conn)
+	res, err := c.ReencryptHeader(ctx, &reencrypt.ReencryptRequest{Oldheader: oldHeader, Publickey: c4ghPubKey})
+	if err != nil {
+		return nil, err
+	}
+
+	return res.Header, nil
+}
```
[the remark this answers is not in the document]
jbygdell: The removal of this block from the API service should be  a separate PR.
BUt creating the functionality in the reencrypt package could be done here.

sda/cmd/rotatekey/rotatekey.go
```diff
@@ -0,0 +1,322 @@
...
+	}()
+
+	go func() {
+		connError := mq.ChannelWatcher()
+		log.Error(connError)
+		forever <- false
+	}()
+
+	log.Info("Starting rotatekey service")
+	var message schema.DatasetMapping
```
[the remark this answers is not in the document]
jbygdell: Given that these actions relate to files already in the system we should use the correlation IDs that already exists.
So go for scenario 1.

sda/cmd/rotatekey/rotatekey.go
```diff
@@ -0,0 +1,322 @@
...
+	newHeader, err := reencryptHeader(header, pubKeyEncoded)
+	if err != nil {
+		return nil, err
+	}
+
+	return newHeader, nil
+}
+
+// Check that the key hash exists in the database
+func getKeyHash() (string, error) {
```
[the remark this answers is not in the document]
aaperis: Done. It made more sense to split this into a db function (c19d1a4) and use it properly (a80dd5f).

sda/cmd/rotatekey/rotatekey.go
```diff
@@ -0,0 +1,322 @@
...
+	}
+	body, _ := json.Marshal(infoErrorMessage)
+
+	if err := mq.SendMessage(delivered.CorrelationId, Conf.Broker.Exchange, "error", body); err != nil {
+		log.Errorf("failed to publish message, reason: (%s)", err.Error())
+	}
+	if err := delivered.Ack(false); err != nil {
+		log.Errorf("failed to Ack message, reason: (%s)", err.Error())
+	}
+}
```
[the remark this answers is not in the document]
aaperis: Fixed in 2f54ad8c19174464e531a16d1e3dcbca5ec0ba02 .

sda/cmd/rotatekey/rotatekey.go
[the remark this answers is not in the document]
aaperis: New schema rotate-key added.

sda/cmd/rotatekey/rotatekey.go
```diff
@@ -0,0 +1,322 @@
...
+				continue
+			}
+
+			corrID, err := db.GetCorrID(reVerify.User, reVerify.FilePath, aID)
+			if err != nil {
+				msg := fmt.Sprintf("failed to get CorrID for %s, %s", reVerify.User, reVerify.FilePath)
+				logAndNack(mq, delivered, msg, err)
+
+				continue
+			}
```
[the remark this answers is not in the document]
aaperis: Done

sda/cmd/rotatekey/rotatekey.go
```diff
@@ -0,0 +1,322 @@
...
+
+			break
+		}
+	}
+	if !found {
+		return "", errors.New("the c4gh key hash is not registered")
+	}
+
+	return keyhash, nil
+}
```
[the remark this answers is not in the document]
aaperis: done here https://github.com/neicnordic/sensitive-data-archive/pull/1918/commits/abc92f4e1696e4ea6b4a7672748f1ef95da037e0

sda/internal/config/config.go
```diff
@@ -70,6 +71,10 @@ type ReEncConfig struct {
 	Timeout            int
 }
 
+type RotateKeyConf struct {
+	Grpc Grpc
+}
```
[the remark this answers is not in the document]
aaperis: After the discussion on how we deal with the pubk key config in the future, this became obsolete.

.github/integration/scripts/make_sda_credentials.sh
nanjiangshu: I over considered the generality. You are right, no need to change if we stick to c4gh.

sda/cmd/sync/sync.go
```diff
@@ -60,10 +60,10 @@ func main() {
...
 	}
 
-	publicKey, err = config.GetC4GHPublicKey()
-	if err != nil {
-		log.Fatal(err)
-	}
+	// publicKey, err = config.GetC4GHPublicKey("sync")
+	// if err != nil {
+	// 	log.Fatal(err)
+	// }
```
jbygdell, line 66: delete or uncomment

sda/cmd/rotatekey/rotatekey.go
```diff
@@ -0,0 +1,235 @@
...
+				continue
+			}
+
+			// Fetch rotate key hash before starting work so that we make sure the hash state
+			// has not changed since the application startup.
+			keyhash := hex.EncodeToString(conf.RotateKey.PublicKey[:])
+			err = db.CheckKeyHash(keyhash)
+			// exit app if target key was modified after app start-up, e.g. if key has been deprecated
+			if err != nil {
+				log.Fatalf("check of target key failed, reason: %v", err)
```
jbygdell, line 99:
```suggestion
				log.Error(err)
				sigc <- syscall.SIGINT
				panic(err)
```

sda/cmd/rotatekey/rotatekey.go
```diff
@@ -0,0 +1,235 @@
...
+				log.Errorf("failed to Ack message, reason: (%s)", err.Error())
+			}
+		}
+	}()
+
+	<-forever
+}
+
+// Nack message without requeue. Send the message to an error queue so it can be analyzed.
+func NackAndSendToErrorQueue(mq *broker.AMQPBroker, delivered amqp091.Delivery, mqExchange, msg, reason string) {
```
jbygdell, line 221:
```suggestion
// Ack message and send the payload to an error queue so it can be analyzed.
func AckAndSendToErrorQueue(mq *broker.AMQPBroker, delivered amqp091.Delivery, mqExchange, msg, reason string) {
```

Update the function name to reflect what's actually happening.

sda/cmd/rotatekey/rotatekey.go
```diff
@@ -0,0 +1,235 @@
...
+
+				continue
+			}
+
+			// Fetch rotate key hash before starting work so that we make sure the hash state
+			// has not changed since the application startup.
+			keyhash := hex.EncodeToString(conf.RotateKey.PublicKey[:])
+			err = db.CheckKeyHash(keyhash)
+			// exit app if target key was modified after app start-up, e.g. if key has been deprecated
+			if err != nil {
```
jbygdell, line 98:
```suggestion
			// exit app if target key was modified after app start-up, e.g. if key has been deprecated
			if err := db.CheckKeyHash(keyhash); err != nil {
```

sda/cmd/rotatekey/rotatekey.go
```diff
@@ -0,0 +1,235 @@
...
+				continue
+			}
+			if newHeader == nil {
+				err := errors.New("reencrypt returned empty header")
+				msg := fmt.Sprintf("failed to rotate c4gh key for file %s", fileID)
+				log.Errorf("%s, reason: %v", msg, err)
+				NackAndSendToErrorQueue(mq, delivered, conf.Broker.Exchange, msg, err.Error())
+
+				continue
+			}
```
jbygdell, line 154: Is this really a possibility that the header is nil and no error?

sda/cmd/rotatekey/rotatekey.go
```diff
@@ -0,0 +1,235 @@
...
+
+			// Rotate keyhash
+			if err := db.SetKeyHash(keyhash, fileID); err != nil {
+				msg := fmt.Sprintf("SetKeyHash failed for file-id: %s", fileID)
+				log.Errorf("%s, reason: %v", msg, err)
+				NackAndSendToErrorQueue(mq, delivered, conf.Broker.Exchange, msg, err.Error())
+
+				continue
+			}
```
jbygdell, line 172: If this fails we are in big trouble since the new header is already in the DB but the hash is wrong.

It's probably better to create a new DB call that does both at the same time.

sda/cmd/rotatekey/rotatekey.md
```diff
@@ -0,0 +1,111 @@
+# rotatekey Service
+
+Rotates the crypt4gh encryption key of ingested files.
```
jbygdell, line 3:
```suggestion
Rotates the crypt4gh encryption key of ingested file headers.
```
We are not reencrypting the entire file.

sda/internal/reencrypt/reencrypt_utils.go
```diff
@@ -0,0 +1,48 @@
...
+	c := NewReencryptClient(conn)
+	res, err := c.ReencryptHeader(ctx, &ReencryptRequest{Oldheader: oldHeader, Publickey: c4ghPubKey})
+	if err != nil {
+		log.Errorf("failed to connect to the reencrypt service, reason %v", err)
+
+		return nil, err
+	}
+
+	return res.Header, nil
+}
```
jbygdell, line 48: This should have a tests

sda/schemas/federated/rotate-key.json
```diff
@@ -0,0 +1,29 @@
...
+            "title": "The message type",
+            "description": "The message type",
+            "const": "key_rotation"
+        },
+        "file_id": {
+            "$id": "#/properties/file_id",
+            "type": "string",
+            "title": "The unique file identifier",
+            "description": "The unique file identifier",
+            "examples": [
```
jbygdell, line 24:
```suggestion
            "description": "The unique file identifier",
            "pattern": "^[a-f0-9]{10}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$",
            "examples": [
```

sda/schemas/isolated/rotate-key.json
```diff
@@ -0,0 +1,29 @@
...
+            "title": "The message type",
+            "description": "The message type",
+            "const": "key_rotation"
+        },
+        "file_id": {
+            "$id": "#/properties/file_id",
+            "type": "string",
+            "title": "The unique file identifier",
+            "description": "The unique file identifier",
+            "examples": [
```
jbygdell, line 24:
```suggestion
            "description": "The unique file identifier",
           "pattern": "^[a-f0-9]{10}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$",
            "examples": [
```

sda/schemas/isolated/rotate-key.json
```diff
@@ -0,0 +1,29 @@
...
+                "420420cc43-e060-4583-a891-9f8170ee66c8"
+            ]
+        }
+    }
+}
```
jbygdell, line 29: Since this file is identical to the other we can symlink them so we only have to edit this in one place.

jbygdell (changes requested):

.github/integration/scripts/make_sda_credentials.sh
```diff
@@ -121,6 +116,22 @@ if [ ! -f "/shared/sync.sec.pem" ]; then
...
+    resp=$(psql -U postgres -h postgres -d sda -At -c "INSERT INTO sda.encryption_keys(key_hash, description) VALUES('$rotateKeyHash', 'this is the new key to rotate to');")
+    if [ "$(echo "$resp" | tr -d '\n')" != "INSERT 0 1" ]; then
+        echo "insert keyhash failed"
+        exit 1
+    fi
+fi
```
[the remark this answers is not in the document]
aaperis: The app won't start without a registered key.

.github/integration/tests/sda/70_rotate_key_test.sh
```diff
@@ -0,0 +1,266 @@
+#!/bin/sh
+set -e
+
+if [ -n "$SYNCTEST" ]; then
+    exit 0
+fi
+
```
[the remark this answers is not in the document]
jbygdell: How come?
All other tests are running together with sync.

sda/cmd/rotatekey/rotatekey.go
aaperis: Yes, I actually encountered this at some point but I can't remember exact details.

sda/cmd/rotatekey/rotatekey.go
```diff
@@ -0,0 +1,235 @@
...
+	"github.com/neicnordic/sensitive-data-archive/internal/config"
+	"github.com/neicnordic/sensitive-data-archive/internal/database"
+	"github.com/neicnordic/sensitive-data-archive/internal/reencrypt"
+	"github.com/neicnordic/sensitive-data-archive/internal/schema"
+	"github.com/rabbitmq/amqp091-go"
+	log "github.com/sirupsen/logrus"
+)
+
+func main() {
+	forever := make(chan bool)
```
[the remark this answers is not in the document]
aaperis: Done.

.github/integration/tests/sda/70_rotate_key_test.sh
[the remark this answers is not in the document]
aaperis: We don't need to run this test twice. If other tests are not needed for testing `sync` then we shouldn't run them either.

.github/integration/scripts/make_sda_credentials.sh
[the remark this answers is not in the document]
aaperis: I prefer not to register keys automatically.

postgresql/migratedb.d/14.sql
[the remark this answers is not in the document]
aaperis: Although this is true, I find it more tidy to drop the function it when done the migration step.

jbygdell (dismissed):

sda/cmd/rotatekey/rotatekey.go
```diff
@@ -0,0 +1,268 @@
...
+			_ = json.Unmarshal(delivered.Body, &message)
+
+			fileID := message.FileID
+
+			// Get current keyhash for the file, send to error queue if this fails
+			oldKeyHash, err := db.GetKeyHash(fileID)
+			if err != nil {
+				msg := fmt.Sprintf("failed to get keyhash for file with file-id: %s", fileID)
+				log.Errorf("%s, reason: %v", msg, err)
+				nackAndSendToErrorQueueckAndSendToErrorQueue(mq, delivered, conf.Broker.Exchange, msg, err.Error())
```
jbygdell, line 154: This should have separate handlings dependent on the error.
type 1 errors: nack and requeue the message so we can retry later.
type 2 errors: ack and send to error queue

sda/cmd/rotatekey/rotatekey.go
```diff
@@ -0,0 +1,268 @@
...
+			log.Debugf("rotating c4gh key for file with file-id: %s", fileID)
+
+			header, err := db.GetHeader(fileID)
+			if err != nil {
+				msg := fmt.Sprintf("GetHeader failed for file-id: %s", fileID)
+				log.Errorf("%s, reason: %v", msg, err)
+				nackAndSendToErrorQueueckAndSendToErrorQueue(mq, delivered, conf.Broker.Exchange, msg, err.Error())
+
+				continue
+			}
```
jbygdell, line 179: same here we need to handle 2 different cases.

sda/cmd/rotatekey/rotatekey.go
```diff
@@ -0,0 +1,268 @@
...
+			if newHeader == nil {
+				err := errors.New("reencrypt returned empty header")
+				msg := fmt.Sprintf("failed to rotate c4gh key for file %s", fileID)
+				log.Errorf("%s, reason: %v", msg, err)
+				nackAndSendToErrorQueueckAndSendToErrorQueue(mq, delivered, conf.Broker.Exchange, msg, err.Error())
+
+				continue
+			}
```
jbygdell, line 196: can this really happen without an error being generated?

sda/cmd/rotatekey/rotatekey.go
```diff
@@ -0,0 +1,268 @@
...
+
+			// Rotate header and keyhash in database
+			if err := db.RotateHeaderKey(newHeader, keyhash, fileID); err != nil {
+				msg := fmt.Sprintf("RotateHeaderKey failed for file-id: %s", fileID)
+				log.Errorf("%s, reason: %v", msg, err)
+				nackAndSendToErrorQueueckAndSendToErrorQueue(mq, delivered, conf.Broker.Exchange, msg, err.Error())
+
+				continue
+			}
```
jbygdell, line 205: This should result in a requeueing of the message.

sda/cmd/rotatekey/rotatekey.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	body, _ := json.Marshal(infoErrorMessage)
+
+	if err := mq.SendMessage(delivered.CorrelationId, mqExchange, "error", body); err != nil {
+		log.Errorf("failed to publish message, reason: (%s)", err.Error())
+	}
+	if err := delivered.Nack(false, false); err != nil {
+		log.Errorf("failed to Ack message, reason: (%s)", err.Error())
+	}
+}
```
jbygdell, line 268: Just because you get an error it doesn't mean that the message should be `nacked`. Look at the logic from within the MQTT and you will see how the different cases work.

* `nack/reject with requeue=1`: the message will be returned to the queue it came from as though it were a new message; this might be useful in case of a temporary failure on the consumer side
* `nack/reject with requeue=0 and a configured Dead Letter Exchange (DLX)`:  publish the message to that exchange, allowing it to be picked up by another queue
* `nack/reject with requeue=0 and no DLX`: simply discard the message
* `ack` will remove the message from the queue even if a DLX is configured

sda/cmd/rotatekey/rotatekey.go
aaperis: The db function both for this and the above have a backoff strategy of tenths of seconds. It should be enough so that we can safely assume there is an actual error here. If we requeue, it may as well lead to infinite loops like we've seen in other services of the stack in recent past.

sda/cmd/rotatekey/rotatekey.go
jbygdell: Not really.
An error in that DB function means that the header and key reference is unchanged and as a result the work needs to be performed again and thus the message should be requeued.

sda/cmd/rotatekey/rotatekey.go
jbygdell: That logic is flawed. The ack/nack is about how the MQTT should deal with the message not if the code generated an error or not.

sda/cmd/rotatekey/rotatekey.go
jbygdell: These 3 scenarios for the reencrypt service call all generates an error:
* bad/missing public key
* bad/missing header
* wrong/missing private key config

sda/cmd/rotatekey/rotatekey.go
aaperis: ...which would both lead to empty rows.

zeidlitz (approved):

Issue #1832, opened 2025-08-11
[sda-download] aws-sdk is deprecated and needs to be upgraded


jbygdell:
## Describe the bug
aws-sdk is deprecated

## Expected behavior
- [ ] upgrade to aws-sdk-v2

## Estimation of size
/medium

## Estimation of priority
high

jbygdell:
Download V1 is being deprecated

PR #2047, opened 2025-10-23
[sda-download] able to block non supported client versions

### LINKED ISSUE #2038: [download] Ability to block older versions of sda-cli
Implement it by checking if a header called `SDA-Client-Version: vX.Y.Z` and checking that it's the correct version, also block if it's not present.

Return http code 412. Also make sure that the user that is using sda-cli gets an error message so they understand that they should download a newer version of the client.

It should be on all of sda-downloads API endpoints.

nanjiangshu:
## Related issue(s) and PR(s)
This PR closes #2038 

## How to test
The integration tests for sda-download pass

## Additional information
I've fixed some of the old linting errors and the remaining linting errors caused by deprecated aws-sdk-go will be fixed in the issue [#1832](https://github.com/neicnordic/sensitive-data-archive/issues/1832)

### DIFF
--- a/sda-download/api/middleware/middleware.go
+++ b/sda-download/api/middleware/middleware.go
@@ -96,7 +96,7 @@ func ClientVersionMiddleware() gin.HandlerFunc {
 
 		// Check if the header is present
 		if clientVersionStr == "" {
-			log.Warnf("request blocked (412): Missing client version header in request")
+			log.Debugln("request blocked (412): Missing client version header in request")
 			c.String(http.StatusPreconditionFailed, "Missing client version header in request")
 			c.Abort()
 
@@ -106,7 +106,7 @@ func ClientVersionMiddleware() gin.HandlerFunc {
 		// Parse the client's provided version (using the processed string)
 		clientVersion, err := semver.NewVersion(clientVersionStr)
 		if err != nil {
-			log.Warnf("client version header '%s' is not a valid semantic version: %v", clientVersionStr, err)
+			log.Debugf("client version header '%s' is not a valid semantic version: %v", clientVersionStr, err)
 			c.String(http.StatusPreconditionFailed, "client version header is not a valid semantic version")
 			c.Abort()
 
@@ -116,7 +116,7 @@ func ClientVersionMiddleware() gin.HandlerFunc {
 		// Check if the client version is sufficient (clientVersion >= minimalVersion)
 		if clientVersion.LessThan(config.Config.App.MinimalCliVersion) {
 			errorMessage := fmt.Sprintf("Error: Your sda-cli client version is outdated, please update to at least version '%s'.", config.Config.App.MinimalCliVersionStr)
-			log.Warnf("request blocked (412): outdated client version '%s'. Required minimum '%s'", clientVersionStr, config.Config.App.MinimalCliVersionStr)
+			log.Debugf("request blocked (412): outdated client version '%s'. Required minimum '%s'", clientVersionStr, config.Config.App.MinimalCliVersionStr)
 			c.String(http.StatusPreconditionFailed, errorMessage)
 			c.Abort()
 


sda-download/internal/config/config.go
```diff
@@ -242,6 +249,7 @@ func (c *Map) applyDefaults() {
 	viper.SetDefault("app.host", "0.0.0.0")
 	viper.SetDefault("app.port", 8080)
 	viper.SetDefault("app.middleware", "default")
+	viper.SetDefault("app.expectedcliversion", "v0.0.0")
```
jbygdell:
```suggestion
	viper.SetDefault("app.minimalcliversion", "v0.0.0")
```

nanjiangshu:
Thanks for your @jbygdell comments, I think all of them should have been addressed now. Please take a look when you have time.

sda-download/api/middleware/middleware.go
```diff
@@ -79,8 +81,83 @@ func TokenMiddleware() gin.HandlerFunc {
...
+
+		// Version is correct, proceed to the next handler/middleware
+		log.Debugf("client version check passed: %s", clientVersionStr)
+	}
+}
+
+// ChainDefaultMiddleware chains the TokenMiddleware and ClientVersionMiddleware,
+// prioritizing authentication before checking client version requirements.
+// It is intended to be the default composite middleware set for the application.
+func ChainDefaultMiddleware() gin.HandlerFunc {
```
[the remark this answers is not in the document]
nanjiangshu: It's a good point. The current setting makes the it a bit awkward if we DO want to configure with a different selectedmiddleware. But it think this issues is better to be addressed in a refactor issue.

sda-download/api/api.go
```diff
@@ -65,12 +71,16 @@ func Setup() *http.Server {
...
 
 	router.HandleMethodNotAllowed = true
+	mw := SelectedMiddleware()
+
+	// protected endpoints
+	router.GET("/metadata/datasets", append(mw, sda.Datasets)...)
+	router.GET("/metadata/datasets/*dataset", append(mw, sda.Files)...)
+	router.GET("/files/:fileid", append(mw, sda.Download)...)
+	router.GET("/s3/*path", append(mw, s3.Download)...)
+	router.HEAD("/s3/*path", append(mw, s3.Download)...)
```
jbygdell:
```suggestion
	// protected endpoints
	router.GET("/metadata/datasets", append(SelectedMiddleware(), sda.Datasets)...)
	router.GET("/metadata/datasets/*dataset", append(SelectedMiddleware(), sda.Files)...)
	router.GET("/files/:fileid", append(SelectedMiddleware(), sda.Download)...)
	router.GET("/s3/*path", append(SelectedMiddleware(), s3.Download)...)
	router.HEAD("/s3/*path", append(SelectedMiddleware(), s3.Download)...)
```

Since the SelectedMiddleware function doesn't return any errors there is no need to generate a variable for this section.

sda-download/api/api.go
```diff
@@ -9,18 +9,24 @@ import (
...
-// available middlewares:
-// "default" for TokenMiddleware
-var SelectedMiddleware = func() gin.HandlerFunc {
-	return nil
+// SelectedMiddleware returns the middleware chain based on configuration.
+// For example, config.Config.App.Middleware could be "default", "token", etc.
+var SelectedMiddleware = func() []gin.HandlerFunc {
+	switch strings.ToLower(config.Config.App.Middleware) {
+	case "default":
+		return middleware.ChainDefaultMiddleware()
```
jbygdell:
```suggestion
		return []gin.HandlerFunc{
		middleware.TokenMiddleware(),
		middleware.ClientVersionMiddleware(),
	}
```

There is not need to hide this behind a separate function call.

sda-download/api/middleware/middleware.go
```diff
@@ -85,6 +87,73 @@ func TokenMiddleware() gin.HandlerFunc {
...
+		clientVersionStr := c.GetHeader(headerName)
+
+		// 1. Check if the header is present
+		if clientVersionStr == "" {
+			errorMessage := fmt.Sprintf(
+				"Error: Missing required header '%s'. Please ensure you are using the latest sda-cli client.",
+				headerName,
+			)
+			log.Warnf("request blocked (412): Missing required header '%s'", headerName)
+			c.String(http.StatusPreconditionFailed, errorMessage)
```
jbygdell:
```suggestion
			c.String(http.StatusPreconditionFailed, "Missing client version header in request")
```

sda-download/api/middleware/middleware.go
```diff
@@ -85,6 +87,73 @@ func TokenMiddleware() gin.HandlerFunc {
...
+			return
+		}
+
+		// 2. Check if the client version is sufficient (clientVersion >= requiredVersion)
+		if clientVersion.LessThan(config.Config.App.MinimalCliVersion) {
+			errorMessage := fmt.Sprintf(
+				"Error: Your sda-cli client version '%s' is insufficient. Please update to at least version '%s' to proceed.",
+				clientVersionStr,
+				config.Config.App.MinimalCliVersionStr,
+			)
```
jbygdell:
```suggestion
			errorMessage := fmt.Sprintf("Error: Your sda-cli client version is too old, update to at least version '%s'.", config.Config.App.MinimalCliVersionStr)
```

jbygdell (changes requested):

sda-download/internal/config/config.go
```diff
@@ -24,7 +25,7 @@ const S3 = "s3"
 
 // availableMiddlewares list the options for middlewares
 // empty string "" is an alias for default, for when the config key is not set, or it's empty
-var availableMiddlewares = []string{"", "default"}
+var availableMiddlewares = []string{"", "default", "token"}
```
jbygdell: This becomes redundant due to the changes to the `SelectedMiddleware` in api.go

sda-download/api/api.go
```diff
@@ -9,18 +9,24 @@ import (
...
 
-// SelectedMiddleware is used to control authentication and authorization
-// behaviour with config app.middleware
-// available middlewares:
-// "default" for TokenMiddleware
-var SelectedMiddleware = func() gin.HandlerFunc {
-	return nil
+// SelectedMiddleware returns the middleware chain based on configuration.
+// For example, config.Config.App.Middleware could be "default", "token", etc.
+var SelectedMiddleware = func() []gin.HandlerFunc {
```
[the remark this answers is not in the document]
nanjiangshu: I don't think such unit tests are necessary since a bad configuration for `app.middleware` will be caught before this by `AppConfig` with 
```golang
if !slices.Contains(availableMiddlewares, c.App.Middleware) {
    return fmt.Errorf("app.middleware value=%v is not one of allowed values=%v", ...)
}
```

sda-download/api/middleware/middleware.go
```diff
@@ -85,6 +87,50 @@ func TokenMiddleware() gin.HandlerFunc {
...
+// ClientVersionMiddleware checks for the required "SDA-Client-Version" header.
+// It aborts the request with 412 (Precondition Failed) if the header is missing or
+// if the version does not meet the minimum required version.
+func ClientVersionMiddleware() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		clientVersionStr := c.GetHeader("SDA-Client-Version")
+
+		// Check if the header is present
+		if clientVersionStr == "" {
+			log.Warnf("request blocked (412): Missing client version header in request")
```
KarlG-nbis: Do we want to warn here and not debug? As this could lead to alot of extra logs from requests which we dont process, and I dont see much value in those logs being on warning level

KarlG-nbis (dismissed):

sda-download/api/api.go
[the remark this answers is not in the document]
KarlG-nbis: And from what I can see we already has such a case in the existing unit tests https://github.com/neicnordic/sensitive-data-archive/blob/main/sda-download/internal/config/config_test.go#L85

sda-download/api/middleware/middleware.go
jbygdell: yes, this has the possibility to become very noisy.
Also there is no point in having the HTTP response code in the log message.

sda-download/api/middleware/middleware.go
```diff
@@ -85,6 +87,50 @@ func TokenMiddleware() gin.HandlerFunc {
...
+			c.String(http.StatusPreconditionFailed, "Missing client version header in request")
+			c.Abort()
+
+			return
+		}
+
+		// Parse the client's provided version (using the processed string)
+		clientVersion, err := semver.NewVersion(clientVersionStr)
+		if err != nil {
+			log.Warnf("client version header '%s' is not a valid semantic version: %v", clientVersionStr, err)
```
jbygdell:
```suggestion
			log.Debugf("client version header '%s' is not a valid semantic version: %v", clientVersionStr, err)
```

nanjiangshu: fixed in 17711b66

jbygdell (approved):
You still have the HTTP response code in the log messages.

PR #2048, opened 2025-10-24
[multiple storage backends] Expand the config with the options

### LINKED ISSUE #1909: [multiple-backends] Configuration extension

jbygdell:
## Related issue(s) and PR(s)
This PR closes #1909 .

Not that the target is *not* main but a feature branch since the subsequent PRs will introduce breaking changes.

## Description
This PR adds functions to parse and validate the expected configurations for the use of multiple storage backends.
The static test tries to capture the most common issues I could think of.

## How to test
Run the static code tests

### DIFF
--- a/sda/internal/config/config.go
+++ b/sda/internal/config/config.go
@@ -4,10 +4,13 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
+	"net/url"
 	"os"
 	"strings"
 	"time"
 
+	"github.com/c2h5oh/datasize"
+	"github.com/go-viper/mapstructure/v2"
 	"github.com/neicnordic/crypt4gh/keys"
 	"github.com/neicnordic/sensitive-data-archive/internal/broker"
 	"github.com/neicnordic/sensitive-data-archive/internal/database"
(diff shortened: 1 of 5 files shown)


sda/internal/config/config.go
```diff
@@ -1121,6 +1124,84 @@ func configS3Storage(prefix string) storage.S3Conf {
...
+			config.ZeroFields = true
+		},
+	); err != nil {
+		return s3endpoints, err
+	}
+
+	for _, e := range list {
+		switch {
+		case e.URL == "":
+			return s3endpoints, errors.New("missing required parameter s3endpoint URL")
```
zeidlitz: I can see a value in keeping the index just for extra clarity so we can return like:

return s3endpoints, fmt.Errorf("s3endpoint[%d]: missing URL", i)

sda/internal/config/config.go
```diff
@@ -1121,6 +1124,84 @@ func configS3Storage(prefix string) storage.S3Conf {
...
+			return s3endpoints, errors.New("missing required parameter s3endpoint secret key")
+		case e.BucketPrefix == "":
+			return s3endpoints, errors.New("missing required parameter s3 bucket prefix")
+		default:
+			u, err := url.ParseRequestURI(e.URL)
+			if err != nil {
+				return s3endpoints, err
+			}
+			switch {
+			case !strings.HasPrefix(u.Scheme, "http"):
```
zeidlitz: I guess It's unlikely but this could accept things like "httpx" or similar. If we want to be super explicit just to be safe

if u.Scheme != "http" && u.Scheme != "https" 

might work?

PR #2126, opened 2025-12-04
bugfix fix sda-download range header issues

### LINKED ISSUE #2125: Fix sda-download Range header, respect the "start"
Currently the sda-download does not correctly handle the "Range" header, and it doesnt respect the "start" input in the value:`bytes=${start}-${end}`

Also if the "Range" header is invalid, 200 is returned instead of 300 to indicate a bad request.

KarlG-nbis:
## Related issue(s) and PR(s)
This PR closes #2125 

## Description
Fix issue with sda-download not respecting the start in the Range header, and return 400 if the range header can not be parsed as expected

## How to test
- Added and existing unit tests pass

### DIFF
--- a/sda-download/api/sda/sda.go
+++ b/sda-download/api/sda/sda.go
@@ -286,18 +286,19 @@ func Download(c *gin.Context) {
 
 		return
 	}
+	start, end, err = calculateCoords(start, end, c.GetHeader("Range"), fileDetails, c.Param("type"))
+	if err != nil {
+		log.Errorf("Byte range coordinates invalid! %v", err)
+		c.String(http.StatusBadRequest, "Range header byte coordinates invalid")
+
+		return
+	}
 
 	wholeFile := true //nolint:staticcheck
 	if start != 0 || end != 0 {
 		wholeFile = false
 	}
 
-	start, end, err = calculateCoords(start, end, c.GetHeader("Range"), fileDetails, c.Param("type"))
-	if err != nil {
-		log.Errorf("Byte range coordinates invalid! %v", err)
-
-		return
-	}
 	if c.Param("type") != "encrypted" {
 		// set the content-length for unencrypted files
 		if start == 0 && end == 0 {
--- a/sda-download/api/sda/sda_test.go
+++ b/sda-download/api/sda/sda_test.go
@@ -939,3 +939,389 @@ func GenerateTestC4ghKey(t *testing.T) (string, error) {
 
 	return privateKeyFile.Name(), nil
 }
+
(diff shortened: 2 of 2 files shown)


nanjiangshu:
I'm not quite sure what's the purpose of this PR, since even with the previous code, the `start` and `end` will be correctly parsed with a header like `"Range:bytes=100-200"`

This PR does fixes the problem with return value problem when parsing is failed for the function `calculateCoords`.

In addition, I don't quite understand the output of the partial file retrieved from the API. For example, when you retrieved the reencrypted data with 

```
curl -H "Authorization: Bearer $TOKEN" -H "Client-Public-Key: $pubkey" -H "Range:bytes=100-200" http://localhost:8080/s3/filepath  -o out.c4gh  
``` 
The output file `out.c4gh` has the `crypt4gh` magic number at the header but it is not a complete c4gh file thus unable to decrypt. I'm not sure how can it be used later on by `htsget`. Previously, for the unencrypted partially retrieved data, we can concatenate them, but now we can only retrieve encrypted data, of these partially retrieved encrypted data are all broken c4gh file, how can we concatenate them?

KarlG-nbis:
> I'm not quite sure what's the purpose of this PR, since even with the previous code, the `start` and `end` will be correctly parsed with a header like `"Range:bytes=100-200"`
> 
> This PR does fixes the problem with return value problem when parsing is failed for the function `calculateCoords`.
> 
> In addition, I don't quite understand the output of the partial file retrieved from the API. For example, when you retrieved the reencrypted data with
> 
> ```
> curl -H "Authorization: Bearer $TOKEN" -H "Client-Public-Key: $pubkey" -H "Range:bytes=100-200" http://localhost:8080/s3/filepath  -o out.c4gh  
> ```
> 
> The output file `out.c4gh` has the `crypt4gh` magic number at the header but it is not a complete c4gh file thus unable to decrypt. I'm not sure how can it be used later on by `htsget`. Previously, for the unencrypted partially retrieved data, we can concatenate them, but now we can only retrieve encrypted data, of these partially retrieved encrypted data are all broken c4gh file, how can we concatenate them?

Hmm good point, issue with the `crypt4gh` magic number when getting partial files was that the htsget-rs is "reconstructing" the file by getting the header and "content" areas separately, so when it was then trying to decrypt the full reconstructed file it failed (since a `crypt4gh` magic number was in the middle of the file). So im not sure atm how this would work when a user then tries to get a specifc area of the file with the range header if they can decrypt it

nanjiangshu (approved):
Great fix!

PR #2138, opened 2025-12-15
[SDA] Prevent starting with an old DB schema

jbygdell:
Components that create or interact with the correlation IDs are tightly bound to the DB schema and as such needs to only start on the correct version they can work with.


jbygdell:
> Shouldn't all services that query the database be added with this sanity check? 

Only those that interact with messages in an automated fashion.

PR #2186, opened 2026-01-15
Fix syntax and typo of the download-secrets chart

nanjiangshu:
## Related issue(s) and PR(s)
This PR fixes the syntax and typo of the chart so that the integration test can pass


KarlG-nbis (approved):

jbygdell (dismissed):
Indentations are irrelevant in a helm manifest.

And you need to bump the version tag in the `Chart.yaml` to 3.1.1 in order for the new chart to be accessible.

kostas-kou (approved):

Issue #2197, opened 2026-01-23
[helm charts] Update helm chart to populate now expected config needed by applications supporting multiple storage backends


KarlG-nbis:
As with the integration test, do this once all the applications have been merged into the temporary branch.

KarlG-nbis:
[PR](https://github.com/neicnordic/sensitive-data-archive/pull/2244) has now been merged

Issue #2196, opened 2026-01-23
[integration test] Update integration tests to verify multiple storage backends functionality


KarlG-nbis:
Once all the applications have been merged into a temporary branch, the integration tests can be updated in this branch.

PR #2203, opened 2026-01-23
Feature/multiple backends sync impl

### LINKED ISSUE #2185: [sda-sync] migrate sync service to support multiple storage backends
Update the sync service to support multiple storage backends introduced by #1910.

# A/C

A demo in sprint-review with sync using the new storage backend.

KarlG-nbis:
## Related issue(s) and PR(s)
This PR closes #2185

Failing integration tests will be fixed in #2196, and failing chart tests will be fixed in #2197 
before [feature/multiple-backends-applications-impl](https://github.com/neicnordic/sensitive-data-archive/tree/feature/multiple-backends-applications-impl) is merged to main

## Description
Update sda-sync to utlise multiple storage backends to be able to read archive file by their known archive location and write to a sync buckets with does not exceed its quota

## How to test
- sync unit tests pass

### DIFF
--- a/sda/cmd/sync/sync.go
+++ b/sda/cmd/sync/sync.go
@@ -36,7 +36,6 @@ var (
 	mqBroker      *broker.AMQPBroker
 	archiveReader storage.Reader
 	syncWriter    storage.Writer
-	message       schema.DatasetMapping
 )
 
 func main() {
@@ -162,6 +161,7 @@ func handleMessage(ctx context.Context, delivered amqp.Delivery) {
 		return
 	}
 
+	var message schema.DatasetMapping
 	// we unmarshal the message in the validation step so this is safe to do
 	_ = json.Unmarshal(delivered.Body, &message)
 


sda/cmd/sync/sync.go
```diff
@@ -4,214 +4,264 @@ package main
...
-		messages, err := mq.GetMessages(conf.Broker.Queue)
+	return nil
+}
+func startConsumer() error {
+	messages, err := mqBroker.GetMessages(mqBroker.Conf.Queue)
+	if err != nil {
+		return err
+	}
+	for delivered := range messages {
+		ctx := context.Background()
```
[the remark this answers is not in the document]
KarlG-nbis: Yes makes sense

sda/cmd/sync/sync.md
```diff
@@ -107,36 +107,17 @@ These settings control how sync connects to the RabbitMQ message broker.
 
 
 ### Storage settings
+The sync service requires access to the "archive", and "sync" storage. To configure that, the following configuration is required:
```
nanjiangshu:
```suggestion
The sync service requires access to the "archive" and "sync" storage backends. To configure these, the following configuration is required:
```

sda/cmd/sync/sync.go
```diff
@@ -4,214 +4,265 @@ package main
...
-	db                       *database.SDAdb
-	conf                     *config.Config
-	archive, syncDestination storage.Backend
+	key           *[32]byte
+	db            *database.SDAdb
+	conf          *config.Config
+	mqBroker      *broker.AMQPBroker
+	archiveReader storage.Reader
+	syncWriter    storage.Writer
+	message       schema.DatasetMapping
```
nanjiangshu: Better to define it as a local variable

KarlG-nbis:
> Great work! I only left a few minor comments.
> 
> One question, in the current setup, new bucket will not be automatically created when the bucket is hitting the limit (<500,000 files), right? Shall we have an issue for that?

Yeah that is true since we dont store the sync location in the database, so we currently dont have a few to check how many objects are in a "sync" location. But that would just require a new `LocationBroker` which supports finding that which we would initialize the sync writer with

PR #2205, opened 2026-01-23
Feature/multiple backends download impl

### LINKED ISSUE #1912: [sda-download] Migrate download service to support multiple storage backends
Update the download service to support multiple storage backends introduced by #1910.

The download service needs to look in the correct bucket/s3 endpoint to find the files.

# A/C

A demo in sprint-review with download using the new storage backend.

KarlG-nbis:
## Related issue(s) and PR(s)
This PR closes #1912

Failing integration tests will be fixed in #2196, and failing chart tests will be fixed in #2197 
before [feature/multiple-backends-applications-impl](https://github.com/neicnordic/sensitive-data-archive/tree/feature/multiple-backends-applications-impl) is merged to main

## Description
Update sda-download to utilise multiple storage backends to be able to read archive files by their known archived location

## How to test
- download unit tests pass


sda-download/cmd/main.go
```diff
@@ -31,6 +33,11 @@ func init() {
 		log.Panicf("database connection failed, reason: %v", err)
 	}
 	defer db.Close()
+	if db.Version < 23 {
+		log.Error("database schema v23 is required")
+
+		return
```
jhagberg: Is it not better to fail fast here instead of return? 
```suggestion
		log.Fatalf("database schema v23 is required (found v%d); process exiting", db.Version)
```
Will not the return allow the application to proceed to main() even if the database version is incompatible?

sda-download/internal/storage/v2/s3/reader/new_file_read_seeker.go
jhagberg: I'm curious about this validation: we compare `object.ContentRange` against `r.currentOffset`, but `prefetchAt(offset)` is fetching at `offset`. If `r.currentOffset` changes while the request is in-flight (e.g. due to seek/read), could we end up dropping valid prefetched data? Should this validate against `offset` (or the requested range) instead of `r.currentOffset`?

sda-download/internal/config/config.go
```diff
@@ -187,12 +185,6 @@ func NewConfig() (*Map, error) {
 		"db.host", "db.user", "db.password", "db.database", "oidc.configuration.url", "grpc.host",
 	}
 
-	if viper.GetString("archive.type") == S3 {
-		requiredConfVars = append(requiredConfVars, []string{"archive.url", "archive.accesskey", "archive.secretkey", "archive.bucket"}...)
```
jhagberg: Noticed this change removes the conditional `archive.*` required config validation from `NewConfig()`. Totally fine if the intention is to move all archive/storage validation into `storage.NewReader("archive")`, but it does mean misconfigurations will now fail later (and potentially with less context).

Would it be worth adding a small explicit “storage.archive is configured” check here (or at least a clearer startup error) so we fail fast with a more user-friendly message?

sda-download/internal/database/database.go
```diff
@@ -125,6 +126,19 @@ func (dbs *SQLdb) checkAndReconnectIfNeeded() {
...
+
+	log.Debug("Fetching database schema version")
+
+	query := "SELECT MAX(version) FROM sda.dbschema_version;"
+
+	var dbVersion = -1
+	err := dbs.DB.QueryRow(query).Scan(&dbVersion)
+
+	return dbVersion, err
+}
```
jhagberg: I'm a bit concerned `SELECT MAX(version)` may return NULL if `sda.dbschema_version` is empty, which will make `Scan(&dbVersion)` fail with a somewhat opaque error.

Would it make sense to either:
- use `SELECT COALESCE(MAX(version), -1) ...` and then handle `-1` explicitly with a clear "dbschema_version not initialized" error, or
- scan into `sql.NullInt64` so we can return a more descriptive message when the value is NULL?

Given our deployment workflow (and the schema v23 requirement), which approach do you prefer?

sda-download/cmd/main.go
KarlG-nbis: Ah yes, good point missed that this logic was in the init func

sda-download/internal/storage/v2/s3/reader/new_file_read_seeker.go
```diff
@@ -257,11 +256,11 @@ func (r *s3Reader) isPrefetching(offset int64) bool {
 }
 
 // wholeReader is a helper for when we read the whole object
-func (r *s3Reader) wholeReader(dst []byte) (int, error) {
+func (r *s3SeekableReader) wholeReader(dst []byte) (int, error) {
 	if r.objectReader == nil {
 		// First call, setup a reader for the object
-		object, err := r.Client.GetObject(&s3.GetObjectInput{
-			Bucket: aws.String(r.Bucket),
+		object, err := r.client.GetObject(context.Background(), &s3.GetObjectInput{
```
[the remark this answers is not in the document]
KarlG-nbis: Added a defer func closing object.Body

sda-download/internal/storage/v2/s3/reader/config.go
```diff
@@ -0,0 +1,158 @@
...
+				if err != nil {
+					return nil, errors.New("could not parse chunk_size as a valid data size")
+				}
+				if s < 5*datasize.MB {
+					return nil, errors.New("chunk_size can not be smaller than 5mb")
+				}
+				if s > 1*datasize.GB {
+					return nil, errors.New("chunk_size can not be bigger than 1gb")
+				}
+			}
```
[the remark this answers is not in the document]
KarlG-nbis: Ahh yes, good catch, will fix

sda-download/internal/database/database.go
KarlG-nbis: Yeah makes sense, will update to scan to a NullInt64 and update SQLdb.Verison from int -> int64

sda-download/internal/storage/v2/s3/reader/new_file_read_seeker.go
KarlG-nbis: I can not give a good answer to the original intent.

But I think this is intended, as if `r.currentOffset` has changed while request was in flight then `seek` must have been called from what I can gather, as in `read` and `prefetchAt` we are checking if the `currentOffset` is already in the cache, meaning a `GetObject` would only be called if we are getting new data. So if `currentOffset` has changed a new prefetchAt would have spawned and be waiting for a look (and could already be fetching data), so even if this content range would be partially valid we could then have 2 overlapping cache entries (which would be handled) but I think yeah this logic could be expanded to evaluate if the fetched content range is still valid from the changed `currentOffset`, eg,   

```go
    if object.ContentRange == nil {
		return
	}

	contentRange, err := parseContentRange(*object.ContentRange)
	if err != nil {
		log.Errorf("failed to parse content range: %v", err)

		return
	}

	if r.currentOffset < contentRange.start || r.currentOffset > contentRange.end {
		return
	}
	
...


type contentRange struct {
	start int64
	end   int64
	size  int64 // -1 if unknown (*)
}

// parseContentRange parses a content range header, with the format bytes $start-$end/$size
func parseContentRange(v string) (*contentRange, error) {
	if !strings.HasPrefix(v, "bytes ") {
		return nil, errors.New("invalid Content-Range header")
	}

	v = strings.TrimPrefix(v, "bytes ")
	parts := strings.Split(v, "/")
	if len(parts) != 2 {
		return nil, errors.New("invalid Content-Range")
	}

	rangePart := parts[0]
	sizePart := parts[1]

	re := strings.Split(rangePart, "-")
	if len(re) != 2 {
		return nil, errors.New("invalid Content-Range header")
	}

	start, err := strconv.ParseInt(re[0], 10, 64)
	if err != nil {
		return nil, fmt.Errorf("failed to parse start in content range header: %v", err)
	}

	end, err := strconv.ParseInt(re[1], 10, 64)
	if err != nil {
		return nil, fmt.Errorf("failed to parse end in content range header: %v", err)
	}

	size := int64(-1)
	if sizePart != "*" {
		size, err = strconv.ParseInt(sizePart, 10, 64)
		if err != nil {
			return nil, fmt.Errorf("failed to parse size in content range header: %v", err)
		}
	}

	return &contentRange{
		start: start,
		end:   end,
		size:  size,
	}, nil
}

```

What do you think?

sda-download/internal/storage/v2/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+| Name:         | Type:  | Default Value: | Description:                                                                                                                                  |         
+|---------------|--------|----------------|-----------------------------------------------------------------------------------------------------------------------------------------------|                               
+| access_key    | string |                | The access key used to authenticate when connecting to the s3                                                                                 |        
+| secret_key    | string |                | The secret key used to authenticate when connecting to the s3                                                                                 |
+| ca_cert       | string |                | The ca certificate of the s3 to be appended to the certs of the system                                                                        |        
+| chunk_size    | string | 50MB           | The chunk size (in bytes) when writing data S3, also when reading chunks with the Seekable Reader. The minimum allowed is 5MB, and max is 1gb |
```
nanjiangshu:
```suggestion
| chunk_size    | string | 50MB           | The chunk size used when writing data to `S3` and when reading data with the `Seekable Reader`. Supports values like 50MB, 100MB. The minimum allowed value is 5MB, and the maximum is 1GB.  |
```

I have clarified the description of chunk_size since otherwise it is unclear if it is a string like "50MB" or an integer number of bytes.

sda-download/internal/storage/v2/README.md
```diff
@@ -0,0 +1,80 @@
...
+| region        | string | us-east-1      | The region of the s3 bucket                                                                                                                   |        
+| endpoint      | string |                | The address of the s3 buckets                                                                                                                 |        
+| disable_https | bool   | false          | If to disable https when connecting to the s3 bucket                                                                                          |        
+| bucket_prefix | string |                | How the reader will identify which buckets to look through when looking for a file for which the location is not known by the caller          |
```
nanjiangshu: Might be good to add a paragraph to clarify the `bucket_prefix` further and explain 
- Whether buckets are searched sequentially
- Whether the first match wins

sda-download/dev_utils/compose-sda.yml
```diff
@@ -17,35 +17,52 @@ services:
...
+      - PGPASSWORD=rootpassword
   s3:
     command: server /data
     container_name: s3
     environment:
-      - MINIO_ACCESS_KEY=access
-      - MINIO_SECRET_KEY=secretkey
+      - MINIO_ROOT_USER=access
+      - MINIO_ROOT_PASSWORD=secretkey
+      - MINIO_SERVER_URL=http://127.0.0.1:9000
```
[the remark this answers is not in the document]
KarlG-nbis: yeah, updated to use https

sda-download/internal/storage/v2/posix/reader/find_file.go
```diff
@@ -0,0 +1,29 @@
...
+	"io/fs"
+	"os"
+	"path/filepath"
+
+	"github.com/neicnordic/sda-download/internal/storage/v2/storageerrors"
+)
+
+func (reader *Reader) FindFile(_ context.Context, filePath string) (string, error) {
+	for _, endpointConf := range reader.configuredEndpoints {
+		_, err := os.Stat(filepath.Join(endpointConf.Path, filePath))
```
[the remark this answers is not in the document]
KarlG-nbis: Well I would file this under a future improvement for the readers / writer to do such validation, atm the file path is gotten from the database / generated, eg [ingest writing to archive](https://github.com/neicnordic/sensitive-data-archive/blob/feature/multiple-backends-ingest-impl/sda/cmd/ingest/ingest.go#L539), [download reading from archive](https://github.com/neicnordic/sensitive-data-archive/blob/feature/multiple-backends-download-impl/sda-download/api/sda/sda.go#L323), only excpetion is for the ingest when a file has been uploaded through means with did not register the file in out database https://github.com/neicnordic/sensitive-data-archive/blob/feature/multiple-backends-ingest-impl/sda/cmd/ingest/ingest.go#L330

sda-download/internal/storage/v2/seekable_multi_reader.go
```diff
@@ -0,0 +1,105 @@
...
+
+func (r *seekableMultiReader) Read(dst []byte) (int, error) {
+	var readerStartAt int64
+
+	for i, reader := range r.readers {
+		if r.currentOffset < readerStartAt {
+			// We want data from a previous reader (? HELP ?)
+			readerStartAt += r.sizes[i]
+
+			continue
```
nanjiangshu:
```suggestion
            return 0, errors.New("internal error: currentOffset before reader start")
```
If this condition happens, it should be an error, isn't it?
when `currentOffset < readerStartAt`, the correct reader was earlier, continuing forward makes the error worse. I think it should either `break` or return an internal error.

sda-download/dev_utils/run_integration_test_no_tls.sh
```diff
@@ -31,52 +31,52 @@ chmod 444 c4gh.sec.pem
 # insert file entry into database
 file_id=$(docker run --rm --name client --network dev_utils_default \
     neicnordic/pg-client:latest \
-    postgresql://postgres:rootpassword@db:5432/lega \
+    postgresql://postgres:rootpassword@db:5432/sda \
     -t -q -c "INSERT INTO sda.files (stable_id, submission_user, \
```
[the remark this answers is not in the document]
nanjiangshu: `psql -At` gives exactly the value you need, with no padding, no headers, so you can just assign it directly.
With `xargs`, if the returned value contains spaces, tabs, or shell-special characters ($, !, *, ?), xargs will split or expand them.

I think the reason why it works in this exact case is because there is one row returned with the psql query. 
But it's safer to use `psql -At`. It works even if wants to retrieve e.g. `file_ids` as a list of IDs. 
After using `psql -At`, the `| xargs` can be removed.

sda-download/internal/storage/v2/posix/reader/config.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/spf13/viper"
+)
+
+type endpointConfig struct {
+	Path string `mapstructure:"path"`
+}
+
+func loadConfig(backendName string) ([]*endpointConfig, error) {
+	var endpointConf []*endpointConfig
+
```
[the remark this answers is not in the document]
nanjiangshu: ok.

sda-download/internal/storage/v2/seekable_multi_reader.go
KarlG-nbis: Yeah tbh didnt look to closely in this code and it was mainly copied from previous implementation, but yeah makes sense to return an error in this scenario

jbygdell (approved):

PR #2217, opened 2026-02-02
fix(storage-v2): fix check if s3 endpoint has http / https prefix

KarlG-nbis:
## Related issue(s) and PR(s)
This PR relates to #1910 

## Description
Bug in storage v2, s3 reader and writer, could not setup a reader / writer with tls as https has http as prefix, add ":" to prefix check to ensure we check for full scheme name


jbygdell (approved):

nanjiangshu (approved):
Looks good!

PR #2226, opened 2026-02-04
feat(download): add support for Htsget-Context-Public-Key header with…

### LINKED ISSUE #2225: [download] Add support for getting encryption key from "Htsget-Context-Public-Key" header
# Description
In an update of the htsget-rs service, it will now send its public key in the "Htsget-Context-Public-Key" header, so the sda-download needs to support getting the public key, either from the "Htsget-Context-Public-Key" or "Client-Public-Key" header, with "Htsget-Context-Public-Key" taking precedence

KarlG-nbis:
## Related issue(s) and PR(s)
This PR closes #2225.

## Description
Add support for Htsget-Context-Public-Key header with takes precedence over Client-Public-Key header

### DIFF
--- a/sda-download/api/sda/sda.go
+++ b/sda-download/api/sda/sda.go
@@ -579,13 +579,13 @@ func getPublicKeyFromRequest(c *gin.Context) (string, error) {
 		return "", errors.New("both Htsget-Context-Public-Key, and Client-Public-Key headers are set")
 	case htsgetRsPublicKeyHeader != "":
 		if _, err := base64.StdEncoding.DecodeString(htsgetRsPublicKeyHeader); err != nil {
-			return "", fmt.Errorf("invalid base64 encoding for Htsget-Context-Public-Key header")
+			return "", errors.New("invalid base64 encoding for Htsget-Context-Public-Key header")
 		}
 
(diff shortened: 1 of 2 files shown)


sda-download/api/sda/sda.go
```diff
@@ -567,3 +562,21 @@ var calculateCoords = func(start, end int64, htsget_range string, fileDetails *d
...
+	htsgetRsPublicKeyHeader := c.GetHeader("Htsget-Context-Public-Key")
+	clientPublicKeyHeader := c.GetHeader("Client-Public-Key")
+
+	switch {
+	case htsgetRsPublicKeyHeader != "":
+		if clientPublicKeyHeader != "" && htsgetRsPublicKeyHeader != clientPublicKeyHeader {
+			log.Warnf("both Htsget-Context-Public-Key, and Client-Public-Key are set and differnt, defaulting to Htsget-Context-Public-Key")
+		}
+
+		return htsgetRsPublicKeyHeader
```
nanjiangshu: I'm not sure if we should ever allow input of multiple keys and then default to one. In the project like SDA, I think explicit is better than defaulting.

pahatz (approved):
Looks good

Issue #1680, opened 2025-05-15
[sda] Create the new download app


jbygdell:
Create the new download app using the swagger.yaml found in the `sda/cmd/download` folder as functional reference.

jhagberg:
and what about LinkedIdentities? 8D in the general requirments https://ga4gh.github.io/data-security/ga4gh-passport should we implement support "a Passport Clearinghouse MUST check the LinkedIdentities Visas by trusted issuers and ensure that trusted sources have asserted that these Visa Identities represent the same end user." ?

PR #2232, opened 2026-02-09
Download API v2.0.0 specification

### LINKED ISSUE #2229: [download] Create an API specification that conforms to the user stories
In particular these from the parent issue.

**US3** _should_
As a data recipient
I want an API that is easy to use and understand
so that I can quickly do my job without issue.

**US10** _should_
As a data recipient
I want to be able to download partial files
so that I can download only the part of the file I need.

# Demo

Show the swagger API in the sprint review.

jhagberg:
## Related issue(s) and PR(s)
This PR closes #2229

## Download API v2.0.0 Specification

Clean-break v2.0.0 spec for the SDA Download API. v1.0 was never deployed, so there is nothing to deprecate.

---

### Three-Tier Download Endpoints

Files are served through three endpoint tiers (design credit: Pontus Freyhult):

| Endpoint | Purpose | Public key required | ETag stability | Range support |
|---|---|---|---|---|
| `GET /files/{fileId}` | Combined stream (convenience) | Yes | Implementation-dependent | Yes (over header+data) |
| `GET /files/{fileId}/header` | Re-encrypted C4GH header only | Yes | Implementation-dependent | No |
| `GET /files/{fileId}/content` | Encrypted data segments only | **No** | Stable (fileId-scoped) | Yes (over data only) |

The split works because Crypt4GH data segments are encrypted with the DEK, which is the same regardless of recipient. Only the header (containing the DEK re-encrypted to a specific recipient) changes.

Each endpoint has a `HEAD` variant for metadata discovery.

### Range Header Semantics

`Range: bytes=X-Y` refers to the **encrypted wire format**, not plaintext. The exact bytes depend on the endpoint:

- `/files/{fileId}` — Range over the combined `[re-encrypted header | data segments]` stream
- `/files/{fileId}/content` — Range over data segments only (no header)

Aligns with [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110) § 14.2. Single range only; multi-range returns 400.

### ETag Scoping

- **Combined endpoint:** ETag scoped to `(fileId, recipientPublicKey)`. Stability depends on whether the server produces deterministic re-encrypted headers. If not, `If-Range` falls back to full 200 (correct per RFC 9110).
- **Content endpoint:** ETag scoped to `fileId` only — stable and recipient-independent. **Recommended for robust resume.**
- Per [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110) § 8.8.3: entity-tags must be unique across representations

### Public Key Headers

Both `X-C4GH-Public-Key` and `Htsget-Context-Public-Key` are marked `required: false` in OpenAPI (3.x cannot express "at least one of" — see [OAI#4174](https://github.com/OAI/OpenAPI-Specification/discussions/4174)). The server enforces that exactly one must be provided (400 if neither or both).

The `/content` endpoint does not require a public key at all.

### Dual Public Key Format

Supports two formats with auto-detection by decoded length:

- **Raw** (preferred): `base64(32-byte X25519 public key)` — decoded length = 32 bytes
- **Legacy PEM**: `base64(PEM-encoded key)` — decoded length > 32 bytes

Raw format aligns with [GA4GH Htsget 1.3+](https://samtools.github.io/hts-specs/htsget.html) (`Htsget-Context-Public-Key`). Legacy PEM maintains backwards compatibility with existing sda-cli.

---

### `/service-info` Endpoint

New `GET /service-info` endpoint per [GA4GH Service Info 1.0](https://github.com/ga4gh-discovery/ga4gh-service-info). Uses `artifact: sda-download` (not `drs`). No authentication required.

### Checksums and Decrypted Size in FileInfo

- `checksums` — array of `{type, checksum}` matching DRS checksums format
- `decryptedSize` — plaintext size in bytes
- `size` — encrypted archive size (excluding re-encrypted header)

### `filePath` and `pathPrefix` Query Filters

`GET /datasets/{datasetId}/files` supports two optional, mutually exclusive filters (providing both returns 400):

- `filePath=path/to/file.c4gh` — exact path equality (returns 0 or 1 results)
- `pathPrefix=samples/controls/` — recursive prefix match (returns all files under the prefix; use trailing `/` for directory boundary)

See user stories US13 (limit file listing) and US14 (pagination) in #1680.

### Cursor-Based Pagination

Both list endpoints (`GET /datasets`, `GET /datasets/{datasetId}/files`) support cursor-based pagination:

- `pageSize` — max items per page (default 100, max 1000)
- `pageToken` — opaque cursor from `nextPageToken` in the previous response

Responses use envelope objects (`{ datasets: [...], nextPageToken }` / `{ files: [...], nextPageToken }`) instead of bare arrays. This is a **breaking change** from v1.

**Pagination contract:**
- Stable, server-defined ordering across pages
- `pageToken` is scoped to the original query parameters — changing filters/pageSize with an existing token returns 400
- Invalid or expired `pageToken` returns 400 (no silent restart)

### Error Responses

`application/problem+json` per [RFC 9457](https://www.rfc-editor.org/rfc/rfc9457). The `BadRequest` (400) response includes named examples for common error cases: invalid Range, filter conflicts, invalid page token, and invalid page size.

### Legacy Endpoints Removed

All legacy endpoints (`/file`, `/file/{fileId}`, `/info/datasets`, `/info/dataset`, `/info/dataset/files`) are removed — v1.0 was never deployed.

### References

- [RFC 9110 — HTTP Semantics](https://www.rfc-editor.org/rfc/rfc9110)
- [RFC 9457 — Problem Details for HTTP APIs](https://www.rfc-editor.org/rfc/rfc9457)
- [GA4GH Service Info 1.0](https://github.com/ga4gh-discovery/ga4gh-service-info)
- [GA4GH DRS 1.5.0](https://ga4gh.github.io/data-repository-service-schemas/preview/release/drs-1.5.0/docs/)
- [GA4GH Htsget 1.3](https://samtools.github.io/hts-specs/htsget.html)
- [Crypt4GH Specification](http://samtools.github.io/hts-specs/crypt4gh.pdf)

## How to test
https://editor.swagger.io/?url=https://raw.githubusercontent.com/neicnordic/sensitive-data-archive/refs/heads/main/sda/cmd/download/swagger_v2.yml


sda/cmd/download/swagger_v1.yml
```diff
@@ -1,226 +1,571 @@
...
+        The public key is required because Content-Length includes the recipient-specific
+        Crypt4GH header, which varies in size per recipient. The ETag also represents
+        the specific representation for the given (fileId, recipient public key) pair;
+        different keys produce different ETags.
+
+        This endpoint is essential for implementing resumable downloads (HEAD → Range + If-Range).
+      parameters:
+        - $ref: "#/components/parameters/FileIdPath"
+        - $ref: "#/components/parameters/C4ghPublicKey"
+        - $ref: "#/components/parameters/HtsgetContextPublicKey"
```
KarlG-nbis: Will `C4ghPublicKey` not always be required even if `HtsgetContextPublicKey` is supplied?

sda/cmd/download/swagger_v1.yml
```diff
@@ -1,226 +1,571 @@
...
+          $ref: "#/components/responses/InternalServerError"
       security:
         - bearerAuth: []
+
   /info/dataset/files:
     get:
-      description: Returns an array with metadata about all files in a dataset
+      tags: [Datasets]
+      operationId: legacyListDatasetFiles
+      deprecated: true
```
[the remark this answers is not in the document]
jhagberg: True

KarlG-nbis:
Also possibly worth looking into https://rclone.org/http/ (or other possible remote types) if we want to support rclone reading, and ensure the spec allows this, eg so that a user could copy a whole dataset with

```
rclone copy https://download.bp.nbis.se/${DATASET_ID}/ --http-headers "Authorization=Bearer <token>"
```

nanjiangshu:
> > Range Header Semantics
> > Range: bytes=X-Y refers to the re-encrypted Crypt4GH byte stream [re-encrypted header | encrypted data segments], not plaintext or the original archive file.
> 
> This will become a problem when calling expected byte ranges from a file. Ir the user wants say the first 750 bytes of a file the request needs to contain information the user has no access to. data = 750 bytes, c4gh-header=at least 128 bytes, c4gh padding (nonce + MAC) for each 64KB data block.
> 
I'm not quite sure about the main concern here. Is that you are worried if the user only have access to 0-1000 bytes and requested 0-750 bytes, but the download-service with anyway return 64KB?

I don't this will be a problem:
- Firstly, authorization happens at the **htsget** level. If the user is not authorized for the data in the rest of that 64KB block the server would theoretically be responsible for not serving that block.
- Secondly, in genomics, permissions are almost always granted at the file/sample level, not the individual base-pair level.

> That means in order for the users to get the expected 750 bytes we need to deliver ~64kB data.

This overhead will only be a problem when a user makes thousands of tiny 750-bytes requests, but I guess this is usually not the case for standard genomic work when downloading slices of BAM files. I believe 64KB is a good block size when they defined this specification. 


> And in the request is larger say 1MB the delivered data will then be 16*64kB plus block overhead and header

sda/cmd/download/swagger_v1.yml
KarlG-nbis: Yeah making both required: false and document the "at least one required" constraint in the description text sounds like the best approach given the circumstances imo, and we enforce it on the server implementation side

sda/cmd/download/swagger_v1.yml
```diff
@@ -1,226 +1,366 @@
 openapi: 3.0.3
 info:
-  title: SDA file download API
-  version: "1.0"
-  description: This is the external data out API for the Sensitive Data Archive. Data is delivered encrypted with a user supplied crypth4gh public key.
+  title: SDA Data Out API
+  version: "1.1.0"
```
KarlG-nbis: Also dont think we need to increment the version as the v1.0.0 was never supported or implemented by anything, unless we want to start it at v2.0.0 now to clearly separate it from the APs served by the current sda-download

jhagberg:
> I'm not sure I see the problem Jocke notes (or see it as something new, maybe most).
> 
> But for me, I've evolved to actually preferring having two separate endpoints for file download, one to fetch a valid header for the file (reencrypted with the public key) and one to download the content (would be static).
> 
> (I'm not sure it's more than a theoretical worry, but given that the header size is said to be changable, it's problematic even simple things like the same range purely in the "content" possibly returning different things on subsequent requests. Also, given the volatile nature of reencrypted headers I'm not sure the E-Tag is usable.)
> 
> I'm sure we might want to keep an endpoint for the combined header + content as well, but I'm not sure if it's required.

I liked your idea! Implemented your split-endpoint design.  The spec now has three tiers:
- GET /files/{fileId} — combined stream (convenience, simple clients)
- GET /files/{fileId}/header — re-encrypted header only (requires public key)
- GET /files/{fileId}/content — encrypted data segments only, no public key needed, stable ETag scoped to fileId only, full Range/If-Range support

Also tightened the Range parameter docs — it now explicitly says what bytes the range refers to per endpoint (combined = header+data, content = data only).

sda/cmd/download/swagger_v2.0.0.yml
```diff
@@ -1,226 +1,571 @@
...
+        "403":
+          $ref: "#/components/responses/Forbidden"
+        "404":
+          $ref: "#/components/responses/NotFound"
+        "500":
+          $ref: "#/components/responses/InternalServerError"
+      security:
+        - bearerAuth: []
+
+  /files/{fileId}:
```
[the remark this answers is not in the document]
KarlG-nbis: Yeah I mean worth asking him if the htsget-rs supports DRS. But from my testing and experience with htsget-rs, it needs to get the bai and bam files which it does based on the user input eg:
```
http://localhost:8088/reads/${id}?encryptionScheme=C4GH&referenceName=11
```
where ${id} would be something like `DATASET0001/htsnexus_test_NA12878`

which would be translated by htsget-rs to

`${backend.url}/DATASET0001/htsnexus_test_NA12878.bam.bai.c4gh` and `${backend.url}/DATASET0001/htsnexus_test_NA12878.bam.c4gh`

where the ${backend.url} is a htsget-rs URL storage config

given the formatting of ${id} https://github.com/umccr/htsget-rs/blob/main/htsget-search/src/search.rs#L265 
[fmt_file](https://github.com/umccr/htsget-rs/blob/main/htsget-config/src/types.rs#L52) -> [index_file_ending](https://github.com/umccr/htsget-rs/blob/main/htsget-config/src/types.rs#L57) ... -> [format_key](https://github.com/umccr/htsget-rs/blob/main/htsget-storage/src/c4gh/storage.rs#L110)

KarlG-nbis:
> > But yeah we have a problem if we expect the c4gh header to change size, but I do not believe I've ever encountered that issue so far during my testing at least
> 
> Crypt4gh supports encrypting the header with multiple public keys and each key used will increase the size of the header.

> It's also supported having multiple symmetric keys for a file (for supporting stitching together) and in theory, a server could use a download range to filter what symmetric keys are needed.


Yeah ofc, but that is not something we support currently with the reencrypt service as far as I can see, but it will become a possible problem if we are to start supporting it

sda/cmd/download/swagger_v2.0.0.yml
```diff
@@ -0,0 +1,1186 @@
...
+            Content-Length:
+              $ref: "#/components/headers/ContentLength"
+        "400":
+          $ref: "#/components/responses/BadRequest"
+        "401":
+          $ref: "#/components/responses/Unauthorized"
+        "403":
+          $ref: "#/components/responses/Forbidden"
+        "404":
+          $ref: "#/components/responses/NotFound"
```
jbygdell:
```suggestion
```

sda/cmd/download/swagger_v2.0.0.yml
```diff
@@ -0,0 +1,1186 @@
...
+            ETag:
+              $ref: "#/components/headers/ContentETag"
+            Last-Modified:
+              $ref: "#/components/headers/LastModified"
+        "401":
+          $ref: "#/components/responses/Unauthorized"
+        "403":
+          $ref: "#/components/responses/Forbidden"
+        "404":
+          $ref: "#/components/responses/NotFound"
```
jbygdell:
```suggestion
```

sda/cmd/download/swagger_v2.0.0.yml
```diff
@@ -0,0 +1,1186 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/DatasetInfo"
+        "401":
+          $ref: "#/components/responses/Unauthorized"
+        "403":
+          $ref: "#/components/responses/Forbidden"
+        "404":
+          $ref: "#/components/responses/NotFound"
```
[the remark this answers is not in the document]
jhagberg: resolved: 87b820c

jbygdell (dismissed):

jbygdell (approved):

sda/cmd/download/swagger_v2.0.0.yml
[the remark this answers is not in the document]
KarlG-nbis: Sounds good!

sda/cmd/download/swagger_v2.0.0.yml
```diff
@@ -0,0 +1,1184 @@
+openapi: 3.0.3
+info:
+  title: SDA Data Out API
+  version: "2.0.0"
```
KarlG-nbis: Minor: do we want the full version in the file name(`swagger_v2.0.0`)? Doesnt that add additional overhead when we need to bump patch / minor in the future to both rename file, and update version here in the file? Would make sense to keep filename as `swagger_v2.yaml` thou as for a new major version it makes sense to create new file to be able to keep old, etc

jhagberg: Yes I was thinking about that...

sda/cmd/download/swagger_v2.yml
```diff
@@ -0,0 +1,1184 @@
+openapi: 3.0.3
+info:
+  title: SDA Data Out API
+  version: "2.0.0"
```
jhagberg: Fixed!

KarlG-nbis (approved):

PR #2242, opened 2026-02-12
Feature/storage v2 non database location broker

### LINKED ISSUE #2241: [storage-v2] Update Location Broker to support being able to know size and object count in other storages besides "inbox", "archive", or "backup"
Currently the [LocationBroker](https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/internal/storage/v2/locationbroker/location_broker.go) is dependant on the database to know how many objects are in a storage, and in the database we only record the storage location of the following storages: "inbox", "archive", and "backup", meaning that for other storages we will not know the size or how many objects are in that location.

The main problem is for "sync".

The aim of this ticket is to implement a "default" way to look up the size and object count of a location for a storage for which we dont store this information in the DB.

KarlG-nbis:
## Related issue(s) and PR(s)
This PR closes #2241 .

## Description
Update storage-v2 Location Broker to be able find the size and object count of a storage for which we do not store that information in the database. So for "inbox", "archive", and "backup" storage backends we have the location per file stored in the database so we can query the database to know the size and object count of a location, but for other storages we can not.

This PR introduces a way for the storage-v2 implementations to register a "default" size and object count finder func to the Location Broker which will be used if the backend name is not "inbox", "archive", or "backup"

### DIFF
--- a/sda/internal/storage/v2/s3/writer/writer.go
+++ b/sda/internal/storage/v2/s3/writer/writer.go
@@ -124,12 +124,10 @@ func findSizeAndObjectCountOfLocation(configuredEndpoints []*endpointConfig) fun
 		var totalSize uint64
 		var totalObjects uint64
 
-		// Create paginator
 		paginator := s3.NewListObjectsV2Paginator(s3Client, &s3.ListObjectsV2Input{
 			Bucket: aws.String(bucket),
 		})
 
-		// Iterate through all pages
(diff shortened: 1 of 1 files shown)


jbygdell:
This feels like a stopgap for the sync related issues.
Probably worth more to actually fix sync so that we can use finalizes backup functionality to copy the files and only send the re-encrypted header and file metadata to the sync API.

Issue #378, opened 2022-10-25
support for multiple s3 buckets


blankdots:
### Description
Different s3 API storage implementations might have limited number of objects per bucket (e.g. 500 000), for this reason there is a need to support multiple buckets for the each step of the ingestion process.

This might also require the sda-db to keep track of the bucket an object resides in.

### Definition of Done
Analyze the implication if this can be done
`sda-pipeline` can support multiple buckets.

### Testing
Integration tests

viklund:
We need to support multiple s3 backends as well. One simple solution for parts of this would be to require the hostname to be part of the filepath in the database.

jbygdell:
Update:

Since the user id cannot be part of the visible path we need to split the full path over two columns, one that holds the relative path and one that holds the endpoint url and bucket.

This way it will be possible to shard each user into a separate bucket on upload.

Archive buckets can be separated in the same manner, by date or submission.

PR #2244, opened 2026-02-13
Feature/multiple backends applications impl

### LINKED ISSUE #2196: [integration test] Update integration tests to verify multiple storage backends functionality

KarlG-nbis:
## Related issue(s) and PR(s)
This PR closes #2196, #2197 (#378).

## Description
Updates the applications using storage backends to utilise the storage-v2 (also contains some storage-v2 fixes).

Updates integration tests to verify multiple storage backends functionality

## How to test
- Unit tests pass
- Integration tests pass

### DIFF
--- a/charts/sda-svc/templates/job-perform-db-optimization.yaml
+++ b/charts/sda-svc/templates/job-perform-db-optimization.yaml
@@ -3,6 +3,10 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: optimize-db
+  annotations:
+    helm.sh/hook: pre-install,pre-upgrade
+    helm.sh/hook-weight: "0" # Hook weight is just one lower than job using config map to ensure it's applied before job
+    helm.sh/hook-delete-policy: hook-succeeded
 data:
   optimize-db.sql: |
     DO
--- a/charts/sda-svc/templates/job-set-missing-keyhash.yaml
+++ b/charts/sda-svc/templates/job-set-missing-keyhash.yaml
@@ -3,6 +3,12 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: set-key-hash
+  annotations:
+    # This is what defines this resource as a hook.
+    # Without this line, the job is considered part of the release.
+    helm.sh/hook: post-install,post-upgrade
+    helm.sh/hook-weight: "-5" # Hook weight is just one lower than job using config map to ensure it's applied before job
+    helm.sh/hook-delete-policy: hook-succeeded
 data:
   set-key-hash.sql: |
     DO
--- a/charts/sda-svc/templates/job-upgrade-from-v1.yaml
+++ b/charts/sda-svc/templates/job-upgrade-from-v1.yaml
@@ -3,6 +3,12 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: upgrade-from-v1
+  annotations:
+    # This is what defines this resource as a hook.
+    # Without this line, the job is considered part of the release.
+    helm.sh/hook: post-install,post-upgrade
+    helm.sh/hook-weight: "-6" # Hook weight is just one lower than job using config map to ensure it's applied before job
+    helm.sh/hook-delete-policy: hook-succeeded
 data:
   upgrade-from-v1.sql: |
     DO


jbygdell (approved):

sda/cmd/verify/verify.go
```diff
@@ -4,408 +4,492 @@ package main
...
-				if strings.Contains(err.Error(), "no such file or directory") || strings.Contains(err.Error(), "NoSuchKey:") || strings.Contains(err.Error(), "NotFound:") {
-					jsonMsg, _ := json.Marshal(map[string]string{"error": err.Error()})
-					if err := db.UpdateFileEventLog(message.FileID, "error", "verify", string(jsonMsg), string(delivered.Body)); err != nil {
-						log.Errorf("failed to set ingestion status for file from message, file-id: %v", message.FileID)
-					}
-				}
+		if err := delivered.Nack(false, true); err != nil {
+			log.Errorf("failed to Nack message, reason: (%s)", err.Error())
+		}
+	}
```
jhagberg: **Bug: Missing `return` after `Nack`**

After `GetArchiveLocation` fails and the message is Nack'd, there's no `return` — execution falls through to the `archiveLocation == ""` check and continues processing with potentially invalid state. This could cause a double-nack or unexpected behavior.

Note: The Nack requeue behavior was discussed in [#2204](https://github.com/neicnordic/sensitive-data-archive/pull/2204), but this fall-through control flow issue is separate.

sda/cmd/api/api.go
```diff
@@ -49,57 +51,112 @@ var (
...
+			}
 		}
-	} else {
-		log.Infof("Starting web server at http://%s:%d", Conf.API.Host, Conf.API.Port)
-		if err := srv.ListenAndServe(); err != nil {
-			shutdown()
-			log.Fatalln(err)
+	}()
+	defer func() {
+		serverShutdownCtx, serverShutdownCancel := context.WithTimeout(ctx, 10*time.Second)
```
jhagberg: **Shutdown context parent may already be cancelled**

The `ctx` used here is created with `context.WithCancel` at the top of `run()`, and `defer cancel()` is registered before this defer. Since defers execute LIFO, `cancel()` may run before this shutdown defer, making the parent context already cancelled. `context.WithTimeout` on a cancelled parent returns an immediately-cancelled context, giving the server no time to shut down gracefully.

```suggestion
		serverShutdownCtx, serverShutdownCancel := context.WithTimeout(context.Background(), 10*time.Second)
```

sda/cmd/finalize/finalize.go
```diff
@@ -3,265 +3,325 @@
...
+		body, _ := json.Marshal(fileError)
 
-				if err := db.SetAccessionID(message.AccessionID, fileID); err != nil {
-					log.Errorf("failed to set accessionID for file, file-id: %s, reason: %v", fileID, err)
-					if err := delivered.Nack(false, true); err != nil {
-						log.Errorf("failed to Nack message, reason: %v", err)
-					}
+		// Send the message to an error queue so it can be analyzed.
+		if e := mqBroker.SendMessage(fileID, mqBroker.Conf.Exchange, "error", body); e != nil {
+			log.Errorf("failed to publish message, reason: %v", err)
```
jhagberg: **Bug: Logs wrong error variable**

The `SendMessage` error is captured as `e`, but this line logs `err` (from the outer scope). This will log the wrong error or panic if `err` is nil.

```suggestion
			log.Errorf("failed to publish message, reason: %v", e)
```

sda/cmd/s3inbox/proxy.go
```diff
@@ -166,14 +167,22 @@ func (p *Proxy) allowedResponse(w http.ResponseWriter, r *http.Request, token jw
 
 	fileIdentifier := uniqueFileID{
 		username: username,
-		filePath: filepath,
+		filePath: filePath,
 	}
 
+	s3Client, err := newS3Client(r.Context(), p.s3Conf)
```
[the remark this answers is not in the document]
KarlG-nbis: Yeah previously we were creating 3-4 s3 clients per request, and it was updated to create 1 client per request, but yeah makes sense to create one per service instead and rely on its underlying connection pooling, etc to support the concurrent requests

jhagberg (commented):
## Follow-up Review (post force-push)

Previous fixes all confirmed intact after the rebase. The new commits (chart jobs, TLS support, version bump) introduce a few new issues — see inline comments.

### Not placeable as inline comment

**Checksum type mismatch in s3inbox** (`sda/cmd/s3inbox/proxy.go:527`)

`checksum.Type` is set to `"sha256"`, but `requestInfo` (line 556) now returns the raw S3 ETag (MD5 for single-part uploads, composite hash for multipart). On `main`, the ETag was SHA256-hashed before returning (`sha256.Sum256([]byte(ETag))`), so the type was correct. After this PR's change to return raw ETag, the type no longer matches the value. Either update the type to `"md5"` / `"etag"`, or restore the SHA256 hashing.

### Summary

| Severity | Count | Details |
|----------|-------|---------|
| **High** | 1 | `grep -q 0` false-positive in location validation job |
| **Medium** | 2 | Certificate hook weight ordering, s3inbox checksum type mismatch |
| **Low-Medium** | 1 | sda-download file reader never closed |
| **Low** | 2 | `pg_isready` doesn't abort, inconsistent credential handling across jobs |

See inline comments for details.

charts/sda-svc/templates/job-ensure-file-locations-populated.yaml
```diff
@@ -0,0 +1,106 @@
...
+        helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+        release: {{ .Release.Name }}
+    spec:
+      containers:
+      - name: {{ template "sda.fullname" . }}-ensure-file-locations-populated
+        command:
+          - "/bin/sh"
+          - "-c"
+          - |
+            pg_isready
```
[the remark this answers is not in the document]
KarlG-nbis: Updated and added `set -e`

nanjiangshu (approved):

PR #2309, opened 2026-03-12
Fix/workflow tag trigger by GitHub

KarlG-nbis:
# Description
Github ignores events that originates from the `secrets.GITHUB_TOKEN` user, meaning that the push of the tag does not trigger the release of the sda.

In the repo we have setup `GHCR_TOKEN` as a repository secret which will be used for the tagging.

## How to test
- Merge a PR where this is present


jhagberg (approved):
LGTM

PR #2320, opened 2026-03-16
docs(rfc): RFC-0002 explore separating the Visa Authorization Service

jhagberg:
## Related issue(s) and PR(s)

Addresses #2228. Originally opened as ADR-0004; converted to RFC-0002 per
[ADR-0005](https://github.com/neicnordic/sensitive-data-archive/blob/main/docs/decisions/0005-introduce-rfcs-as-upstream-exploration-phase.md)
because the open questions (Olric maturity, latency budget, timing relative
to other download-service work) haven't converged yet.

## Description

Explores extracting GA4GH visa validation from the download service into a
dedicated Visa Authorization Service.

The download service embeds visa validation in its HTTP middleware. Each pod
maintains a per-pod ristretto session cache — when a user's request is
routed to a different pod, the full visa re-validation flow runs again (OIDC
userinfo call + JWKS fetch per visa issuer). This couples authorization
caching to file-streaming scale and generates unnecessary calls to the OIDC
provider.

The RFC explores using [Olric](https://github.com/buraksezer/olric) for
distributed in-process caching across 2+ replicas, making download pods
stateless for authorization. Includes architecture diagrams (current vs
proposed), a GA4GH Passport Clearinghouse gap analysis grounded in the v2
download API implementation (`sda/cmd/download`), and per-option pros/cons.

**Direction currently favoured:** option 1, the separate service with Olric.
Status in the front matter is `exploring`.

## What changed in the ADR → RFC conversion

- Moved file to `docs/rfcs/0002-separate-visa-authorization-service.md`
- `status: proposed` → `status: exploring`; `decision-makers:` → `authors:`
- Removed `## Decision Outcome` and `### Confirmation`
- Added `## Open Questions` (Olric production maturity, latency budget,
  memberlist in NeIC k8s, wait-or-extract, replica/TTL parameters, legacy
  `sda-download` coexistence, naming)
- Added `### Direction currently favoured` in More Information
- Mermaid diagrams, Olric design notes, and Gap Analysis preserved verbatim
- Added a row in `docs/rfcs/README.md` index

## How to test

- Review the RFC at `docs/rfcs/0002-separate-visa-authorization-service.md`
- Verify the Open Questions reflect what is actually unresolved on the team
- Mermaid diagrams render on GitHub
- Promoted to an ADR per ADR-0005's procedure once the Open Questions land

### DIFF
--- a/docs/rfcs/0002-separate-visa-authorization-service.md
+++ b/docs/rfcs/0002-separate-visa-authorization-service.md
@@ -1,0 +1,338 @@
+---
+status: exploring
+date: "2026-05-20"
+authors:
+  - "@neicnordic/sensitive-data-development-collaboration"
+consulted: []
+informed: []
+---
+
+# Separate Visa Authorization Service?
+
+## Context and Problem Statement
+
+The download API v2 service (`sda/cmd/download`) embeds GA4GH visa validation
+directly in its HTTP middleware ([auth.go][v2-auth]). On each request without a
+cached session, the service:
+
+1. Fetches visas from the OIDC userinfo endpoint (or extracts them from the
+   token itself, configurable via `visa.source`)
+   ([validator.go#L104][v2-validator-L104])
+2. For each `ControlledAccessGrants` visa, validates all required claims:
+   `by`, `value`, `source`, `asserted`, and rejects non-empty `conditions`
+   ([validator.go#L393][v2-validator-L393])
+3. Verifies the visa JWT signature by fetching the signing key via `jku` from
+   the trusted issuer allowlist ([jwks_cache.go][v2-jwks])
+4. Checks each visa's dataset value against the local database
+5. Caches results in a per-pod ristretto cache with three tiers: session
+   cookie, token hash, and per-visa validation ([auth.go#L465][v2-auth-L465])
+
+The v2 implementation has significantly improved GA4GH compliance over the
+legacy `sda-download` service (see [Gap Analysis](#gap-analysis) below).
+However, two architectural problems remain:
+
+**Per-pod session loss.** The session cache is local to each pod. When a
+user's request is routed to a different download pod (e.g., after scaling or
+pod restart), the cached session is not available and the full visa
+re-validation flow must run again — involving external HTTP calls to the OIDC
+userinfo endpoint and each visa issuer's JWKS endpoint. This is not an
+interactive re-authentication (the bearer token is still valid), but it adds
+latency and external service load on every pod switch.
+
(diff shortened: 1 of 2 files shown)


pontus (dismissed):
I guess my latency concerns does not directly affect this proposal, although I considered making a suggestion noting `internalTrafficPolicy: Local` and pod co-location as a possible mitigation for the network hop.

pontus (approved):

jbygdell (approved):

viklund:
The way this was explained I thought about it like this:

```mermaid
sequenceDiagram
    User->>+Download: Request file
    Download->>+ClearingHouse: VISA Ok?
    ClearingHouse->>-Download: VISA Ok!
    Download->>-User: Send file
```

But why isn't it like this:

```mermaid
sequenceDiagram
    User->>+API-Gateway: Request file
    API-Gateway->>+ClearingHouse: VISA Ok?
    ClearingHouse->>-API-Gateway: Visa OK!
    API-Gateway->>+Download: Request file
    Download->>-API-Gateway: File
    API-Gateway->>-User: File
```

In practice I imagine that there are a bunch of "middlewares" that a request might go through via some API-gateway before ending up in the download service. This "Visa OK" can be annotated with a more restricted auth token as well, to limit the following request to download to only be valid for the file that was requested.

jhagberg:
@pontus @jbygdell @viklund I have pushed two follow-up commits since you approved this in March, and the file has changed enough that I would rather not merge on the old approvals.

ced8191c folds the thread into the file as the new RFC process asks: the affinity, sidecar and gateway points are now Options 4 and 5 and objections under Option 1, with names and dates. 02a2f78c corrects things I got wrong: the affinity annotation was removed on purpose in c40f9876, the downloadV2 ingress never had it, and the favoured direction has moved away from Olric towards measuring first, with a Postgres backed cache (Option 6) only if the numbers say so.

Could you have a fresh look at "Direction currently favoured" and at your own option? The status stays exploring, and the gateway question is kept as the first open question for the next SDA-Devs meet-up.

jbygdell:
A comment on the gateway option:
* It's only an option once we decide to implement a standalone visa service. 
* We should not try to implement any form of that gateway functionality since `Cilium`, `Envoy` and the like already have it in production grade versions. All we need is a separate header object like `x-gaa4gh-visa` in order to get the routing done correctly.

jbygdell (approved):

jhagberg:
Thanks, folded in (f0dc58a1): gateway only after a standalone service is decided, and then Cilium/Envoy on a visa header, not ours. Only failure semantics and the measurement block promotion now.

PR #2367, opened 2026-03-31
Improvement/add pagination to file listing

pahatz:
When getting the User's file list we get timeouts for queries that return roughly >~80K files. This PR adds cursor-based keyset pagination for `/files` and `/users/:username/files`.

**Database:**
- Denormalizes the latest event name (`last_event TEXT`) onto `sda.files` via an `AFTER INSERT` trigger on `file_event_log`, removing the need to join `file_event_log` in user-file queries
- Reuses the existing `(submission_user, submission_file_path, id)` access pattern for paginated fetches

**API:**
- `GET /users/:username/files` and `GET /files` accept optional `limit`, `cursor`, and `path_prefix` query params
- Returns `X-Next-Cursor` when more pages exist
- Invalid cursors and invalid limits return HTTP 400

**sda-admin:**
- `file list -user <username>` auto-paginates through all pages
- Prompts the user to press Enter/Space to continue or Ctrl+C to abort between pages

**Validator orchestrator:**
- Reuses one HTTP client across paginated SDA API calls
- Stops fetching more pages once all requested file paths have been found
- Preserves deterministic missing-file reporting order


pahatz:
@copilot apply changes based on the comments in [this thread](https://github.com/neicnordic/sensitive-data-archive/pull/2367#pullrequestreview-4036709063)

pahatz:
> > @copilot apply changes based on the comments in [this thread](https://github.com/neicnordic/sensitive-data-archive/pull/2367#pullrequestreview-4036709063)
> 
> Applied all changes from the review thread in commit [a290a9f](https://github.com/neicnordic/sensitive-data-archive/commit/a290a9fcf78fc97d2963a1a8254408957cfa39bb):
> 
>     * **SQL comment** updated to accurately state that `last_event_at` supports keyset pagination ordering, but a lateral join on `file_event_log` is still needed for event status.
> 
>     * **`disabled` filter moved to SQL** (`AND COALESCE(fel.event, '') != 'disabled'`), so `LIMIT` applies only to non-disabled rows.
> 
>     * **`lim+1` probe-row** technique: fetch one extra row to detect `hasMore`; the cursor is based on the `lim`-th row (DB order) and only emitted when a next page genuinely exists.
> 
>     * **`base64.RawURLEncoding`** replaces `base64.StdEncoding` in both encode and decode paths, making cursors safe as URL query parameters without extra client-side escaping.
> 
>     * **`ErrInvalidCursor` sentinel** added in the database package; both `/files` and `/users/:username/files` handlers now return HTTP 400 (not 502/500) for malformed cursors.
> 
>     * **`url.QueryEscape(cursor)`** used in sda-validator when building the request URL.
> 
>     * **`GREATEST(last_event_at, NEW.started_at)`** used in both trigger functions (`02_functions.sql` and `24_add_last_event_at_to_files.sql`) to prevent `last_event_at` from regressing on out-of-order inserts.
> 
>     * **Pagination tests** added: multi-page iteration with duplicate detection, invalid base64 cursor, and valid-base64 but wrong-format cursor.



> > @copilot apply changes based on the comments in [this thread](https://github.com/neicnordic/sensitive-data-archive/pull/2367#pullrequestreview-4036709063)
> 
> Applied all changes from the review thread in commit [a290a9f](https://github.com/neicnordic/sensitive-data-archive/commit/a290a9fcf78fc97d2963a1a8254408957cfa39bb):
> 
>     * **SQL comment** updated to accurately state that `last_event_at` supports keyset pagination ordering, but a lateral join on `file_event_log` is still needed for event status.
> 
>     * **`disabled` filter moved to SQL** (`AND COALESCE(fel.event, '') != 'disabled'`), so `LIMIT` applies only to non-disabled rows.
> 
>     * **`lim+1` probe-row** technique: fetch one extra row to detect `hasMore`; the cursor is based on the `lim`-th row (DB order) and only emitted when a next page genuinely exists.
> 
>     * **`base64.RawURLEncoding`** replaces `base64.StdEncoding` in both encode and decode paths, making cursors safe as URL query parameters without extra client-side escaping.
> 
>     * **`ErrInvalidCursor` sentinel** added in the database package; both `/files` and `/users/:username/files` handlers now return HTTP 400 (not 502/500) for malformed cursors.
> 
>     * **`url.QueryEscape(cursor)`** used in sda-validator when building the request URL.
> 
>     * **`GREATEST(last_event_at, NEW.started_at)`** used in both trigger functions (`02_functions.sql` and `24_add_last_event_at_to_files.sql`) to prevent `last_event_at` from regressing on out-of-order inserts.
> 
>     * **Pagination tests** added: multi-page iteration with duplicate detection, invalid base64 cursor, and valid-base64 but wrong-format cursor.

This came in very handy, but I'll take a closer look first at what it did - apart from changing my description too.

postgresql/migratedb.d/24_add_last_event_at_to_files.sql
```diff
@@ -0,0 +1,50 @@
...
+  sourcever INTEGER := 23;
+  changes VARCHAR := 'Add last_event_at column to files for pagination performance';
+BEGIN
+  IF (select max(version) from sda.dbschema_version) = sourcever then
+    RAISE NOTICE 'Doing migration from schema version % to %', sourcever, sourcever+1;
+    RAISE NOTICE 'Changes: %', changes;
+    INSERT INTO sda.dbschema_version VALUES(sourcever+1, now(), changes);
+
+    -- Add denormalized column
+    ALTER TABLE sda.files ADD COLUMN last_event_at TIMESTAMP WITH TIME ZONE;
```
KarlG-nbis: Wouldnt it be better as sometimes discussed to just add the latest event(eg registered, uploaded, submitted, etc) instead of the timestamp of the latest event as then we wouldnt need the join on the sda.file_event_log table

And we could use created_at, etc for the cursor

sda/internal/database/db_functions.go
```diff
@@ -1030,24 +1040,58 @@ ORDER BY f.id, fel.started_at DESC;`
...
+		parts := strings.SplitN(string(decoded), "|", 2)
+		if len(parts) != 2 {
+			return nil, "", fmt.Errorf("%w: unexpected format", ErrInvalidCursor)
+		}
+		unixnano, perr := strconv.ParseInt(parts[0], 10, 64)
+		if perr != nil {
+			return nil, "", fmt.Errorf("%w: invalid timestamp", ErrInvalidCursor)
+		}
+		cursorTime := time.Unix(0, unixnano)
+		cursorID := parts[1]
```
KarlG-nbis: Why does the cursor need to be based both on the timestamp and id? Wouldnt id be enough if we order by id?

sda/internal/database/db_functions.go
```diff
@@ -1030,24 +1040,58 @@ ORDER BY f.id, fel.started_at DESC;`
 		pathPrefixArg.String = pathPrefix
 	}
 
-	// nolint:rowserrcheck
-	rows, err := db.Query(query, userID, pathPrefixArg, pathPrefixLen)
+	// default limit
+	lim := limit
+	if lim <= 0 {
```
KarlG-nbis, line 1080: Do we silently want to raise the limit to 1000 if not set? Should a user not get a 403, etc if a no limit is requested?

sda/cmd/api/api.go
```diff
@@ -358,14 +359,40 @@ func getFiles(c *gin.Context) {
 		return
 	}
 
-	files, err := Conf.API.DB.GetUserFiles(token.Subject(), c.Query("path_prefix"), false)
+	// parse optional pagination params
+	limit := 0
+	if l := c.DefaultQuery("limit", "0"); l != "0" {
```
KarlG-nbis: limit and cursors query args should be described in the [swagger.yaml](https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/cmd/api/swagger_v1.yml#L275) to ensure its up to date

postgresql/migratedb.d/24_add_last_event_at_to_files.sql
```diff
@@ -0,0 +1,50 @@
...
+
+    -- Backfill from existing event log data
+    UPDATE sda.files AS f
```
KarlG-nbis, line 23: Imo data migration should not be part of these schema migration files, these files should in my opinion just deal with schema migrations

sda-admin/main.go
```diff
@@ -326,7 +326,11 @@ func handleUserCommand() error {
 func handleFileListCommand() error {
 	listFilesCmd := flag.NewFlagSet("list", flag.ExitOnError)
 	var username string
+	var limit int
+	var cursor string
 	listFilesCmd.StringVar(&username, "user", "", "Filter files by username")
+	listFilesCmd.IntVar(&limit, "limit", 0, "Limit number of returned files")
+	listFilesCmd.StringVar(&cursor, "cursor", "", "Cursor for keyset pagination")
```
KarlG-nbis: Should these new flag args be described in the `sda-admin` `README.md`?

KarlG-nbis: Or should these be removed and the sda-admin cmd call does the full pagination and returns the concatenated responses from all required calls?

sda/internal/database/db_functions.go
```diff
@@ -1030,24 +1040,58 @@ ORDER BY f.id, fel.started_at DESC;`
...
+		lim = 1000
+	}
```
jbygdell, line 1082:
```suggestion
	lim := 1000
	if limit > 0 {
		lim = limit
	}
```

1. The default limit should be documented.
2. validation  that the limit is a positive integer is done in the calling code so this snippet can be made cleaner like this.
3. Or it can be removed and we rely on the calling function to ensure that the supplied value is > 0

postgresql/initdb.d/01_main.sql
```diff
@@ -79,6 +80,9 @@ CREATE TABLE files (
     encryption_method    TEXT,
     key_hash             TEXT REFERENCES encryption_keys(key_hash),
 
+    -- Denormalized from file_event_log for pagination performance
+    last_event_at        TIMESTAMP WITH TIME ZONE,
+
```
jbygdell: This table already has a `last_modified` column.
This entry should contain the `event` from the loga table so we don't need to join in that table in the quereies.

sda-admin/main.go
jbygdell: The cursor shouldn't be part of the call.
the cli should handle all the pagination and the only input from the user should be [space] or [enter] until all rows has been displayed. Or that the user aborts teh command [Ctrl + C]

sda-admin/file/file.go
```diff
@@ -22,13 +22,23 @@ type RequestBodyFileAccession struct {
...
 	if err != nil {
 		return err
 	}
 	parsedURL.Path = path.Join(parsedURL.Path, "users", username, "files")
 
+	// Add pagination query params if provided
+	q := parsedURL.Query()
+	if limit > 0 {
+		q.Set("limit", fmt.Sprintf("%d", limit))
+	}
```
jbygdell:
```suggestion
	if limit < 1 {
		return errors.New("limit must be a positive integer")
	}
	q.Set("limit", fmt.Sprintf("%d", limit))
```

sda/cmd/api/api.go
```diff
@@ -358,14 +359,40 @@ func getFiles(c *gin.Context) {
...
+		li, err := strconv.Atoi(l)
+		if err != nil || li < 1 {
+			c.JSON(400, "invalid limit parameter: must be a positive integer")
+
+			return
+		}
+		const maxLimit = 10000
+		if li > maxLimit {
+			li = maxLimit
```
jbygdell: What's the reasoning behind enforcing a max value?

sda/cmd/api/api.go
```diff
@@ -358,14 +359,40 @@ func getFiles(c *gin.Context) {
 		return
 	}
 
-	files, err := Conf.API.DB.GetUserFiles(token.Subject(), c.Query("path_prefix"), false)
+	// parse optional pagination params
+	limit := 0
```
jbygdell:
```suggestion
	limit := 1000
```

Better to initialize the variable with a sane default

sda/cmd/api/api.go
```diff
@@ -991,13 +1018,38 @@ func listUserFiles(c *gin.Context) {
 	username = strings.TrimSuffix(username, "/files")
 	log.Debugln(username)
 
-	files, err := Conf.API.DB.GetUserFiles(username, c.Query("path_prefix"), true)
+	// parse optional pagination params
+	limit := 0
```
jbygdell:
```suggestion
	limit := 1000
```

postgresql/migratedb.d/24_add_last_event_at_to_files.sql
pahatz: Sure, but we should probably discuss where to move all of them at some point.

postgresql/initdb.d/01_main.sql
```diff
@@ -79,6 +80,9 @@ CREATE TABLE files (
     encryption_method    TEXT,
     key_hash             TEXT REFERENCES encryption_keys(key_hash),
 
+    -- Denormalized from file_event_log to avoid join when listing files
+    last_event           TEXT,
```
jbygdell:
```suggestion
    last_event           TEXT REFERENCES file_events(title),
```
Just to be sure we don't have any odd things happening

postgresql/migratedb.d/24_add_last_event_at_to_files.sql
```diff
@@ -0,0 +1,51 @@
...
+  sourcever INTEGER := 23;
+  changes VARCHAR := 'Add last_event column to files to avoid join on file_event_log';
+BEGIN
+  IF (select max(version) from sda.dbschema_version) = sourcever then
+    RAISE NOTICE 'Doing migration from schema version % to %', sourcever, sourcever+1;
+    RAISE NOTICE 'Changes: %', changes;
+    INSERT INTO sda.dbschema_version VALUES(sourcever+1, now(), changes);
+
+    -- Add denormalized column storing the latest event name (e.g. 'registered', 'uploaded')
+    ALTER TABLE sda.files ADD COLUMN last_event TEXT;
```
jbygdell:
```suggestion
    ALTER TABLE sda.files ADD COLUMN last_event TEXT REFERENCES file_events(title);
```

sda/cmd/api/api.go
pahatz: This was for the case the user sends a big limit on the API request. I 'll change it to return a 400 to the user.

sda/internal/database/db_functions.go
pahatz: will change it. I guess this was for returning them sorted but we don't really need that.

postgresql/migratedb.d/24_add_last_event_to_files.sql
```diff
@@ -0,0 +1,51 @@
...
+    -- Grant UPDATE on last_event to roles that insert into file_event_log
+    -- but previously only had SELECT on sda.files (e.g. api, mapper)
+    GRANT UPDATE (last_event) ON sda.files TO api;
+    GRANT UPDATE (last_event) ON sda.files TO mapper;
+
+  ELSE
+    RAISE NOTICE 'Schema migration from % to % does not apply now, skipping', sourcever, sourcever+1;
+  END IF;
+END
+$$
```
jbygdell, line 64: duplicated file?

postgresql/migratedb.d/24_add_last_event_at_to_files.sql
```diff
@@ -0,0 +1,51 @@
...
+
+    CREATE TRIGGER file_event_log_update_last_event
+        AFTER INSERT ON sda.file_event_log
+        FOR EACH ROW
+        EXECUTE PROCEDURE sda.update_files_last_event();
+
+    -- Grant UPDATE on last_event to roles that insert into file_event_log
+    -- but previously only had SELECT on sda.files (e.g. api, mapper)
+    GRANT UPDATE (last_event) ON sda.files TO api;
+    GRANT UPDATE (last_event) ON sda.files TO mapper;
```
jbygdell: this should be kept out of this file. Permissions is not part of the schema migration.

postgresql/initdb.d/01_main.sql
```diff
@@ -161,6 +165,9 @@ VALUES ( 5, 'registered'  , 'Upload to the inbox has started'),
        ( 1, 'disabled'    , 'Disables the file for all actions'),
        ( 2, 'enabled'     , 'Reenables a disabled file');
 
+-- Add FK now that file_events exists
+ALTER TABLE files ADD CONSTRAINT files_last_event_fk FOREIGN KEY (last_event) REFERENCES file_events(title);
+
```
jbygdell: The proper way to do this is to move lines 146-166 before the creation of the files table on line 61

pahatz: Right! let me tidy up

jhagberg (commented):
A few follow-up observations from a fresh read, mostly around the new trigger and the validator loop. Nothing blocking here; happy to defer any of these.

postgresql/initdb.d/02_functions.sql
```diff
@@ -16,6 +16,20 @@ CREATE TRIGGER files_last_modified
     FOR EACH ROW
     EXECUTE PROCEDURE files_updated();
 
+-- Keep files.last_event in sync with file_event_log inserts
+CREATE FUNCTION update_files_last_event()
```
jhagberg: The migration variant of this function (in `24_add_last_event_to_files.sql`) declares `SECURITY DEFINER` with a pinned `search_path = pg_catalog, sda`, but the init variant here runs as the invoking role with no pinned path. Today it works because every role that inserts into `file_event_log` also holds UPDATE on `files`, so the boundary does not bite. The worry is future divergence: if a role ever lands with only INSERT on the event log, fresh DBs would fail the trigger while migrated DBs would keep working. Might be worth aligning the two definitions so there is only one shape to reason about.

sda-validator/orchestrator/api/api.go
```diff
@@ -286,62 +287,88 @@ type getUserFilesResponse struct {
...
 	}
-	defer res.Body.Close()
 
-	// Check the status code
-	if res.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("server returned status %d: url: %s", res.StatusCode, req.URL.String())
-	}
+	cursor := ""
+	client := &http.Client{} // reuse across pages for connection pooling
+	for {
```
jhagberg, line 303: One safety thought: if the SDA API ever returns the same non-empty `X-Next-Cursor` on every response (bug, misconfig, replayed cursor), this loop has nothing that would make it stop. A small cap would catch that class of failure loudly. Something like a max page count, or exiting when the cursor value has not advanced.

sda-validator/orchestrator/api/api_test.go
```diff
@@ -38,7 +38,7 @@ func (ts *ValidatorAPITestSuite) SetupSuite() {
 	ts.mockBroker = &mockBroker{}
 
 	ts.httpTestServer = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
```
jhagberg, line 41: The mock server returns all five files in one response and never sets `X-Next-Cursor`, so the pagination loop only runs once. That leaves the multi-page path and the `len(needed) == 0` early-exit untested. Could be worth adding two cases: one where page 1 returns a cursor plus a partial subset (exercises pagination), and one where page 1 contains every requested path but still carries a cursor (exercises the early-exit).

sda/internal/database/db_functions_test.go
```diff
@@ -688,9 +690,45 @@ func (suite *DatabaseTests) TestGetUserFiles() {
 		assert.Contains(suite.T(), fileInfo.AccessionID, "stableID-00", "incorrect file accession ID")
 	}
 
-	filteredFilelist, err := db.GetUserFiles(testUser, fmt.Sprintf("%s/submission_b", testUser), true)
+	filteredFilelist, nextCursor, err := db.GetUserFiles(testUser, fmt.Sprintf("%s/submission_b", testUser), true, 0, "")
 	assert.NoError(suite.T(), err, "failed to get file list")
 	assert.Equal(suite.T(), 3, len(filteredFilelist), "file list is of incorrect length")
+	assert.Equal(suite.T(), "", nextCursor)
+
+	// Verify keyset pagination: use a page size smaller than the total to force multiple pages.
```
jhagberg, line 743: I could not find a direct assertion that the new trigger does its job: that an insert into `file_event_log` sets `files.last_event`, and that a file with `last_event = 'disabled'` is excluded from `GetUserFiles`. Since the pagination filter now depends on that denormalization, a short test would pin both behaviors in place.

postgresql/initdb.d/02_functions.sql
```diff
@@ -16,6 +16,20 @@ CREATE TRIGGER files_last_modified
...
+RETURNS TRIGGER AS $update_files_last_event$
+BEGIN
+    UPDATE sda.files SET last_event = NEW.event WHERE id = NEW.file_id;
+    RETURN NEW;
+END;
+$update_files_last_event$ LANGUAGE plpgsql;
+
+CREATE TRIGGER file_event_log_update_last_event
```
KarlG-nbis, line 40: my worry is that if we rely on such a trigger and execute the file_event_log update within a transaction which is then rollbacked, this wouldnt be rolled back and the sda.files last_event would be incorrect

pahatz: Trigger update is part of the same transaction as the `file_event_log` insert, so it is rolled back with the transaction on postgres.
I am adding a test.

postgresql/initdb.d/02_functions.sql
pahatz: I addressed wthat looked important, please take another look.

sda/cmd/api/api.go
```diff
@@ -346,6 +347,32 @@ func rbac(e *casbin.Enforcer) gin.HandlerFunc {
...
+// returns the effective limit. On validation failure it writes a 400 response
+// via c and returns -1 so the caller can return early.
+func parseLimitParam(c *gin.Context) int {
+	l := c.DefaultQuery("limit", "0")
+	if l == "0" {
+		return defaultPageLimit
+	}
+	li, err := strconv.Atoi(l)
+	if err != nil || li < 1 {
+		c.AbortWithStatusJSON(http.StatusBadRequest, "invalid limit parameter: must be a positive integer")
```
KarlG-nbis: minor: personally I would prefer if the http response was not written within this func, as it is somewhat hidden imo, and not super clear that from the func name `parseLimitParam` it would do that(even thou it mention it in the comment), and instead it could return an the limit, and error, eg:

```go
limit, err := parseLimitParam(c.DefaultQuery("limit", "0"))
if err != nil {
   c.AbortWithStatusJSON(http.StatusBadRequest, err.Error())

  return
}
``` 

as imo the possible responses are more clear if they are all detailed on the `getFiles` level, but feel free to disagree and disregard comment

sda/internal/database/db_functions.go
```diff
@@ -1030,24 +1039,54 @@ ORDER BY f.id, fel.started_at DESC;`
...
+		decoded, derr := base64.RawURLEncoding.DecodeString(cursor)
+		if derr != nil {
+			return nil, "", fmt.Errorf("%w: %v", ErrInvalidCursor, derr)
+		}
+		parts := strings.SplitN(string(decoded), "|", 2)
+		if len(parts) != 2 {
+			return nil, "", fmt.Errorf("%w: unexpected format", ErrInvalidCursor)
+		}
+		cursorID := parts[0]
+		cursorPath := parts[1]
```
KarlG-nbis: Why does the cursor need to consist of both id and path? Wouldnt id be enough?

KarlG-nbis: As if we dont need path we wouldnt need to construct the query differently depending if we have cursor or not, as we could just have a check cursor arg and then include everything if not provided: eg

```sql
SELECT f.id, f.submission_file_path, f.stable_id, COALESCE(f.last_event, '') as event, f.created_at, f.submission_file_size
FROM sda.files AS f
	LEFT JOIN sda.file_dataset AS fd ON fd.file_id = f.id
WHERE f.submission_user = $1 AND ($2::TEXT IS NULL OR substr(f.submission_file_path, 1, $3) = $2::TEXT)
  AND fd.file_id IS NULL AND COALESCE(f.last_event, '') != 'disabled'
  AND ($4::TEXT IS NULL OR f.id > $4) 
ORDER BY f.id ASC LIMIT $5;"
```
eg:

```go

cursorArg := sql.NullString{}
if cursor != "" {
  cursorArg.Valid = true
  cursorArg.String, err = base64.RawURLEncoding.DecodeString(cursor)
  if err != nil {
    return err
  }
}

rows, err = db.Query(query, userID, pathPrefixArg, pathPrefixLen, cursorArg, fetchLim)
```

kostas-kou (commented):
I do not have any additional comments. When the last comments are resolved we can merge it I think. Ping me for the final check

sda/internal/database/db_functions.go
```diff
@@ -988,40 +993,45 @@ func (dbs *SDAdb) setSubmissionFileSize(fileID string, submissionFileSize int64)
...
+	// works for both the first page (cursorArg IS NULL) and subsequent pages (f.id > cursorArg).
 	const query = `
-SELECT DISTINCT ON (f.id) f.id, f.submission_file_path, f.stable_id, fel.event, f.created_at, f.submission_file_size FROM sda.files AS f
-    LEFT JOIN sda.file_event_log AS fel ON fel.file_id = f.id
-    LEFT JOIN sda.file_dataset AS fd ON fd.file_id = f.id
-WHERE f.submission_user = $1 AND ($2::TEXT IS NULL OR substr(f.submission_file_path, 1, $3) = $2::TEXT)
-    AND fd.file_id IS NULL
-ORDER BY f.id, fel.started_at DESC;`
-
+SELECT f.id, f.submission_file_path, f.stable_id, COALESCE(f.last_event, '') as event, f.created_at, f.submission_file_size
```
jhagberg, line 1061: One thing to flag: `getVersion()` is fetched at startup (`database.go:117`), but I can't find anywhere it's compared against a required minimum. If the API ever starts against a database still at schema v23, this query succeeds at startup and only fails at request time with `column "last_event" does not exist`. That is a harder failure to diagnose than a startup refusal. Could be worth a min-version check on `dbs.Version` after `NewSDAdb()`, either gating startup or just the paginated handlers.

postgresql/migratedb.d/24_add_last_event_to_files.sql
```diff
@@ -0,0 +1,53 @@
...
+  IF (select max(version) from sda.dbschema_version) = sourcever then
+    RAISE NOTICE 'Doing migration from schema version % to %', sourcever, sourcever+1;
+    RAISE NOTICE 'Changes: %', changes;
+    INSERT INTO sda.dbschema_version VALUES(sourcever+1, now(), changes);
+
+    -- Add denormalized column storing the latest event name (e.g. 'registered', 'uploaded')
+    ALTER TABLE sda.files ADD COLUMN last_event TEXT REFERENCES sda.file_events(title);
+
+    -- Backfill from existing event log data
+    UPDATE sda.files AS f
```
jhagberg, line 23: The `files_updated()` trigger at `02_functions.sql:5-12` fires on `BEFORE UPDATE` and rewrites `last_modified` and `last_modified_by`. This backfill UPDATEs every existing row, so on upgrade every archived file gets those audit fields overwritten with the migration time and the `migrate` user. If `last_modified` surfaces downstream (I believe sda-download exposes it as the S3 Last-Modified header), long-untouched files would appear modified at migration time. Two ways out: disable the trigger around the backfill (`ALTER TABLE sda.files DISABLE TRIGGER files_last_modified;` then re-enable), or preserve the values inline with `SET last_event = sub.event, last_modified = f.last_modified, last_modified_by = f.last_modified_by`.

pahatz: added

sda/internal/database/db_functions.go
pahatz: I am not sure we 're doing such checks for similar cases in the past and mostly rely on the admin doing the migration correctly. These checks might also start piling up. It might be better to create an issue and handle these in a better way in general.

sda/cmd/api/api.md
```diff
@@ -21,8 +21,23 @@ Endpoints:
 
     It is possible to limit the returned results by supplying a base path prefix in the query. In this case only files that have a path that starts with `submission-1` will be returned.
 
+    The endpoint also supports keyset pagination using a `cursor` query parameter and an optional `limit`.
+
+    - `limit` (optional): maximum number of items per page. If omitted, the API defaults to `1000`. The value must be a positive integer; non-positive values such as `0` are rejected with `400 Bad Request`.
```
KarlG-nbis: minor: could be worth mentioning that the maxPageSize limit is 10000 here

postgresql/initdb.d/01_main.sql
```diff
@@ -89,6 +115,7 @@ CREATE TABLE files (
 );
 -- Add indexes to the files table
 CREATE INDEX files_submission_user_submission_file_path_idx ON files(submission_user, submission_file_path);
+CREATE INDEX files_submission_user_id_idx ON files(submission_user, id);
```
KarlG-nbis: I do not believe this index is needed, as we will have index on `submission_user` from `files_submission_user_submission_file_path_idx` (as it is the first column specified in the multi column index), and we will have index on `id` (as it is the primary key)

KarlG-nbis (approved):

zeidlitz (approved):

PR #2381, opened 2026-04-10
feat(broker): broker v2

### LINKED ISSUE #1959: [broker] Reduce cognitive complexity and amount of code when consuming from rabbitmq
A lot of boiler plate code is currently needed when consuming message from rabbitmq, with a lot of indentations causing a lot of cognitive complexity eg
```go
go func() {
  messages, err := app.MQ.GetMessages(app.Conf.Broker.Queue)
  if err != nil {
    log.Panic(err)
  }
  for delivered := range messages {
    ...
  }
}
```

with some refactor of the [GetMessages](https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/internal/broker/broker.go#L176) we could reduce this code and generalise the ack / nack handling

For example if we refactor GetMessages -> 
```go
// Subscribe reads messages from the queue and handles them with the handleFunc
func (broker *AMQPBroker) Subscribe(ctx context.Context, queue string, handleFunc func(amqp.Delivery) error) error {
	messageChan, err := broker.Channel.Consume(
		queue, // queue
		"",    // consumer
		false, // auto-ack
		false, // exclusive
		false, // no-local
		false, // no-wait
		nil,   // args
	)
	if err != nil {
		return err
	}
	
	for {
		select {
		case message, ok := <-messageChan:
			if !ok {
				log.Debugf("channel closed")
				return nil
			}
			if err := handleFunc(message); err != nil {
				if err := message.Nack(false, false); err != nil {
					log.Errorf("failed to nack message, reason: %v", err)
				}
				continue
			}
			if err := message.Ack(false); err != nil {
				log.Errorf("failed to ack message, reason: %v", err)
			}
		case <-ctx.Done():
			// Here we cancel the consumer and let loop continue until messageChan has closed
			if err := broker.Channel.Cancel("", true); err != nil {
				log.Errorf("failed to cancel channel, reason: %v", err)
			}
		}
	}
}
```

then instead in each service what we would need is
```go
go func {
  // passing ctx such that we can gracefully shut down the subscribe 
  if err := app.MQ.Subscribe(ctx, app.Conf.Broker.Queue, handleFunc); err != nil {
    // handle err  
  }
}

...

func handleFunc(delivery amqp.Delivery) error {
  ... // Handle the message 
}
```

and the handle func would not need to deal with acking / nacking of messages, if the handleFunc returns `err != nil` message will be nacked and `err == nil` then message is acked.

zeidlitz:
## Related issue(s) and PR(s)
This PR closes #1959

## Description
this change introduces a new v2 broker package that aims to improve
message handling and to reduce the complexity of message handling for
other services using the broker package. 

key points beeing addressed:

  - introducing a broker interface 

  - introducing a rabbitmq implementation that can deal with Ack / Nack
    logic based on return type and callback functions that can execute
    regardless of Ack / Nack

  - sepperate channels for consume & publish (best practice according to
    rabbitmq documentation)

  - adding broker v2 to ingest.go to reduce complexity

  - refactoring of ingest.go

  - decrypt c4gh files using a set of two tee readers to get rid of a
    lot of []bytes and buffer overhead, reducing ~100 LOC

  - refactoring of ingest_test.go

  - using mock messages in ingest_test.go instead of a container running
    a rabbitmq implementation

  - removing read / writes to disks / tmp folders in ingest_test.go and
    store the data in memory where possible


sda/internal/broker/v2/broker.go
```diff
@@ -0,0 +1,19 @@
+package v2
+
+import "context"
+
+type Broker interface {
+	// Subscribe starts subscribing to the specified sourceQueue through the consumerGroup and handling each incoming message with the handleFunc
+	// if handleFunc returns an error the message should be nacked and marked for reconsumption
+	// handleFunc supports returning a slice of callback functions that will be invocated after the message has either been acknowledged or not acknowledged
+	Subscribe(ctx context.Context, consumerGroup, sourceQueue string, handleFunc func(ctx context.Context, msg *Message) ([]func(), error)) error
```
KarlG-nbis: consumerGroup arg not needed as not used atm

sda/internal/broker/v2/rabbitmq/rabbitmq.go
```diff
@@ -0,0 +1,264 @@
...
+		if done := b.consumeMessages(ctx, messageChan, handleFunc); done {
+			return ctx.Err()
+		}
+	}
+}
+
+func (b *rmqBroker) Publish(ctx context.Context, destinationQueue string, message broker.Message) error {
+	b.mu.Lock()
+	ch := b.publishChannel
+	defer b.mu.Unlock()
```
KarlG-nbis: Whats the reasoning for the locking? why can we not publish messages concurrently?

sda/internal/broker/v2/rabbitmq/rabbitmq.go
```diff
@@ -0,0 +1,264 @@
...
+	return ch.Consume(sourceQueue, "", autoAck, excluseive, noLocal, noWait, nil)
+}
+
+func (b *rmqBroker) consumeMessages(ctx context.Context, messageChan <-chan amqp.Delivery, handleFunc func(context.Context, *broker.Message) ([]func(), error)) bool {
+	for {
+		select {
+		case <-ctx.Done():
+			b.mu.Lock()
+			if b.consumeChannel != nil {
+				b.consumeChannel.Cancel("", true)
```
KarlG-nbis: unhandled error from `b.consumeChannel.Cancel`, if we are to ignore it better to assign it to a blank identifier to be explicit that we are ignoring it

sda/internal/broker/v2/rabbitmq/rabbitmq.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	callbacks, err := handleFunc(ctx, msg)
+	for _, cb := range callbacks {
+		cb()
+	}
+
+	if err != nil {
+		log.Errorf("error handling message %s: %v", msg.Key, err)
+		delivery.Nack(false, false)
```
KarlG-nbis: Would it not be better to requeue all messages with an err?

sda/internal/broker/v2/rabbitmq/rabbitmq.go
```diff
@@ -0,0 +1,264 @@
...
+
+func (b *rmqBroker) ensureConnected() error {
+	if b.Alive() {
+		return nil
+	}
+	log.Warn("broker connection dead, attempting to reconnect")
+	for {
+		if err := b.connect(); err != nil {
+			log.Error("faield to connect, reason: %v", err)
+			time.Sleep(time.Duration(b.config.timeout))
```
KarlG-nbis: Here we could get stuck forever trying to connect again, possibly worth having some backoff policy with max retries, etc

sda/cmd/ingest/ingest.go
```diff
@@ -232,444 +194,230 @@ func (app *Ingest) registerC4GHKey() error {
 	return nil
 }
 
-func (app *Ingest) cancelFile(ctx context.Context, fileID string, message schema.IngestionTrigger) string {
-	m, _ := json.Marshal(message)
-
-	// Check if file can be cancelled
-	inDataset, err := app.DB.IsFileInDataset(ctx, fileID)
+func (app *Ingest) cancelFile(ctx context.Context, fileID string, message *v2.Message) ([]func(), error) {
```
KarlG-nbis: not much sense for 'cancelFile` func to return `[]func()` if it just always return it as nil

sda/cmd/ingest/ingest.go
```diff
@@ -232,444 +194,230 @@ func (app *Ingest) registerC4GHKey() error {
...
-				OriginalMessage: message,
-			}); err != nil {
-				return "reject"
-			}
+	case "uploaded", "disabled":
+		break
 
-			return "ack"
+	case "":
+		fileID, err = app.DB.RegisterFile(&fileID, submissionLocation, filePath, user)
```
KarlG-nbis: `submissionLocation` would be `""` here if a file was uploaded through means which did not already register it in the database, the idea of `app.InboxReader.FindFile` was to find in which location such files are such that we can register them with it

sda/internal/database/database_interface.go
```diff
@@ -12,5 +12,5 @@ type Database interface {
 	GetMappingData(accessionID string) (*MappingData, error)
 	GetSubmissionLocation(ctx context.Context, fileID string) (string, error)
 	CancelFile(ctx context.Context, fileID, message string) error
-	IsFileInDataset(ctx context.Context, fileID string) (bool, error)
+	IsFileInDataset(ctx context.Context, fileID string) (string, error)
```
KarlG-nbis: Whats the reasoning for this change? As if the func is now to return the datasetID a file belongs to the name of the func should be updated, and also in the future if a file is in multiple dataset we should probably return all the ids, etc

zeidlitz: Ah right. So I just thought It made much more intuitive sense for a db function called IsFileInDataset to also return the dataset id that file is in. This is not really needed now, I remember thinking that I might want to use the datasetID for something down the line in the cancel logic, but that was not the case.

sda/cmd/ingest/ingest.go
zeidlitz: Got it! Will adjust back.

sda/cmd/ingest/ingest.go
zeidlitz: Agreed.

sda/internal/broker/v2/rabbitmq/rabbitmq.go
zeidlitz: Agreed. Will change.

sda/internal/broker/v2/rabbitmq/rabbitmq.go
zeidlitz: Sure! Makes sense to me, I believe this was the 'old' logic or similar to how it was before? I was just thinking to have any Nack / Ack stuff consistent with the previous behavior.

sda/internal/broker/v2/rabbitmq/rabbitmq.go
zeidlitz: Agreed.

sda/internal/broker/v2/broker.go
zeidlitz: Yes, this is a leftover, will adjust.

sda/internal/broker/v2/rabbitmq/rabbitmq.go
zeidlitz: Ah this could be a leftover from when there was one channel for both publishing and consuming and I wanted to make sure no race conditions could occur here. Agreed, should be safe to publish concurrently

sda/cmd/ingest/config/config.go
```diff
@@ -0,0 +1,60 @@
...
+var (
+	sourceQueue   string
+	archivedQueue string
+	schemaPath    string
+)
+
+func init() {
+	config.RegisterFlags(
+		&config.Flag{
+			Name: "ingest.sourceQueue",
```
jhagberg: Ran `make integrationtest-sda-s3` on this branch and the `ingest` container CrashLoops with:

```
failed to load config, due to: missing required flag: ingest.sourceQueue
missing required flag: ingest.archiveQueue
missing required flag: ingest.schemaPath
```

The three flags are registered as `Required: true` with empty defaults. `.github/integration/sda/config.yaml` doesn't set them, so the existing test stack can't start the service. Helm values under `charts/sda-svc/` are missing them too.

Two reasonable fixes: give the flags sensible defaults matching what main used (queues `ingest`/`archived`, schema path derived from `schema.type`), or add the config sections to both the integration yaml and the chart in this PR. I patched the integration config locally with `ingest.sourceQueue: "ingest"`, `ingest.archiveQueue: "archived"`, `ingest.schemaPath: "/schemas/isolated/"` and ingest then booted. Whichever direction you pick, please also update the chart so the deployment side doesn't regress.

sda/internal/broker/v2/rabbitmq/rabbitmq.go
```diff
@@ -0,0 +1,291 @@
...
+	ch := b.publishChannel
+	confirmChan := b.publishConfirmChan
+
+	if ch == nil {
+		return fmt.Errorf("cannot publish: broker channel is not initialized")
+	}
+
+	err := ch.PublishWithContext(
+		ctx,
+		"",
```
jhagberg: Line 75 passes `""` as the exchange to `PublishWithContext`, which routes through `amq.default`. The SDA rabbit users (`ingest`, `verify`, `finalize`, ...) only have write permissions on the named `sda` exchange. After patching past the config crash above, every publish fails in the integration run with:

```
Exception (403) Reason: "ACCESS_REFUSED - write access to exchange 'amq.default' in vhost 'sda' refused for user 'ingest'"
```

followed by `channel/connection is not open` on the retries. The `archived` message never leaves ingest, verify starves, the pipeline stalls.

`broker.exchange` is registered in `rabbitmq/config.go:89-95`, stored on `options.exchange`, propagated through `clone()`, and then never read. Main's `SendMessage(corrID, exchange, routingKey, body)` took the exchange as a parameter and callers passed the configured value. Looks like the config plumbing is all there, just needs to land on the `Publish` call.

sda/cmd/ingest/ingest.go
```diff
@@ -232,444 +194,230 @@ func (app *Ingest) registerC4GHKey() error {
...
-				Reason:          findFileErr.Error(),
-				OriginalMessage: message,
-			}); err != nil {
-				return "reject"
-			}
+	case "uploaded", "disabled":
+		break
 
-			return "ack"
+	case "":
```
jhagberg, line 296: For files not already in `sda.files`, `GetSubmissionLocation` returns `"", nil` on `sql.ErrNoRows` (`db_functions.go:806-807`). The switch then enters `case ""`, calls `RegisterFile`, and returns `nil, nil`, which the broker treats as an ack. The file never reaches `ArchiveWriter.WriteFile`.

Main had an `InboxReader.FindFile(ctx, message.FilePath)` step in the same branch (`main:317-338`) that discovered the submission location across the configured storages, then fell through the switch into `NewFileReader` and the rest of the ingest path in the same delivery. So any deployment that relies on ingest to pick up uploads not pre-registered by `s3inbox`/`sftpInbox` silently loses data with this PR.

Is the intent that every caller must pre-register, or is the `FindFile` fallback something you want to keep? If the former, it probably deserves a short changelog entry since it's a behaviour change.

sda/cmd/ingest/ingest.go
```diff
@@ -232,444 +194,230 @@ func (app *Ingest) registerC4GHKey() error {
...
-		log.Errorf("failed to set error status for file from message, file-id: %s, reason: %s", fileID, err.Error())
-	}
-	body, _ := json.Marshal(infoError)
-	if err := app.MQ.SendMessage(fileID, app.MQ.Conf.Exchange, "error", body); err != nil {
-		log.Errorf("failed to publish message, reason: %v", err)
-
-		return err
+func (app *Ingest) deadLetterQueue(message *v2.Message) func() {
+	return func() {
+		if err := app.MQ.Publish(context.TODO(), "error", *message); err != nil {
```
jhagberg: The callback publishes the raw ingestion trigger: `app.MQ.Publish(context.TODO(), "error", *message)`. `cmd/notify/notify.go:92-94` still parses error-queue messages as `broker.InfoError` and base64-decodes `original-message`. I put together a small standalone repro (marshal the new payload, unmarshal via notify's parser) and got back `error=""`, `reason=""`, `user=""`. Main wrapped the payload via `setFileEventErrorAndSendToErrorQueue` which marshalled `broker.InfoError{Error, Reason, OriginalMessage}` first.

Am I missing an adapter somewhere, or is `notify` knowingly out-of-scope for this PR?

sda/cmd/ingest/ingest.go
```diff
@@ -121,98 +129,52 @@ func run() error {
...
-	messages, err := app.MQ.GetMessages(app.MQ.Conf.Queue)
-	if err != nil {
-		return err
-	}
-
-	for delivered := range messages {
-		app.handleMessage(ctx, delivered)
-	}
+	go func() {
+		err := app.MQ.Subscribe(ctx, sourceQueue, app.handleMessage)
```
jhagberg: `Subscribe` runs in a goroutine whose error is only logged. `run()` then blocks on `<-sigc` alone. If the consumer exits because the broker channel or connection closed, the process stays up but does nothing. Healthcheck happy, queue growing. Main's `run()` had a 4-way select over `consumeErr`, `Connection.NotifyClose`, `Channel.NotifyClose`, and the signal (`main:124-142`), so broker death surfaced as an exit.

Either plumb an error channel from the goroutine back into `run()`, or cancel `ctx` from inside the goroutine on fatal errors so the outer select unblocks.

sda/cmd/ingest/ingest.go
jhagberg: Main emitted `UpdateFileEventLog(fileID, "submitted", "ingest", "{}", string(m))` right after `RegisterFile` (`main:388`). I can't find an equivalent call anywhere in the new flow. Audit queries and anything watching `sda.file_event_log` for the `submitted` transition will stop seeing it.

Dropped by accident, or intentional?

sda/internal/broker/v2/broker.go
```diff
@@ -0,0 +1,19 @@
+package v2
+
+import "context"
+
+type Broker interface {
+	// Subscribe starts subscribing to the specified sourceQueue through the consumerGroup and handling each incoming message with the handleFunc
```
jhagberg, line 6: Two lines of the doc disagree with `handleDelivery` in `rabbitmq.go:215-226`:

- "callbacks that will be invocated after the message has either been acknowledged or not acknowledged". Callbacks run *before* `Ack`/`Nack`, not after.
- "the message should be nacked and marked for reconsumption". `Nack(false, false)` has `requeue=false`, so the message drops to DLX or is discarded, not redelivered.

Every caller building on this contract will get the opposite behaviour from what the doc promises. Whichever way the real intent lands, worth tightening before more callers come online.

jhagberg (changes requested):

jhagberg (changes requested):
Had another pass at this. The `decryptAndArchive` rewrite is a genuine win. The `TeeReader` approach drops ~150 lines of goroutine pipeline without losing behaviour. Same for the interface-first shape of the new broker package.

That said, I ran `make integrationtest-sda-s3` against this branch and a few things don't boot. In priority order before merge:

1. `ingest` crashes on startup against the existing integration config. The new `broker v2` config adds three `Required: true` flags (`ingest.sourceQueue`, `ingest.archiveQueue`, `ingest.schemaPath`) with no defaults, and neither `.github/integration/sda/config.yaml` nor the helm chart under `charts/sda-svc/` provides them. Any existing deployment that merges this PR will lose ingest on the next restart.
2. Once you pass config, every publish fails with `ACCESS_REFUSED - write access to exchange 'amq.default'`. `rabbitmq.go:75` hardcodes `""` as the exchange, while the SDA rabbit users only have write perms on the named `sda` exchange. The `broker.exchange` flag is loaded and cloned but never read.
3. `cmd/ingest/ingest_test.go` doesn't compile against the new signatures (`go vet ./cmd/ingest` fails with 10+ type errors). CI stays red until this is ported.
4. A couple of older behaviours look like silent drops: unregistered files that used to fall through `InboxReader.FindFile` now get registered and acked without being ingested, and the DLQ payload format changed in a way that `cmd/notify` can't parse anymore. Happy to be wrong about either. Details inline.

No new tests on the `broker/v2` package itself, which is the same surface we'd want covered after these fixes.

Rest of the nits inline.

sda/cmd/ingest/config/config.go
zeidlitz: Will do! :+1:

sda/internal/broker/v2/rabbitmq/rabbitmq.go
zeidlitz: Ah alright, yes this is a oversight, will adjust

sda/cmd/ingest/ingest.go
zeidlitz: This is a oversight, there should be no change in the logic in this case. I wrongly just assumed FindFile would return the same as GetSubmissionLocation. Will change.

sda/cmd/ingest/ingest.go
zeidlitz: Hm.. sounds like I might have simplified too much when trying to break up the setFileEventErrorAndSendToErrorQueue logic. Will take a look at it again.

sda/cmd/ingest/ingest.go
zeidlitz: Will adjust :+1:

sda/cmd/ingest/ingest.go
zeidlitz: Not intentional. Will take a look at it :+1:

sda/internal/broker/v2/broker.go
zeidlitz: Good input! What is the "doc" you are referring to in this context?

zeidlitz:
> Had another pass at this. The `decryptAndArchive` rewrite is a genuine win. The `TeeReader` approach drops ~150 lines of goroutine pipeline without losing behaviour. Same for the interface-first shape of the new broker package.
> 
> That said, I ran `make integrationtest-sda-s3` against this branch and a few things don't boot. In priority order before merge:
> 
> 1. `ingest` crashes on startup against the existing integration config. The new `broker v2` config adds three `Required: true` flags (`ingest.sourceQueue`, `ingest.archiveQueue`, `ingest.schemaPath`) with no defaults, and neither `.github/integration/sda/config.yaml` nor the helm chart under `charts/sda-svc/` provides them. Any existing deployment that merges this PR will lose ingest on the next restart.
> 2. Once you pass config, every publish fails with `ACCESS_REFUSED - write access to exchange 'amq.default'`. `rabbitmq.go:75` hardcodes `""` as the exchange, while the SDA rabbit users only have write perms on the named `sda` exchange. The `broker.exchange` flag is loaded and cloned but never read.
> 3. `cmd/ingest/ingest_test.go` doesn't compile against the new signatures (`go vet ./cmd/ingest` fails with 10+ type errors). CI stays red until this is ported.
> 4. A couple of older behaviours look like silent drops: unregistered files that used to fall through `InboxReader.FindFile` now get registered and acked without being ingested, and the DLQ payload format changed in a way that `cmd/notify` can't parse anymore. Happy to be wrong about either. Details inline.
> 
> No new tests on the `broker/v2` package itself, which is the same surface we'd want covered after these fixes.
> 
> Rest of the nits inline.

Appreciate the input! Very constructive and detailed. I knowingly published some of the changes a bit too early so I could get early feedback on the new structure / style changes, so a lot of the broken stuff will be fixed. Tests will also be included in the final change.

sda/internal/broker/v2/broker.go
jhagberg: Ah sorry, meant the godoc comments just above the Subscribe signature, lines 7-8:

```go
// if handleFunc returns an error the message should be nacked and marked for reconsumption
// handleFunc supports returning a slice of callback functions that will be invocated after the message has either been acknowledged or not acknowledged
```

Two things don't match `handleDelivery`: callbacks run before the ack/nack (not after), and `Nack(false, false)` drops rather than requeues. Someone going by the godoc would expect the opposite behaviour.

zeidlitz: Ah yes, I see it! Yes the godoc is correct and I will adjust the function to reflect it :+1:

sda/cmd/ingest/ingest.go
zeidlitz: so from what I could find, only the notify.go service seems to expect to consume from this error queue? Or is there any other service that is dependent on these error messages that I am missing? 

Seeing that the notify.go service is incomplete and not something we are using (as far as I can tell) I would not worry to much about it for now.

sda/internal/broker/v2/rabbitmq/rabbitmq.go
jbygdell: The current logic is to a large extend based on what is needed for FEGA.
Errors that has to to with things that can be user caused or remedied (files that are broken, can't be verified and so on) are sent to the error queue and then copied to FEGA so they show up in the user portal.

sda/cmd/ingest/ingest.go
jbygdell: We had an idea (before the submissions in BP took off) that in standalone/BP mode the user would get an email when something whent wrong.

zeidlitz: alright, good to know. Leaving this for now then :+1:

sda/internal/broker/v2/rabbitmq/rabbitmq.go
zeidlitz: Good to know. Will leave as is then :+1:

charts/sda-svc/templates/ingest-secrets.yaml
```diff
@@ -54,11 +54,16 @@ stringData:
     location_broker:
       cache_ttl: {{ .Values.global.archive.locationBrokerCacheTTL }}
     {{- end }}
+    sourceQueue: {{ default "ingest" .Values.ingest.sourceQueue }}
+    archivedQueue: {{ default "archived" .Values.ingest.archivedQueue }}
     broker:
     {{- if .Values.global.tls.enabled }}
       caCert: {{ template "tlsPath" . }}/ca.crt
```
KarlG-nbis: These "old" values should not be needed anymore

charts/sda-svc/values.yaml
```diff
@@ -449,6 +449,8 @@ credentials:
     mqPassword: ""
 
   ingest:
+    sourceQueue: "ingest"
```
KarlG-nbis, line 460: Ideally these should be described in the charts/sda-svc/README.md and have a `## @param` comment

sda/cmd/ingest/ingest.go
```diff
@@ -20,15 +20,17 @@ import (
 	"github.com/neicnordic/crypt4gh/keys"
 	"github.com/neicnordic/crypt4gh/model/headers"
 	"github.com/neicnordic/crypt4gh/streaming"
-	"github.com/neicnordic/sensitive-data-archive/internal/broker"
+	ingestconf "github.com/neicnordic/sensitive-data-archive/cmd/ingest/config"
+	v2 "github.com/neicnordic/sensitive-data-archive/internal/broker/v2" //nolint: revive
```
KarlG-nbis: minor: for readability would be clearer to name import by broker, eg `brokerv2` etc

sda/cmd/ingest/ingest.go
```diff
@@ -39,43 +41,52 @@ type Ingest struct {
 	ArchiveKeyList []*[32]byte
 	DB             *database.SDAdb
 	InboxReader    storage.Reader
-	MQ             *broker.AMQPBroker
+	MQ             v2.Broker
```
KarlG-nbis: minor: would make sense to rename from `MQ` -> `broker` as it could in theory be other implementations than MQ, and makes sense to make it private, as nothing outside this package should access this

sda/cmd/ingest/ingest.go
```diff
@@ -121,98 +130,65 @@ func run() error {
...
-	}
+	case sig := <-sigc:
+		log.Infof("recieved signal: %v, shutting down gracefully", sig)
+		cancel()
 
-	return nil
-}
+		return nil
+	case err := <-consumeErr:
+		if err != nil && err != context.Canceled {
```
KarlG-nbis: Use `!errors.Is(err, context.Canceled)` for error comparison

sda/cmd/ingest/ingest.go
```diff
@@ -121,98 +130,65 @@ func run() error {
...
 }
 
-func (app *Ingest) handleMessage(ctx context.Context, delivered amqp.Delivery) {
+func (app *Ingest) handleMessage(ctx context.Context, message *v2.Message) ([]func(), error) {
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
-	log.Debugf("received a message (correlation-id: %s, message: %s)", delivered.CorrelationId, delivered.Body)
 
-	err := schema.ValidateJSON(fmt.Sprintf("%s/ingestion-trigger.json", app.MQ.Conf.SchemasPath), delivered.Body)
+	log.Debugf("received message: %s", message.Key)
```
KarlG-nbis: minor: imo this debug message does not really provide any useful information

sda/cmd/ingest/ingest.go
```diff
@@ -232,444 +208,269 @@ func (app *Ingest) registerC4GHKey() error {
...
 
-		return err
+		detailsJSON, err := json.Marshal(detailsMap)
+		if err != nil {
+			log.Errorf("failed to marshal details to JSON, due to: %v", err)
+			detailsJSON = []byte("{}")
+		}
+		err = app.DB.UpdateFileEventLog(message.Key, "error", "ingest", string(detailsJSON), string(message.Body))
+		if err != nil {
+			log.Debugf("error from database when setting error event, due to: %v", err)
```
KarlG-nbis: minor: this should not be at debug level imo, possibly warn or error, as if it happends someone should investigate

sda/cmd/ingest/ingest.go
```diff
@@ -232,444 +208,269 @@ func (app *Ingest) registerC4GHKey() error {
...
-		if err := app.MQ.SendMessage(fileID, app.MQ.Conf.Exchange, "error", body); err != nil {
-			log.Errorf("failed to publish message, reason: %v", err)
-
-			return "reject"
-		}
+	if fileExistsInDataset {
+		log.Warnf("cannot cancel file: %s, as it has been added to a dataset", fileID)
 
-		return "ack"
+		return nil, nil
```
KarlG-nbis: what is the reasoning for no longer sending to the error queue in this scenario?

sda/cmd/ingest/ingest.go
```diff
@@ -232,444 +208,269 @@ func (app *Ingest) registerC4GHKey() error {
...
 	submissionLocation, err := app.DB.GetSubmissionLocation(ctx, fileID)
 	if err != nil {
-		log.Errorf("failed to get submission location for file, fileID: %s, reason: (%s)", fileID, err.Error())
-
-		return "nack"
-	}
-
-	if status != "" && submissionLocation == "" {
-		log.Errorf("file %s has been registered but has no submission location", fileID)
+		log.Errorf("failed to find submission location for file in all configured storage locations, file-id: %s", fileID)
```
KarlG-nbis: Error message does not align with the actual error here, as here error would be db connection issue or similar most likely

sda/cmd/ingest/ingest.go
```diff
@@ -232,444 +208,269 @@ func (app *Ingest) registerC4GHKey() error {
...
+	case "uploaded", "disabled":
+
 	case "":
-		// Catch all for implementations inbox uploading that does not register the file in the DB, e.g. for those not using S3inbox or sftpInbox
-		// Since we dont have the submission location in storage, we need to look through all configured storage locations.
 		var findFileErr, registerErr error
-		submissionLocation, findFileErr = app.InboxReader.FindFile(ctx, message.FilePath)
-		// Register file even if FindFile didnt succeed with submissionLocation == "", as we will add an error file event log to it in that case
-		fileID, registerErr = app.DB.RegisterFile(&fileID, submissionLocation, message.FilePath, message.User)
+		location, findFileErr := app.InboxReader.FindFile(ctx, message.Key)
```
KarlG-nbis: Here we should also set `submissionLocation` so that on the `app.InboxReader.NewFileReader` at [L301](https://github.com/neicnordic/sensitive-data-archive/pull/2381/changes#diff-45dde38c19e468fde98089670403b2ad2423071576a9193328cce6b79ed6032cR301) it will use same location where file was found

sda/cmd/ingest/ingest.go
```diff
@@ -232,444 +208,269 @@ func (app *Ingest) registerC4GHKey() error {
...
-			return "ack"
-		}
-		log.Errorf("unexpected error when opening file for reading, file-id: %s, filepath: %s, reason: %s", fileID, message.FilePath, err.Error())
+		return []func(){app.errorQueue(message), app.setErrorEvent(err.Error(), message)}, err
+	}
+	defer sourceReader.Close()
 
-		return "nack"
+	if err := app.DB.UpdateFileEventLog(fileID, "submitted", "ingest", "{}", string(message.Body)); err != nil {
+		log.Errorf("failed to set ingestion status for file from message, file-id: %s, due to: %v", fileID, err)
```
KarlG-nbis, line 348: missing return incase of err?

sda/cmd/ingest/ingest.go
```diff
@@ -232,444 +208,269 @@ func (app *Ingest) registerC4GHKey() error {
...
-		return "nack"
+	if status == "disabled" {
+		return nil
 	}
 
-	if err := app.MQ.SendMessage(fileID, app.MQ.Conf.Exchange, app.MQ.Conf.RoutingKey, archivedMsg); err != nil {
-		// TODO fix resend mechanism
-		log.Errorf("failed to publish message, reason: %v", err)
+	if err := app.DB.SetArchived(location, fileInfo, fileID); err != nil {
+		return fmt.Errorf("failed to mark file as archived, file-id: %s, due to: %v", fileID, err)
```
KarlG-nbis, line 452: Could be worth using `%w` in error wrap, so that callers could use `Errors.Is` to evaluate root error in future, etc

sda/cmd/ingest/mocks_test.go
```diff
@@ -0,0 +1,91 @@
+package main
```
KarlG-nbis, line 1: minor: imo file should include _test ending such that its clear it for testing, etc

sda/internal/broker/v2/rabbitmq/config.go
```diff
@@ -0,0 +1,228 @@
...
+			AssignFunc: func(flagName string) {
+				defaultConfig.port = viper.GetInt(flagName)
+			},
+		},
+		&config.Flag{
+			Name: "broker.user",
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.String(flagName, "", "User used to connect to the rabbitmq server")
+			},
+			Required: false,
```
KarlG-nbis, line 61: makes sense to set this as required, given empty default, and rabbitmq will require a username (and possibly similar for others)

sda/internal/broker/v2/rabbitmq/rabbitmq.go
KarlG-nbis: Yeah but then if we from an application perspective dont want to requeue something we dont return err from the `handleFunc` and in a callback func send a message to the error queue.

So main 3 scenarios I can see:
1. Everything goes well and message is processed as expected
  - `handleFunc` returns nil error such that message is acked
2. Something that we do not expect to work on a retry occurs (eg, files that are broken, can't be verified and so on)
  - `handleFunc` returns a callback func which sends message to error queue, and a nil error such that message is acked
3. Something occurs that we do expect to work if retried (eg, mainly connection issues)
  - `handleFunc` does not return a callback func which sends message to error queue, and a error such that message is nacked and requeued

sda/internal/broker/v2/broker.go
KarlG-nbis: This still seems to be the case

charts/sda-svc/values.yaml
zeidlitz: done

sda/cmd/ingest/ingest.go
zeidlitz: agreed! adjusted

sda/cmd/ingest/ingest.go
zeidlitz: done

sda/cmd/ingest/ingest.go
zeidlitz: done

sda/cmd/ingest/ingest.go
zeidlitz: done

sda/cmd/ingest/ingest.go
zeidlitz: ah looks like a oversight, will add it back

sda/cmd/ingest/ingest.go
zeidlitz: adjusted

sda/cmd/ingest/ingest.go
zeidlitz: Won't the `submissionLocation` at L301 be used from L267?

sda/cmd/ingest/ingest.go
zeidlitz: unsure, I just copied the logic from main here and there this error is just logged and then move on. Unsure what the reasoning is, but I guess we could err out here

sda/cmd/ingest/ingest.go
zeidlitz: done

sda/cmd/ingest/ingest.go
zeidlitz: agreed, adjusted

sda/cmd/ingest/mocks_test.go
zeidlitz: sure, will adjust

sda/internal/broker/v2/rabbitmq/config.go
zeidlitz: sure, will adjust

charts/sda-svc/templates/ingest-secrets.yaml
zeidlitz: aren't they used by the v1 broker package that other services might still use?

sda/internal/broker/v2/rabbitmq/rabbitmq.go
zeidlitz: Right, makes sense to me. What do you say @jbygdell , will what @KarlG-nbis described above work for the fega needs?

sda/internal/broker/v2/broker.go
zeidlitz: So I adjusted the doc in this case not the logic. So now it goes callbacks -> ack/nack. Is it important that we do ack/nack -> callbacks ? In that case I can restructure it as so. My thinking was that the important part was that the callbacks were executed regardless if the message is ack/nacked, but I could be misstaken.

sda/cmd/ingest/ingest.go
KarlG-nbis: In the majority of scenarios yes, but this `case ""` means we've recieved a ingest message for a file which has not been registered in the database(ie, not uploaded through the s3inbox, or equivalent), so the `submissionLocation` from L267 will be "", which is why here we find where the file is, and register it, but then we need to read it in the location where it was found

charts/sda-svc/templates/ingest-secrets.yaml
KarlG-nbis: Yeah they are used by the other services, but this secret is constructing the config file for the ingest service specifically

sda/internal/broker/v2/broker.go
KarlG-nbis: Well the intended idea was to execute it after nack/ack, as otherwise the callbacks dont really serve much purpose, and the handler could just do the actions before returning. 

The idea with doing it after nack/ack is if a handler has an action that benefits from done afterwards, for example, commiting a transaction, etc

But Im not sure how much value we get from this callback functionality, possibly worth considering if it just adds unnecessary complexity without much benefit

sda/cmd/ingest/ingest.go
KarlG-nbis: aah ok, but yeah imo makes sense to return error here

charts/sda-svc/templates/ingest-secrets.yaml
zeidlitz: Ah right, of course, will get rid of the old ones.

sda/cmd/ingest/ingest.go
zeidlitz: alright, I think I see it. So simply just ensure we use `submissionLocation, findFileErr := app.InboxReader.FindFile(ctx, message.Key)` instead? Or am I just lost?

sda/cmd/ingest/ingest.go
zeidlitz: agreed, can adjust and see if it impacts anything in the integration tests

sda/internal/broker/v2/broker.go
zeidlitz: Maybe in this iteration they don't add that much obvious value. But I think there is value in having the structure for them in place so we have the ability. Will adjust so we run the ack/nack and after execute the callbacks and reflect it in the doc.

sda/cmd/ingest/ingest.go
KarlG-nbis: Yeah precisely, but not with a `:=` but `=` as otherwise the assignment of `submissionLocation` from `app.InboxReader.FindFile` will just be within that switch case

zeidlitz: ah ofc, got it!

sda/cmd/ingest/ingest.go
```diff
@@ -232,444 +206,271 @@ func (app *Ingest) registerC4GHKey() error {
...
-		if err := app.MQ.SendMessage(fileID, app.MQ.Conf.Exchange, "error", body); err != nil {
-			log.Errorf("failed to publish message, reason: %v", err)
+	if fileExistsInDataset {
+		log.Warnf("cannot cancel file: %s, as it has been added to a dataset", fileID)
 
-			return "reject"
-		}
-
-		return "ack"
+		return []func(){app.errorQueue(message), app.setErrorEvent(err.Error(), message)}, nil
```
jhagberg: I noticed that the new `if fileExistsInDataset` block in `cancelFile` (`sda/cmd/ingest/ingest.go:215-219`) calls `app.setErrorEvent(err.Error(), message)`, but `err` is `nil` at this point. `IsFileInDataset` returned without error on the line above (the `err != nil` check at line 211 already handled the non-nil case), so the `(true, nil)` path reaches line 218 with `err == nil`.

Would not this mean `err.Error()` panics the consumer goroutine the first time a cancel message arrives for a file that has already been added to a dataset?

The intent looks like a fixed-string detail, something along the lines of `app.setErrorEvent("cannot cancel file: already added to a dataset", message)`. Would that match what you had in mind when adding the error queue back?

zeidlitz: Ah very good catch! Thanks. Yes this is a sloppy copy paste on my side. Will adjust

charts/sda-svc/templates/ingest-secrets.yaml
zeidlitz: Hm.. I think due to ingest still using the v1 config package to conf the database I need to keep clientCert and clientKey for now to make the chart tests pass since they are flag as needed in the v1 config package.

KarlG-nbis: Yeah this is since the "ingest" is still calling `c.configBroker()`, so if we update

```go
case "finalize", "ingest", "mapper", "verify":
	err := c.configBroker()
	if err != nil {
		return nil, err
	}
	err = c.configDatabase()
	if err != nil {
		return nil, err
	}
	c.configSchemas()
```
=>

``` go
case "ingest":
  	err = c.configDatabase()
	if err != nil {
		return nil, err
	}
case "finalize", "mapper", "verify":
	err := c.configBroker()
	if err != nil {
		return nil, err
	}
	err = c.configDatabase()
	if err != nil {
		return nil, err
	}
	c.configSchemas()
``` 
at https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/internal/config/config.go#L494
it should be all good, as ingest no longer need to setup the "v1" broker conf

and remove the "broker.*" configs from the required configs for ingest at 
https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/internal/config/config.go#L268

sda/cmd/ingest/ingest.go
jhagberg: Coming back to this before approval: would it be worth opening a follow-up issue to track the error-queue contract before merging? `cmd/notify/notify.go:91-94` still parses messages from this queue as `broker.InfoError` with a base64-encoded `OriginalMessage`, while the new callbacks in this PR publish the raw ingestion trigger. A tracking issue would keep this PR focused on broker v2, and leave the payload format and routing semantics to be decided in one place.

charts/sda-svc/values.yaml
```diff
@@ -449,6 +449,12 @@ credentials:
...
+    ## @param ingest.sourceQueue
+    ## Name of the queue the ingest service consumes ingest triggers from
+    archivedQueue: "archived"
+    ## @param ingest.archivedQueue
+    ## Name of the queue the ingest produces archived messages to
```
jbygdell, line 465: Nit pick: Messages are not sent to a queue but to an `exchange` tagged with a specific `routing key`,  based on the configuration of the MQ server delivers the message(s) to a specific set of queue(s). 

```suggestion
    routingKey: "archived"
    ## @param ingest.routingKey
    ## Name of the routingKey the ingest delivers archived messages with
```

sda/cmd/auth/main.go
```diff
@@ -222,7 +222,7 @@ func (auth AuthHandler) getEGAConf(ctx iris.Context) {
 // getOIDC redirects to the oidc page defined in auth.Config
 func (auth AuthHandler) getOIDC(ctx iris.Context) {
 	state := uuid.New()
-	ctx.SetCookie(&http.Cookie{Name: "state", Value: state.String(), Secure: true})
+	ctx.SetCookie(&http.Cookie{Name: "state", Value: state.String(), Secure: true, HttpOnly: true, SameSite: http.SameSiteLaxMode})
```
jbygdell, line 228: hard coding these values will prevent the auth app to work properly in a non TLS setup

sda/cmd/ingest/config/config.go
```diff
@@ -0,0 +1,69 @@
...
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.String(flagName, "ingest", "The queue where the ingest service consumes ingest messaages from")
+			},
+			Required: false,
+			AssignFunc: func(flagName string) {
+				sourceQueue = viper.GetString(flagName)
+			},
+		},
+		&config.Flag{
+			Name: "archivedQueue",
```
jbygdell, line 30:
```suggestion
			Name: "routingKey",
```

sda/internal/broker/v2/rabbitmq/config.go
```diff
@@ -0,0 +1,228 @@
...
+		&config.Flag{
+			Name: "broker.password",
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.String(flagName, "", "Password used to connect to the rabbitmq server")
+			},
+			Required: true,
+			AssignFunc: func(flagName string) {
+				defaultConfig.password = viper.GetString(flagName)
+			},
+		},
```
jbygdell, line 75: Credentials should be optional since RabbitMQ supports mTLS.

sda/internal/broker/v2/rabbitmq/config.go
```diff
@@ -0,0 +1,228 @@
...
+		&config.Flag{
+			Name: "broker.vhost",
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.String(flagName, "/sda", "Virtual host used to connect the rabbitmq server")
```
jbygdell, line 79:
```suggestion
				flagSet.String(flagName, "/", "Virtual host used to connect the rabbitmq server")
```

Better not hard code this since it is dependent on the MQ configuration.

sda/internal/broker/v2/rabbitmq/config.go
```diff
@@ -0,0 +1,228 @@
...
+			},
+			Required: false,
+			AssignFunc: func(flagName string) {
+				defaultConfig.clientKey = viper.GetString(flagName)
+			},
+		},
+		&config.Flag{
+			Name: "broker.prefetch_count",
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.Int(flagName, 2, "How many messages the rabbitmq server will deliver to a client before receiving delivery acknowledgements")
```
jbygdell:
```suggestion
				flagSet.Int(flagName, 1, "How many messages the rabbitmq server will deliver to a client before receiving delivery acknowledgements")
```

Prefetch of `1` is required for horizontal scaling.

sda/internal/broker/v2/rabbitmq/config.go
```diff
@@ -0,0 +1,228 @@
...
+
+// buildMQURI builds the MQ connection URI
+func (cfg *options) buildMQURI() string {
+	protocol := "amqp"
+	if cfg.ssl {
+		protocol = "amqps"
+	}
+
+	return fmt.Sprintf("%s://%s:%s@%s:%d%s", protocol, cfg.user, cfg.password, cfg.host, cfg.port, cfg.vhost)
+}
```
jbygdell, line 198: This needs to handle all combinations of the the optional settings.
* user/pass
* port
* vhost

sda/internal/broker/v2/rabbitmq/config.go
```diff
@@ -0,0 +1,228 @@
...
+func (cfg *options) setupTLSConfig() (*tls.Config, error) {
+	systemCAs, err := x509.SystemCertPool()
+	if err != nil {
+		log.Errorf("failed to read system CAs: %v", err)
+
+		return nil, err
+	}
+
+	tlsConfig := tls.Config{
+		MinVersion: tls.VersionTLS12,
```
jbygdell, line 209:
```suggestion
		MinVersion: tls.VersionTLS13,
```

Might be time to bump this to TLS1.3

sda/internal/broker/v2/rabbitmq/rabbitmq.go
```diff
@@ -0,0 +1,324 @@
...
+			log.Errorf("failed to start consuming: %v", err)
+			time.Sleep(time.Duration(b.config.timeout) * time.Second)
+
+			continue
+		}
+
+		if done := b.consumeMessages(ctx, messageChan, handleFunc); done {
+			return ctx.Err()
+		}
+	}
```
jbygdell, line 64: What's up with the for loops in this package?

If you want things to run indefinitely there are better solutions.

sda/internal/broker/v2/rabbitmq/rabbitmq.go
jbygdell: That's the reason the current code uses the third (reject) option built in to RabbitMQ.
we need to be able to separate the cases based on type and expectancy.
* All ok (Ack)
* Soft Error (Nack) - Comms timeout or temporary upstream errors
* Hard Error type 1 (Ack) - send to error queue (file not found, not decryptable, checksums that differ from what the users supplies ...)
* Hard Error type 2 (Reject) - Unexpected error that is not user related

zeidlitz: The way I see it, everything you mention is covered by what Karl wrote above? Or am I missing something?

sda/cmd/auth/main.go
zeidlitz: Are you sure? Which one enforces TLS? It would be nice to get rid of the linting error this line gives in the github action

sda/internal/broker/v2/rabbitmq/rabbitmq.go
zeidlitz: Please be more constructive. This is very poor feedback.

sda/internal/broker/v2/rabbitmq/config.go
zeidlitz: It's not hard coded, It's just a default value.

sda/internal/broker/v2/rabbitmq/config.go
zeidlitz: @KarlG-nbis any thoughts, I believe you set this in the first draft? I have no opinion regarding this

KarlG-nbis: I set it as 2 as that was the default in "broker v1", https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/internal/config/config.go#L694

sda/internal/broker/v2/rabbitmq/rabbitmq.go
jbygdell: It was, I'm just explaining how the current code uses the different responses built into rabbitMQ.

sda/internal/broker/v2/rabbitmq/config.go
jbygdell: Given that we can't process messages in parallel it makes no sense to fetch multiple messages.

zeidlitz: Will adjust

sda/internal/broker/v2/rabbitmq/rabbitmq.go
zeidlitz: Having the re-queue logic is the way forward imo. With that said I will opt for leaving the implementation as is for now since I would have to rethink a lot about the ack/nack requeue flow in the ingest service and think it would be good to not introduce to much changing behavior for now. This could be a follow up change.

charts/sda-svc/templates/ingest-secrets.yaml
```diff
@@ -54,8 +54,13 @@ stringData:
...
       cache_ttl: {{ .Values.global.archive.locationBrokerCacheTTL }}
     {{- end }}
+    sourceQueue: {{ default "ingest" .Values.ingest.sourceQueue }}
+    archivedQueue: {{ default "archived" .Values.ingest.archivedQueue }}
     broker:
     {{- if .Values.global.tls.enabled }}
+      ca_cert: {{ template "tlsPath" . }}/ca.crt
+      client_cert: {{ template "tlsPath" . }}/tls.crt
+      client_key: {{ template "tlsPath" . }}/tls.key
       caCert: {{ template "tlsPath" . }}/ca.crt
```
KarlG-nbis: Do we still need these configs that where needed for v1 but has been renamed for v2?

sda/cmd/ingest/config/config.go
```diff
@@ -0,0 +1,69 @@
...
+			Name: "schemaType",
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.String(flagName, "isolated", "Path to JSON schemas to validate rabbitmq messages against")
+			},
+			Required: false,
+			AssignFunc: func(flagName string) {
+				switch viper.GetString("schema.type") {
+				case "federated":
+					schemaPath = "/schemas/federated/"
+				default:
```
KarlG-nbis, line 52: Here we could have the default be some panic, fatal, etc so that invalid `schemaType` config isnt accepted, and mention the supported types in the description, or have the `case "isolated"`, and have the default case log a warning, and then default to "isolated"

sda/cmd/ingest/ingest.go
```diff
@@ -41,46 +43,47 @@ type Ingest struct {
 	ArchiveKeyList []*[32]byte
 	db             database.Database
 	InboxReader    storage.Reader
-	MQ             *broker.AMQPBroker
+	Broker         brokerv2.Broker
+}
+
+type DecryptResult struct {
```
KarlG-nbis: struct should not be exported as it isnt used outside the `ingest` package (and shouldnt)

sda/internal/broker/v2/rabbitmq/config.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+func (cfg *options) buildMQURI() string {
+	protocol := "amqp"
+	if cfg.ssl {
+		protocol = "amqps"
+	}
+
+	if cfg.user != "" && cfg.password != "" {
+		return fmt.Sprintf("%s://%s:%s@%s:%d%s", protocol, cfg.user, cfg.password, cfg.host, cfg.port, cfg.vhost)
```
jbygdell:
```suggestion
		return fmt.Sprintf("%s://%s:%s@%s:%d%s", protocol, url.PathEscape(cfg.user), url.PathEscape(cfg.password), cfg.host, cfg.port, cfg.vhost)
```

Credentials need to rendered web safe or a random string password will cause the app to fail startup in the credentials contains characters like ( + / & ?)

sda/internal/broker/v2/rabbitmq/rabbitmq.go
```diff
@@ -0,0 +1,324 @@
...
+
+	if b.consumeChannel == nil || b.consumeChannel.IsClosed() {
+		return false
+	}
+
+	return true
+}
+
+func (b *rmqBroker) ensureConnected(ctx context.Context) error {
```
KarlG-nbis, line 158: hmm not sure if we need this `ensureConnected` to retry forever (with 30sec delay between each attempt), would probably be enough with just one attempt and `Publish` would abort if error, and `Subscribe` would continue to attempt another loop, etc

sda/internal/broker/v2/rabbitmq/rabbitmq.go
```diff
@@ -0,0 +1,324 @@
...
+	msg := &broker.Message{
+		Key:     delivery.CorrelationId,
+		Headers: delivery.Headers,
+		Body:    delivery.Body,
+	}
+
+	callbacks, err := handleFunc(ctx, msg)
+	if err != nil {
+		log.Errorf("error handling message %s: %v", msg.Key, err)
+		delivery.Nack(false, false)
```
KarlG-nbis: As discussed in [comment](https://github.com/neicnordic/sensitive-data-archive/pull/2381/changes#r3092301382) updating ingest to properly handle  ack/nack/nack requeue will be in a future issue.

But I think it makes sense to update the broker/v2 library to support it as part of this, as that shouldnt require much effort, should just be something like:

```go
callbacks, err := handleFunc(ctx, msg)
if err != nil {
	log.Errorf("error handling message %s: %v", msg.Key, err)
	delivery.Nack(false, false)
} else {
	delivery.Ack(false)
}
```

=>

```go
callbacks, err := handleFunc(ctx, msg)
if err != nil {
	delivery.Nack(false, true)
} 

delivery.Ack(false)
```

sda/internal/broker/v2/rabbitmq/rabbitmq.go
```diff
@@ -0,0 +1,324 @@
...
+func (b *rmqBroker) handleDelivery(ctx context.Context, delivery amqp.Delivery, handleFunc func(context.Context, *broker.Message) ([]func(), error)) {
+	msg := &broker.Message{
+		Key:     delivery.CorrelationId,
+		Headers: delivery.Headers,
+		Body:    delivery.Body,
+	}
+
+	callbacks, err := handleFunc(ctx, msg)
+	if err != nil {
+		log.Errorf("error handling message %s: %v", msg.Key, err)
```
KarlG-nbis: Imo this log should be removed as lib packages should only log errors, etc if error originates from the lib, otherwise we should rely on the application to log the relevant information

sda/cmd/ingest/ingest.go
```diff
@@ -239,478 +205,292 @@ func (app *Ingest) registerC4GHKey(ctx context.Context) error {
...
-	}
-	if err := tx.CancelFile(ctx, fileID, string(m)); err != nil {
-		log.Errorf("failed to cancel file with id: %s, due to %v", fileID, err)
-
-		if err := tx.Rollback(); err != nil {
-			log.Errorf("failed to rollback CancelFile transaction, reason: %v", err)
-		}
-
-		return "nack"
+	if err := app.db.CancelFile(ctx, fileID, string(message.Body)); err != nil {
```
KarlG-nbis: the transaction around `CancelFile` has now been removed? Was it intentional?

ie the:
```go
// Ideally this transaction should span the whole message processing, but for now just spans the CancelFile
tx, err := app.db.BeginTransaction(ctx)
if err != nil {
	log.Errorf("failed to begin transaction, reason: %v", err)

	return "nack"
}
if err := tx.CancelFile(ctx, fileID, string(m)); err != nil {
	log.Errorf("failed to cancel file with id: %s, due to %v", fileID, err)

	if err := tx.Rollback(); err != nil {
		log.Errorf("failed to rollback CancelFile transaction, reason: %v", err)
	}
    return nil, err
}

if err := tx.Commit(); err != nil {
	log.Errorf("failed to commit CancelFile transaction, reason: %v", err)
	_ = tx.Rollback()

	return "nack"
}
```

sda/cmd/ingest/ingest.go
```diff
@@ -239,478 +205,292 @@ func (app *Ingest) registerC4GHKey(ctx context.Context) error {
...
-				return
-			}
-		}
+	return nil, nil
+}
 
-		_ = contentWriter.Close()
-		_ = file.Close()
-	}()
+func (app *Ingest) decrypt(source io.ReadCloser) (DecryptResult, error) {
```
KarlG-nbis: Any reason for not returning a `*DecryptResult` as otherwise we are returning a copy of the struct instead of just a pointer to it, leading to higher memory usage (thou probably in an insignificant amount)

sda/cmd/ingest/ingest.go
zeidlitz: You are correct, the transaction should be included. Must have picked the wrong during rebase.

sda/cmd/ingest/ingest.go
zeidlitz: So personally I think of it like this: 

Is it a short lived small object that just needs to hold some data -> return reference. 
Is it a long lived larger object that will share data and pass it around to other packages -> return pointer

Here I think returning the reference signals the intent.

charts/sda-svc/templates/ingest-secrets.yaml
zeidlitz: I think I removed them and it started failing some of the chart tests, but have this changed since rebase? I can remove them and see if it is still needed

sda/cmd/ingest/config/config.go
zeidlitz: sure!

sda/cmd/ingest/ingest.go
zeidlitz: agreed!

sda/internal/broker/v2/rabbitmq/config.go
zeidlitz: will add!

sda/internal/broker/v2/rabbitmq/rabbitmq.go
zeidlitz: agreed! Will adjust

sda/internal/broker/v2/rabbitmq/rabbitmq.go
zeidlitz: ah good to know! Makes sense

sda/internal/broker/v2/rabbitmq/rabbitmq.go
zeidlitz: I did try this change, but It breaks the integration tests when I add it and I don't want to add to much behavior change in the ingest service right now to make it work.

charts/sda-svc/templates/ingest-secrets.yaml
KarlG-nbis: Yeah this should have been fixed when you made the config (v1) package not call `c.configBroker()` for the ingest service

sda/internal/broker/v2/rabbitmq/rabbitmq.go
KarlG-nbis: hmm personally I think this is a quite an important behaviour and that this shouldnt be released until its supported. But im willing to be overruled if team thinks otherwise.

 And the changes in ingest, should only be to return nil error incases where it shouldnt retry(ie in all places, currently) afaik

sda/cmd/ingest/ingest.go
```diff
@@ -301,416 +249,268 @@ func (app *Ingest) cancelFile(ctx context.Context, fileID string, message schema
...
 	}
+}
 
-	return nil
+func (app *Ingest) errorQueue(message *brokerv2.Message) func() {
+	return func() {
+		if err := app.Broker.Publish(context.Background(), "error", *message); err != nil {
+			log.Errorf("failed to publish to error queue: %v", err)
+		}
+		log.Info("published message to error queue", "message", message.Key)
```
KarlG-nbis: malformed log message, (guessing you were thinking to use slog), currently becomes:

```published message to error queuemessageb8c48126-d747-4252-b211-564579eb5f42```

File: sda/internal/broker/v2/rabbitmq/rabbitmq.go (hunk 44 above)
Comment (KarlG-nbis) [reply sda/internal/broker/v2/rabbitmq/rabbitmq.go]: As also doesnt align with the comment `if handleFunc returns an error the message should be nacked and marked for reconsumption` on

```
type Broker interface {
	// Subscribe starts subscribing to the specified sourceQueue through the consumerGroup and handling each incoming message with the handleFunc
    // if handleFunc returns an error the message should be nacked and marked for reconsumption
	// handleFunc can return a slice of callbacks that trigger after acknowledgment of the message's regardless of acknowledgment status.
	Subscribe(ctx context.Context, sourceQueue string, handleFunc func(ctx context.Context, msg *Message) ([]func(), error)) error
  ...
```

File: sda/cmd/ingest/config/config.go [hunk 49]
```diff
@@ -0,0 +1,74 @@
...
+				flagSet.String(flagName, "isolated", "Path to JSON schemas to validate rabbitmq messages against")
+			},
+			Required: false,
+			AssignFunc: func(flagName string) {
+				schemaType := viper.GetString("schemaType")
+				switch schemaType {
+				case "federated":
+					schemaPath = "/schemas/federated/"
+				case "isolated":
+					schemaPath = "/scehmas/isolated/"
```
KarlG-nbis: Typo in `scehmas`

sda/cmd/ingest/ingest.go
```diff
@@ -128,98 +129,63 @@ func run() error {
...
-func (app *Ingest) handleMessage(ctx context.Context, delivered amqp.Delivery) {
+func (app *Ingest) handleMessage(ctx context.Context, message *brokerv2.Message) ([]func(), error) {
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
-	log.Debugf("received a message (correlation-id: %s, message: %s)", delivered.CorrelationId, delivered.Body)
 
-	err := schema.ValidateJSON(fmt.Sprintf("%s/ingestion-trigger.json", app.MQ.Conf.SchemasPath), delivered.Body)
+	err := schema.ValidateJSON(fmt.Sprintf("%s/ingestion-trigger.json", ingestconf.SchemaPath()), message.Body)
 	if err != nil {
-		log.Errorf("validation of incoming message (ingestion-trigger) failed, correlation-id: %s, reason: (%s)", delivered.CorrelationId, err.Error())
```
KarlG-nbis, line 172: Could keep this log message or something similar such that we have an error log incase `err := schema.ValidateJSON` failes

sda/internal/broker/v2/rabbitmq/rabbitmq.go
zeidlitz: Sure, can give it a try!

sda/cmd/ingest/ingest.go
zeidlitz: adjusted!

sda/cmd/ingest/config/config.go
zeidlitz: fixed!

sda/cmd/ingest/ingest.go
zeidlitz: added!

charts/sda-svc/templates/ingest-secrets.yaml
zeidlitz: ah ofc, will fix

charts/sda-svc/templates/ingest-secrets.yaml
zeidlitz: done!

KarlG-nbis (approved):
Nice work!

sda/cmd/ingest/ingest.go
```diff
@@ -128,98 +129,65 @@ func run() error {
...
+		return []func(){app.errorQueue(message), app.setErrorEvent(err.Error(), message)}, nil
+	}
+
+	var ingestionTrigger schema.IngestionTrigger
+	err = json.Unmarshal(message.Body, &ingestionTrigger)
+	if err != nil {
+		log.Errorf("could not unmarshall message, due to: %v", err)
 
-		return
+		return []func(){app.errorQueue(message), app.setErrorEvent(err.Error(), message)}, err
```
jhagberg: I noticed that `ValidateJSON` at line 166 now correctly returns `nil` as the error (so the message gets Ack'd and the callbacks route it to the error queue), but `json.Unmarshal` right below at line 174 still returns the raw `err`. Since `handleDelivery` in `rabbitmq.go` now calls `Nack(false, true)` when the handler returns an error, would not a malformed message that passes schema validation but fails unmarshalling loop indefinitely between the broker and the consumer?

The `default:` status case in `ingestFile` at line 336 has the same pattern, and that one is more likely to be hit in practice (a file that has already been archived receives a duplicate trigger). The `"unknonw ingest type"` at line 187 is similar (also a typo in the error string).

Should these follow the same `(callbacks, nil)` pattern as `ValidateJSON`?

sda/internal/broker/v2/rabbitmq/config.go
```diff
@@ -0,0 +1,232 @@
...
+	if cfg.caCert != "" {
+		cacert, err := os.ReadFile(cfg.caCert)
+		if err != nil {
+			return nil, err
+		}
+		if ok := tlsConfig.RootCAs.AppendCertsFromPEM(cacert); !ok {
+			log.Warnln("No certs appended, using system certs only")
+		}
+	}
+
```
jhagberg, line 223: Nit: the guard here checks `defaultConfig.clientKey` instead of the receiver `cfg.clientKey`. Today they always hold the same value since no option functions are passed to `NewRabbitMQBroker`, but if option functions are added later this would silently skip loading the client certificate.

jhagberg (approved):
Looks good overall. The broker v2 landing is solid and the transaction wrapping around CancelFile/RegisterFile is a welcome addition. Left one inline comment about the nack-requeue interaction with a few error return paths, and a nit on the TLS config guard. Nothing blocking from my side.

sda/cmd/ingest/ingest.go
zeidlitz: yes, correct, no reason to re-queue a json that can't be unmarshalled. This looks like a copy-paste miss from me. Will adjust.

sda/internal/broker/v2/rabbitmq/config.go
zeidlitz: Ah, are you saying that on [L215](https://github.com/neicnordic/sensitive-data-archive/pull/2381/changes#diff-23590bea3a95da2a84e755f9557703b6ea5b2b05aea512606038d6afa9c4d31fR215) It should read

```go
	if cfg.clientCert != "" && cfg.clientKey != "" {
```

instead?

sda/internal/broker/v2/rabbitmq/config.go
```diff
@@ -0,0 +1,232 @@
...
+	if cfg.ssl {
+		protocol = "amqps"
+	}
+
+	if cfg.user != "" && cfg.password != "" {
+		return fmt.Sprintf("%s://%s:%s@%s:%d%s", protocol, url.PathEscape(cfg.user), url.PathEscape(cfg.password), cfg.host, cfg.port, cfg.vhost)
+	}
+
+	return fmt.Sprintf("%s://%s:%d%s", protocol, cfg.host, cfg.port, cfg.vhost)
+}
```
jbygdell, line 198:
```suggestion
func (cfg *options) buildMQURI() string {
	protocol := "amqp"
	if cfg.ssl {
		protocol = "amqps"
	}

	connURL = fmt.Sprintf("%s://@%s", protocol, cfg.host)

	if cfg.user != "" && cfg.password != "" {
		connURL = fmt.Sprintf("%s://%s:%s@%s", protocol, url.PathEscape(cfg.user), url.PathEscape(cfg.password), cfg.host)
	}	
	if cfg.port != 0 {
		connURL = fmt.Sprintf(%s:%d, connURL, cfg.Port)
	}
	if cfg.vhost != "" {
		connURL = path.Join(connURL, cfg.vhost)
	}
	
	return connURL
}
```

Port and vhost are not mandatory

sda/internal/broker/v2/rabbitmq/config.go
jbygdell: Given that vhost is not required when configuring the client the default should actually be `""`

sda/internal/broker/v2/rabbitmq/config.go
zeidlitz: This is starting to look a bit to messy for my taste. I will just redo it with `url.Url` and I think I can get it a bit cleaner.

sda/internal/broker/v2/rabbitmq/config.go
```diff
@@ -0,0 +1,241 @@
...
+		},
+		&config.Flag{
+			Name: "broker.vhost",
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.String(flagName, "/sda", "Virtual host used to connect the rabbitmq server")
+			},
+			Required: false,
+			AssignFunc: func(flagName string) {
+				defaultConfig.vhost = viper.GetString(flagName)
+			},
```
jbygdell, line 84:
```suggestion
			AssignFunc: func(flagName string) {
				vhost := viper.GetString(flagName)
				if len(vhost) > 1 {
					strings.trimLeft(, "/")
				}
				defaultConfig.vhost = url.PathEscape(vhost)
			},
```

The vhost component of the URI does not include the leading "/" character from the path.
The vhost is a single segment. Therefore, any "/" characters that appear in the vhost name must be percent-encoded. URIs with multi-segment paths do not obey this specification.

If the default vhost `/` is to be used it needs to be URL escaped to `%2F`

zeidlitz: so should it be enough to do ```url.PathEscape(cfg.vhost)``` ?

jbygdell: If the vhost string is `/example` it becomes `%2Fexample` which in the connection URI becomes `amqp://host/%2Fexample` and that will cause a failure if the vhost in the MQ is `example`

sda/internal/broker/v2/rabbitmq/config.go
jhagberg: Yes exactly, that's what I meant. Both the guard and the file read should reference `cfg` consistently.

zeidlitz: Makes sense, thanks for clarifying

sda/internal/broker/v2/rabbitmq/config.go
zeidlitz: Are you sure? 

When I run it It, If I set `vhost=/example` the connection string becomes `/example`

jbygdell: The easiest way to ensure this doesn't becomes an issue later on is to write a test.

|vhost variable| connUR|
|--|--|
|nil|"amqp://example.org"|
|""|"amqp://example.org/"|
| "/" | "amqp://example.org/%2F" |
|"foo"|"amqp://example.org/foo"|
|"foo/bar"|"amqp://example.org/foo%2Fbar"|

The tricky case is if the vhost variable starts with a `/` do we trim it or escape it.
Vhosts are configured by name and not path so a config variable that starts with a `/` can either mean sloppy config or that the vhost actually is named that (although not best practice).

zeidlitz: This is pretty much a non issue imo. I don't see the point in supporting non standard configurations, if someone tries to configure a vhost to be named literally `/vhost` It should not be that strange that the connection string will be malformed.

PR #2319, opened 2026-03-16
docs(adr): propose ADR-0003 replace s3inbox in-memory file ID cache

jhagberg:
## Related issue(s) and PR(s)
This PR addresses #2295 and relates to #1358.

## Description

Proposes ADR-0003: replace the s3inbox in-memory `map[uniqueFileID]string` with
database lookups via `GetFileIDByUserPathAndStatus`.

The in-memory map caused a fatal concurrent map writes crash in production
(#2295). A hotfix (822354f) added `sync.RWMutex`, but the underlying design
has further problems: per-pod state loss on reconnection (#1358), TOCTOU races,
memory leaks on incomplete uploads, and a bug where the existing DB fallback
sends a broker message with an empty file ID before looking up the actual ID.

The ADR evaluates 4 options:
1. **Remove cache, use PostgreSQL** (recommended)
2. sync.RWMutex with read-once pattern
3. Ristretto (in-process cache with TTL)
4. Redis (deferred — may be evaluated for other needs)

Includes Mermaid sequence diagrams showing the current bug (empty fileID sent
to broker on cross-pod reconnect) vs the proposed fix.

## ADR

- [x] This PR includes an architecturally significant decision (see `docs/decisions/README.md`)
  - [x] A decision record has been added or updated in `docs/decisions/`
  - [x] The decision record is listed in the `docs/decisions/README.md` index table

## How to test

This is a documentation-only PR (ADR proposal). Review the decision record at
`docs/decisions/0003-shared-state-strategy-for-s3inbox-and-caching.md` for
clarity, completeness, and technical accuracy. The Mermaid diagrams render on
GitHub.

Issue #2295, opened 2026-03-05
[s3inbox] port v3.1.0 hotfix for concurrent map read / write to main


KarlG-nbis:
We had issue in PRD with concurrent map read / writes in the s3inbox with concurrent uploads.

```
fatal error: concurrent map writes

goroutine 13331 [running]:
internal/runtime/maps.fatal({0xb71015?, 0xab27a0?})
        /usr/local/go/src/runtime/panic.go:1058 +0x18
internal/runtime/maps.(*Map).Delete(0xc0004dd860, 0xab27a0, 0xc0005fb958)
        /usr/local/go/src/internal/runtime/maps/map.go:672 +0x10f
main.(*Proxy).allowedResponse(0xc0000a8180, {0xc92558, 0xc0034001c0}, 0xc0025f8000, {0xc99558, 0xc003e4a850})
        /go/cmd/s3inbox/proxy.go:254 +0x10f6
main.(*Proxy).ServeHTTP(0xc0000a8180, {0xc92558, 0xc0034001c0}, 0xc0025f8000)
        /go/cmd/s3inbox/proxy.go:109 +0x17b
github.com/gorilla/mux.(*Router).ServeHTTP(0xc0000a80c0, {0xc92558, 0xc0034001c0}, 0xc00330bb80)
        /root/go/pkg/mod/github.com/gorilla/mux@v1.8.1/mux.go:212 +0x1e2
net/http.serverHandler.ServeHTTP({0xc8e690?}, {0xc92558?, 0xc0034001c0?}, 0x6?)
        /usr/local/go/src/net/http/server.go:3301 +0x8e
net/http.(*conn).serve(0xc004859c20, {0xc931c8, 0xc0004ddf50})
        /usr/local/go/src/net/http/server.go:2102 +0x625
created by net/http.(*Server).Serve in goroutine 1
        /usr/local/go/src/net/http/server.go:3454 +0x485
```

Hotfix was prepared based on v3.1.0 so we could fix it in PRD: https://github.com/neicnordic/sensitive-data-archive/commit/822354f668c08f3a3531cfa9dfc0ad8d3292805f but we also need to port this to main which is the aim of this task.

In addition the sync.RWMutex there are additional improvements to be made, eg:
* resuse fileID read from cache instead of reading each time
  * better performance since each call to .read() will trigger a RLock() and RUnlock().
[...]

PR #2382, opened 2026-04-10
feat(s3inbox): use postgres instead of in memory cache for fileIds

### LINKED ISSUE #2375: Implement ADR-0003: Replace s3inbox in-memory file ID cache with postgreSQL
# Please describe the feature

Remove the in-memory cache entirely and retrieve the file ID from PostgreSQL when needed

ADR described in detail in #2319.

# Acceptance criteria

- [x] Double check that the code-path has tests, if not implement some.
- [x] Demo in sprint review how the database is used.
- [x] Demo in sprint reivew that a reuploaded file does not generate a new `registered` event in the file event log.

# Additional context

This is inestad of implementing #2295.

# Estimation of size

small/medium/big

# Estimation of priority

low/medium/high

KarlG-nbis:
# Related issue(s) and PR(s)
This PR closes #2375.

# Description
- Add new DB func GetFileIDInInbox which returns the file id of a file if the latest file event deems it should exist in the inbox

- Also update dedection and mapping of s3 actions
  - Only explicitly allow GetBucketLocation, ListObjects, ListObjectsV2, PutObject, UploadPart, CreateMultiPartUpload, CompleteMultiPartUpload, AbortMultiPartUpload, ListMultiPartUpload, ListParts

- Remove unecessary logging which do not provide any useful information

- Update handling of s3 action to hopefully be more readable

- Reduce information exposed on errors, ie do not return errors, etc instead log them and return generic error message

- Update AlwaysAllow authenticator to set "dummy" as the subject on the token

[ADR-0003](https://github.com/neicnordic/sensitive-data-archive/pull/2319) states to specifically use GetFileIDByUserPathAndStatus, but instead new GetFileIDInInbox with different filter semantics was used to ensure that we find files based on all file event states that indicate that a file should be in the inbox.

# How to test
- Unit tests pass
- Integration tests pass

### DIFF
--- a/.github/integration/tests/sda/01_install_dependencies.sh
+++ b/.github/integration/tests/sda/01_install_dependencies.sh
@@ -2,7 +2,7 @@
 set -e
 
 # install tools if missing
-for t in curl expect jq openssh-client postgresql-client xxd; do
+for t in curl expect jq openssh-client postgresql-client xxd awscli; do
     if [ ! "$(command -v $t)" ]; then
         if [ "$(id -u)" != 0 ]; then
             echo "$t is missing, unable to install it"
--- a/.github/integration/tests/sda/15_s3cmd-errorcodes_test.sh
+++ b/.github/integration/tests/sda/15_s3cmd-errorcodes_test.sh
@@ -14,8 +14,8 @@ if ! [[ "$list_all_buckets" =~ "Forbidden" ]]; then
 fi
 
 # test disallowed characters
-disallowed_characters=$(s3cmd -c s3cfg -q put s3cfg s3://test_dummy.org/fai\| 2>&1)
-if ! [[ "$disallowed_characters" =~ "filepath contains disallowed characters" ]];then
+bad_request=$(s3cmd -c s3cfg -q put s3cfg s3://test_dummy.org/fai\| 2>&1)
+if ! [[ "$bad_request" =~ "Bad Request" ]];then
     echo "test with disallowed characters failed"
     exit 1
 fi
@@ -34,13 +34,13 @@ aws configure set aws_access_key_id dummy # Value dont matter as we authenticate
 aws configure set aws_secret_access_key dummy # Value dont matter as we authenticate with the aws_session_token
 aws configure set aws_session_token "$(grep 'access_token' s3cfg | cut -d "=" -f 2 | xargs)"
 
-listResponse=$(aws s3api list-objects --endpoint http://s3inbox:8000 --bucket test_dummy.org)
+listResponse=$(aws s3api list-objects --endpoint-url http://s3inbox:8000 --bucket test_dummy.org)
 if (( $(echo "$listResponse" | grep -c "test_dummy.org") == 0 )); then
   echo "found no files for users when listing towards s3inbox with ListObjects call"
   exit 1
 fi
 
-listV2Response=$(aws s3api list-objects-v2 --endpoint http://s3inbox:8000 --bucket test_dummy.org)
+listV2Response=$(aws s3api list-objects-v2 --endpoint-url http://s3inbox:8000 --bucket test_dummy.org)
 if (( $(echo "$listV2Response" | grep -c "test_dummy.org") == 0 )); then
   echo "found no files for users when listing towards s3inbox with ListObjectsV2 call"
   exit 1
@@ -49,13 +49,13 @@ fi
 
 aws configure set aws_session_token "$(cat token)"
 
-listResponse=$(aws s3api list-objects --endpoint http://s3inbox:8000 --bucket testu_lifescience-ri.eu)
+listResponse=$(aws s3api list-objects --endpoint-url http://s3inbox:8000 --bucket testu_lifescience-ri.eu)
 if ! (( $(echo "$listResponse" | grep -c "test_dummy.org") == 0 )); then
   echo "found other users files when listing towards s3inbox with ListObjects call"
   exit 1
 fi
 
-listV2Response=$(aws s3api list-objects-v2 --endpoint http://s3inbox:8000 --bucket testu_lifescience-ri.eu)
+listV2Response=$(aws s3api list-objects-v2 --endpoint-url http://s3inbox:8000 --bucket testu_lifescience-ri.eu)
 if ! (( $(echo "$listV2Response" | grep -c "test_dummy.org") == 0 )); then
   echo "found other users files when listing towards s3inbox with ListObjectsV2 call"
   exit 1
--- a/.github/integration/tests/sda/15_s3cmd-errorcodes_test.sh
+++ b/.github/integration/tests/sda/15_s3cmd-errorcodes_test.sh
@@ -30,5 +30,36 @@ if ! [[ "$unathorized" =~ "Unauthorized" ]]; then
     exit 1
 fi
 
+aws configure set aws_access_key_id dummy # Value dont matter as we authenticate with the aws_session_token
+aws configure set aws_secret_access_key dummy # Value dont matter as we authenticate with the aws_session_token
+aws configure set aws_session_token "$(grep 'access_token' s3cfg | cut -d "=" -f 2 | xargs)"
+
+listResponse=$(aws s3api list-objects --endpoint http://s3inbox:8000 --bucket test_dummy.org)
+if (( $(echo "$listResponse" | grep -c "test_dummy.org") == 0 )); then
+  echo "found no files for users when listing towards s3inbox with ListObjects call"
+  exit 1
+fi
+
+listV2Response=$(aws s3api list-objects --endpoint http://s3inbox:8000 --bucket test_dummy.org)
+if (( $(echo "$listV2Response" | grep -c "test_dummy.org") == 0 )); then
+  echo "found no files for users when listing towards s3inbox with ListObjectsV2 call"
+  exit 1
+fi
+
+
+aws configure set aws_session_token "$(cat token)"
+
+listResponse=$(aws s3api list-objects --endpoint http://s3inbox:8000 --bucket testu_lifescience-ri.eu)
+if ! (( $(echo "$listResponse" | grep -c "test_dummy.org") == 0 )); then
+  echo "found other users files when listing towards s3inbox with ListObjects call"
+  exit 1
+fi
+
+listV2Response=$(aws s3api list-objects-v2 --endpoint http://s3inbox:8000 --bucket testu_lifescience-ri.eu)
+if ! (( $(echo "$listV2Response" | grep -c "test_dummy.org") == 0 )); then
+  echo "found other users files when listing towards s3inbox with ListObjectsV2 call"
+  exit 1
+fi
+
 
 echo "s3cmd error messages tested successfully"
--- a/.github/integration/tests/sda/15_s3cmd-errorcodes_test.sh
+++ b/.github/integration/tests/sda/15_s3cmd-errorcodes_test.sh
@@ -8,7 +8,7 @@ cd shared || true
 
 # test that listing all buckets is not allowed
 list_all_buckets=$(s3cmd -c s3cfg -q ls 2>&1)
-if ! [[ "$list_all_buckets" =~ "not allowed response" ]]; then
+if ! [[ "$list_all_buckets" =~ "Forbidden" ]]; then
     echo "list buckets should fail"
     exit 1
 fi
@@ -25,7 +25,7 @@ cp s3cfg bads3cfg
 sed -i "s/access_token=.*/access_token=invalid/" bads3cfg
 
 unathorized=$(s3cmd -c bads3cfg -q ls s3://test_dummy.org/ 2>&1)
-if ! [[ "$unathorized" =~ "not authorized" ]]; then
+if ! [[ "$unathorized" =~ "Unauthorized" ]]; then
     echo "testing unathorized call failed"
     exit 1
 fi
--- a/.github/integration/tests/sda/15_s3cmd-errorcodes_test.sh
+++ b/.github/integration/tests/sda/15_s3cmd-errorcodes_test.sh
@@ -40,7 +40,7 @@ if (( $(echo "$listResponse" | grep -c "test_dummy.org") == 0 )); then
   exit 1
 fi
 
-listV2Response=$(aws s3api list-objects --endpoint http://s3inbox:8000 --bucket test_dummy.org)
+listV2Response=$(aws s3api list-objects-v2 --endpoint http://s3inbox:8000 --bucket test_dummy.org)
 if (( $(echo "$listV2Response" | grep -c "test_dummy.org") == 0 )); then
   echo "found no files for users when listing towards s3inbox with ListObjectsV2 call"
   exit 1
--- a/sda/cmd/s3inbox/proxy.go
+++ b/sda/cmd/s3inbox/proxy.go
@@ -96,7 +96,8 @@ func NewProxy(s3conf config.S3InboxConf, s3Client *s3.Client, auth userauth.Auth
 func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	token, err := p.auth.Authenticate(r)
 	if err != nil {
-		p.notAuthorized(w, err.Error())
+		log.Warnf("unauthorized user attempted: method: %s, path: %s, query: %s", r.Method, r.URL.Path, r.URL.RawQuery)
+		reportErrorToClient(http.StatusUnauthorized, "Unauthorized", w)
 
 		return
 	}
@@ -110,24 +111,15 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	case PutObject, CreateMultiPartUpload, CompleteMultiPartUpload:
 		p.handleUpload(s3RequestType, w, r, token)
 	default:
-		p.notAllowedResponse(w, fmt.Sprintf("user: %s, attempted to do not allowed request: method: %s, path: %s, query: %s", token.Subject(), r.Method, r.URL.Path, r.URL.RawQuery))
+		log.Warnf("user: %s, attempted to do not allowed request: method: %s, path: %s, query: %s", token.Subject(), r.Method, r.URL.Path, r.URL.RawQuery)
+		reportErrorToClient(http.StatusForbidden, "Forbidden", w)
 	}
 }
 
 // Report 500 to the user, log the original error
-func (p *Proxy) internalServerError(w http.ResponseWriter, err string) {
-	log.Error(err)
+func (p *Proxy) internalServerError(w http.ResponseWriter, tokenSubject, httpMethod, path, query, err string) {
+	log.Errorf("user: %s, method: %s, path: %s, query: %s, encountered internal error: %s", tokenSubject, httpMethod, path, query, err)
 	reportErrorToClient(http.StatusInternalServerError, "Internal Error", w)
-}
-
-func (p *Proxy) notAllowedResponse(w http.ResponseWriter, err string) {
-	log.Warn(err)
-	reportErrorToClient(http.StatusForbidden, "Forbidden", w)
-}
-
-func (p *Proxy) notAuthorized(w http.ResponseWriter, err string) {
-	log.Warn(err)
-	reportErrorToClient(http.StatusUnauthorized, "Unauthorized", w)
 }
 
 // prepareForwardPathAndQuery prepares the new path and query to be used for the s3 request to be user specific when
@@ -201,13 +193,13 @@ func (p *Proxy) forwardRequest(s3RequestType S3RequestType, w http.ResponseWrite
 
 	s3Response, err := p.forwardRequestToBackend(r)
 	if err != nil {
-		p.internalServerError(w, fmt.Sprintf("forwarding error: %v", err))
+		p.internalServerError(w, token.Subject(), r.Method, r.URL.Path, r.URL.RawQuery, fmt.Sprintf("forwarding error: %v", err))
 
 		return
 	}
 
 	if err := p.forwardResponseToClient(s3Response, w); err != nil {
-		p.internalServerError(w, fmt.Sprintf("failed to forward response to client: %v", err))
+		p.internalServerError(w, token.Subject(), r.Method, r.URL.Path, r.URL.RawQuery, fmt.Sprintf("failed to forward response to client: %v", err))
 	}
 
 	_ = s3Response.Body.Close()
@@ -235,7 +227,7 @@ func (p *Proxy) handleUpload(s3RequestType S3RequestType, w http.ResponseWriter,
 
 	fileID, err := p.database.GetFileIDInInbox(r.Context(), username, filePath)
 	if err != nil {
-		p.internalServerError(w, fmt.Sprintf("failed to check/get existing file id from database: %v", err))
+		p.internalServerError(w, token.Subject(), r.Method, r.URL.Path, r.URL.RawQuery, fmt.Sprintf("failed to check/get existing file id from database: %v", err))
 
 		return
 	}
@@ -244,7 +236,7 @@ func (p *Proxy) handleUpload(s3RequestType S3RequestType, w http.ResponseWriter,
 	if fileID == "" {
 		fileID, err = p.database.RegisterFile(nil, p.s3Conf.Endpoint+"/"+p.s3Conf.Bucket, filePath, username)
 		if err != nil {
-			p.internalServerError(w, fmt.Sprintf("failed to register file in database: %v", err))
+			p.internalServerError(w, token.Subject(), r.Method, r.URL.Path, r.URL.RawQuery, fmt.Sprintf("failed to register file in database: %v", err))
 
 			return
 		}
@@ -256,7 +248,7 @@ func (p *Proxy) handleUpload(s3RequestType S3RequestType, w http.ResponseWriter,
 	if s3RequestType == PutObject || s3RequestType == CompleteMultiPartUpload {
 		isReupload, err = p.checkFileExists(r.Context(), s3FilePath)
 		if err != nil {
-			p.internalServerError(w, err.Error())
+			p.internalServerError(w, token.Subject(), r.Method, r.URL.Path, r.URL.RawQuery, err.Error())
 
 			return
 		}
@@ -264,7 +256,7 @@ func (p *Proxy) handleUpload(s3RequestType S3RequestType, w http.ResponseWriter,
 
 	s3Response, err := p.forwardRequestToBackend(r)
 	if err != nil {
-		p.internalServerError(w, fmt.Sprintf("forwarding error: %v", err))
+		p.internalServerError(w, token.Subject(), r.Method, r.URL.Path, r.URL.RawQuery, fmt.Sprintf("forwarding error: %v", err))
 
 		return
 	}
@@ -277,31 +269,31 @@ func (p *Proxy) handleUpload(s3RequestType S3RequestType, w http.ResponseWriter,
 	if s3Response.StatusCode == 200 && (s3RequestType == PutObject || s3RequestType == CompleteMultiPartUpload) {
 		message, err := p.CreateMessageFromRequest(r.Context(), token.Subject(), s3FilePath)
 		if err != nil {
-			p.internalServerError(w, err.Error())
+			p.internalServerError(w, token.Subject(), r.Method, r.URL.Path, r.URL.RawQuery, err.Error())
 
 			return
 		}
 		jsonMessage, err := json.Marshal(message)
 		if err != nil {
-			p.internalServerError(w, fmt.Sprintf("failed to marshal rabbitmq message to json: %v", err))
+			p.internalServerError(w, token.Subject(), r.Method, r.URL.Path, r.URL.RawQuery, fmt.Sprintf("failed to marshal rabbitmq message to json: %v", err))
 
 			return
 		}
 
 		if err = p.checkAndSendMessage(fileID, jsonMessage); err != nil {
-			p.internalServerError(w, fmt.Sprintf("broker error: %v", err))
+			p.internalServerError(w, token.Subject(), r.Method, r.URL.Path, r.URL.RawQuery, fmt.Sprintf("broker error: %v", err))
 
 			return
 		}
 
 		if err := p.storeObjectSizeInDB(r.Context(), s3FilePath, fileID); err != nil {
-			p.internalServerError(w, fmt.Sprintf("storeObjectSizeInDB failed because: %v", err))
+			p.internalServerError(w, token.Subject(), r.Method, r.URL.Path, r.URL.RawQuery, fmt.Sprintf("storeObjectSizeInDB failed because: %v", err))
 
 			return
 		}
 
 		if err := p.database.UpdateFileEventLog(fileID, "uploaded", "inbox", "{}", string(jsonMessage)); err != nil {
-			p.internalServerError(w, fmt.Sprintf("could not connect to db: %v", err))
+			p.internalServerError(w, token.Subject(), r.Method, r.URL.Path, r.URL.RawQuery, fmt.Sprintf("could not connect to db: %v", err))
 
 			return
 		}
@@ -309,7 +301,7 @@ func (p *Proxy) handleUpload(s3RequestType S3RequestType, w http.ResponseWriter,
 		if isReupload {
 			log.Infof("user: %s, reuploaded file: %s, with id: %s", username, filePath, fileID)
 			if err := p.sendMessageOnOverwrite(username, fileID, s3FilePath); err != nil {
-				p.internalServerError(w, err.Error())
+				p.internalServerError(w, token.Subject(), r.Method, r.URL.Path, r.URL.RawQuery, err.Error())
 
 				return
 			}
@@ -319,7 +311,7 @@ func (p *Proxy) handleUpload(s3RequestType S3RequestType, w http.ResponseWriter,
 	}
 
 	if err := p.forwardResponseToClient(s3Response, w); err != nil {
-		p.internalServerError(w, fmt.Sprintf("failed to forward response to client: %v", err))
+		p.internalServerError(w, token.Subject(), r.Method, r.URL.Path, r.URL.RawQuery, fmt.Sprintf("failed to forward response to client: %v", err))
 	}
 }
 
--- a/sda/cmd/s3inbox/proxy.go
+++ b/sda/cmd/s3inbox/proxy.go
@@ -157,6 +157,8 @@ func (p *Proxy) prepareForwardPathAndQuery(s3RequestType S3RequestType, originPa
 
 	var newPath, newQuery string
 	switch s3RequestType {
+	case GetBucketLocation:
+		newPath = "/" + p.s3Conf.Bucket
 	case ListObjects, ListObjectsV2, ListMultiPartUploads:
 		newPath = "/" + p.s3Conf.Bucket
 
--- a/sda/cmd/s3inbox/proxy.go
+++ b/sda/cmd/s3inbox/proxy.go
@@ -28,10 +28,6 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-type uniqueFileID struct {
-	username, filePath string
-}
-
 // Proxy represents the toplevel object in this application
 type Proxy struct {
 	s3Conf    config.S3InboxConf
@@ -40,7 +36,6 @@ type Proxy struct {
 	messenger *broker.AMQPBroker
 	database  *database.SDAdb
 	client    *http.Client
-	fileIDs   map[uniqueFileID]string
 }
 
 // The Event struct
@@ -70,15 +65,16 @@ type ErrorResponse struct {
 
 // The different types of requests
 const (
-	MakeBucket S3RequestType = iota
-	RemoveBucket
-	List
-	Put
-	Get
-	Delete
-	AbortMultipart
-	Policy
-	Other
+	Unsupported S3RequestType = iota
+	ListObjectsV2
+	ListObjects
+	PutObject
+	UploadPart
+	CreateMultiPartUpload
+	CompleteMultiPartUpload
+	ListMultiPartUpload
+	AbortMultiPartUpload
+	GetBucketLocation
 )
 
 // NewProxy creates a new S3Proxy. This implements the ServerHTTP interface.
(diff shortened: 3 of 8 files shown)


sda/internal/database/db_functions.go
```diff
@@ -118,6 +118,38 @@ WHERE id_and_event.event = $3;`
...
+SELECT id_and_event.id
+FROM (
+    SELECT DISTINCT ON (f.id) f.id, fel.event FROM sda.files AS f
+        LEFT JOIN sda.file_event_log AS fel ON fel.file_id = f.id
+    WHERE f.submission_user = $1
+      AND f.submission_file_path = $2
+      AND f.stable_id IS NULL
+    ORDER BY f.id, fel.started_at DESC LIMIT 1
+    ) AS id_and_event
+WHERE id_and_event.event IN ('registered', 'uploaded', 'disabled');`
```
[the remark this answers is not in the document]
KarlG-nbis: That is intended

sda/cmd/s3inbox/proxy.go
```diff
@@ -449,103 +390,121 @@ func (p *Proxy) resignHeader(r *http.Request) *http.Request {
...
+//
+//	"replication", "analytics", "metrics", "inventory", "ownershipControls", "publicAccessBlock", "object-lock"
+//
+// * PutObject == PUT /${bucket}/${object}
+// For aws docs see: https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html
+// partNumber and uploadId query arguments not present
+// We ensure x-amz-copy-source is not present to not allow CopyObject
+//
+// * UploadPart == PUT /${bucket}/${object}
+// For aws docs see:  https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPart.html
```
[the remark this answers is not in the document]
KarlG-nbis: N/A

sda/cmd/s3inbox/proxy.go
```diff
@@ -93,208 +90,224 @@ func NewProxy(s3conf config.S3InboxConf, s3Client *s3.Client, auth userauth.Auth
...
+	}
 
-		// check if the file already exists, in that case send an overwrite message,
-		// so that the FEGA portal is informed that a new version of the file exists.
-		err = p.sendMessageOnOverwrite(r, p.fileIDs[fileIdentifier], rawFilepath, token)
-		if err != nil {
-			p.internalServerError(w, r, err.Error())
+	// check if the file already exists when an upload completes, in that case send an overwrite message,
+	// so that the FEGA portal is informed that a new version
+	if s3RequestType == PutObject || s3RequestType == CompleteMultiPartUpload {
```
pahatz, line 248: Wasn't the second clause covered before somewhere?

sda/cmd/s3inbox/proxy.go
```diff
@@ -93,208 +90,224 @@ func NewProxy(s3conf config.S3InboxConf, s3Client *s3.Client, auth userauth.Auth
...
 
-func (p *Proxy) allowedResponse(w http.ResponseWriter, r *http.Request, token jwt.Token) {
-	log.Debug("prepend")
-	// Check whether token username and filepath match
-	str, err := url.ParseRequestURI(r.URL.Path)
-	if err != nil || str.Path == "" {
-		reportError(http.StatusBadRequest, err.Error(), w)
+// prepareForwardPathAndQuery prepares the new path and query to be used for the s3 request to be user specific when
+// reaching the s3 backend
+func (p *Proxy) prepareForwardPathAndQuery(s3RequestType S3RequestType, originPath, originQuery, tokenSubject string) (string, string, error) {
```
pahatz, line 127: This one could maybe benefit from some test, but the logic seems fine.

pahatz (dismissed):
Looks good to me.

sda/cmd/s3inbox/proxy.go
KarlG-nbis: Yeah its mainly based upon existing logic, but yeah some more dedicated unit test for this could be good, but could perhaps be done at a later point

sda/cmd/s3inbox/proxy.go
KarlG-nbis: it would have been covered by the first "UploadPart" (as that is also a PUT request) of a multi part upload

but imo more sensible to do it when the multipart upload completes as before that point the old object is still available

sda/cmd/s3inbox/proxy.go
```diff
@@ -93,208 +90,224 @@ func NewProxy(s3conf config.S3InboxConf, s3Client *s3.Client, auth userauth.Auth
...
 
-		// check if the file already exists, in that case send an overwrite message,
-		// so that the FEGA portal is informed that a new version of the file exists.
-		err = p.sendMessageOnOverwrite(r, p.fileIDs[fileIdentifier], rawFilepath, token)
-		if err != nil {
-			p.internalServerError(w, r, err.Error())
+	// check if the file already exists when an upload completes, in that case send an overwrite message,
+	// so that the FEGA portal is informed that a new version
+	if s3RequestType == PutObject || s3RequestType == CompleteMultiPartUpload {
+		if err := p.sendMessageOnOverwrite(r.Context(), username, fileID, s3FilePath, filePath); err != nil {
```
nanjiangshu: There might be a problem about the timing of `sendMessageOnOverwrite`. 
For example, if the message is sent, but S3 upload fails (due to e.g., a 500 error from the storage), then the "remove" message is sent to the broker, however the S3 storage still has the old version of the file.

It is probably better to move  `p.sendMessageOnOverwrite` after `s3Response.StatusCode == 200`.

kostas-kou (dismissed):

sda/cmd/s3inbox/proxy.go
KarlG-nbis: Yeah that is true, but will require some additional code changes, as we need to check if it will be an overwrite before the successful upload as otherwise the HeadObject check will just find the newly uploaded object

KarlG-nbis: Fixed in commit: https://github.com/neicnordic/sensitive-data-archive/commit/03b3d362bc79afd17cdbd91320219a21278a416f

sda/cmd/s3inbox/proxy.go
```diff
@@ -93,208 +90,236 @@ func NewProxy(s3conf config.S3InboxConf, s3Client *s3.Client, auth userauth.Auth
...
 		return
 	}
 
 	// Send message to upstream and set file as uploaded in the database
 	// nolint: nestif // We need a nested if statement for checking whether fileId is persisted during possible reconnections
-	if p.uploadFinishedSuccessfully(r, s3response) {
-		log.Debug("create message")
-		message, err := p.CreateMessageFromRequest(r, token, anonymizedFilepath)
+	if s3Response.StatusCode == 200 && (s3RequestType == PutObject || s3RequestType == CompleteMultiPartUpload) {
+		message, err := p.CreateMessageFromRequest(r.Context(), token.Subject(), s3FilePath)
```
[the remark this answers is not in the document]
KarlG-nbis: This is as expected and we want the full path including the username(with @ -> _ replacement) in the inbox message

nanjiangshu (dismissed):
Looks good to me now. 

A small tip for `updating the branch to main` but without creating an additional commit 
e.g. d4fc6c5

If you run the following commands instead of pressing `update branch` on Github, no additional merge commit will be created. 

```bash
    git checkout feature-branch
    git fetch origin main
    git rebase origin/main
    git push --force 
```

sda/cmd/s3inbox/proxy.go
```diff
@@ -93,208 +90,237 @@ func NewProxy(s3conf config.S3InboxConf, s3Client *s3.Client, auth userauth.Auth
...
+	log.Warn(err)
+	reportError(http.StatusForbidden, "Forbidden", w)
 }
 
-func (p *Proxy) notAuthorized(w http.ResponseWriter, _ *http.Request) {
-	reportError(http.StatusUnauthorized, "not authorized", w)
+func (p *Proxy) notAuthorized(w http.ResponseWriter, err string) {
+	log.Warn(err)
+	reportError(http.StatusUnauthorized, "Unauthorized", w)
 }
```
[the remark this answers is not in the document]
jbygdell: @KarlG-nbis probably need to fix this.
Two errors are logged one, or `warn` and the other [on `error` level](https://github.com/neicnordic/sensitive-data-archive/blob/7b8a6aea7eda63e2f6354a4f44b85a5036c681ff/sda/cmd/s3inbox/proxy.go#L596)

jbygdell (dismissed):

.github/integration/tests/sda/15_s3cmd-errorcodes_test.sh
```diff
@@ -25,10 +25,41 @@ cp s3cfg bads3cfg
...
+aws configure set aws_secret_access_key dummy # Value dont matter as we authenticate with the aws_session_token
+aws configure set aws_session_token "$(grep 'access_token' s3cfg | cut -d "=" -f 2 | xargs)"
+
+listResponse=$(aws s3api list-objects --endpoint http://s3inbox:8000 --bucket test_dummy.org)
+if (( $(echo "$listResponse" | grep -c "test_dummy.org") == 0 )); then
+  echo "found no files for users when listing towards s3inbox with ListObjects call"
+  exit 1
+fi
+
+listV2Response=$(aws s3api list-objects --endpoint http://s3inbox:8000 --bucket test_dummy.org)
```
jhagberg: This block looks like a copy-paste from the V1 check above — the variable name and error message say V2 but the command is still `list-objects` (V1), so the V2 self-listing case never actually exercises V2. The `list-objects-v2` block further down is fine.

```suggestion
listV2Response=$(aws s3api list-objects-v2 --endpoint http://s3inbox:8000 --bucket test_dummy.org)
```

.github/integration/tests/sda/15_s3cmd-errorcodes_test.sh
```diff
@@ -25,10 +25,41 @@ cp s3cfg bads3cfg
...
+aws configure set aws_session_token "$(grep 'access_token' s3cfg | cut -d "=" -f 2 | xargs)"
+
+listResponse=$(aws s3api list-objects --endpoint http://s3inbox:8000 --bucket test_dummy.org)
+if (( $(echo "$listResponse" | grep -c "test_dummy.org") == 0 )); then
+  echo "found no files for users when listing towards s3inbox with ListObjects call"
+  exit 1
+fi
+
+listV2Response=$(aws s3api list-objects-v2 --endpoint http://s3inbox:8000 --bucket test_dummy.org)
+if (( $(echo "$listV2Response" | grep -c "test_dummy.org") == 0 )); then
```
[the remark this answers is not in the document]
KarlG-nbis: N/A

.github/integration/tests/sda/15_s3cmd-errorcodes_test.sh
```diff
@@ -25,10 +25,41 @@ cp s3cfg bads3cfg
...
+aws configure set aws_session_token "$(cat token)"
+
+listResponse=$(aws s3api list-objects --endpoint http://s3inbox:8000 --bucket testu_lifescience-ri.eu)
+if ! (( $(echo "$listResponse" | grep -c "test_dummy.org") == 0 )); then
+  echo "found other users files when listing towards s3inbox with ListObjects call"
+  exit 1
+fi
+
+listV2Response=$(aws s3api list-objects-v2 --endpoint http://s3inbox:8000 --bucket testu_lifescience-ri.eu)
+if ! (( $(echo "$listV2Response" | grep -c "test_dummy.org") == 0 )); then
```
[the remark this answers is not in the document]
KarlG-nbis: N/A

.github/integration/tests/sda/01_install_dependencies.sh
```diff
@@ -2,7 +2,7 @@
 set -e
 
 # install tools if missing
-for t in curl expect jq openssh-client postgresql-client xxd; do
+for t in curl expect jq openssh-client postgresql-client xxd awscli; do
     if [ ! "$(command -v $t)" ]; then
```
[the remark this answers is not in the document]
KarlG-nbis: N/A

sda/cmd/s3inbox/proxy.go
```diff
@@ -93,208 +90,237 @@ func NewProxy(s3conf config.S3InboxConf, s3Client *s3.Client, auth userauth.Auth
...
+
+	return newPath, newQuery, nil
+}
+
+// forwardRequest forwards the request to the s3 backend after making request user specific, then forwards response to client
+func (p *Proxy) forwardRequest(s3RequestType S3RequestType, w http.ResponseWriter, r *http.Request, token jwt.Token) {
+	var err error
+	r.URL.Path, r.URL.RawQuery, err = p.prepareForwardPathAndQuery(s3RequestType, r.URL.Path, r.URL.RawQuery, token.Subject())
+	if err != nil {
+		reportError(http.StatusBadRequest, err.Error(), w)
```
jhagberg: Since 39b5bfd7 removed the `log.Error` from `reportError`, these BadRequest paths don't log anything server-side anymore. But `err.Error()` still reaches the client in the XML body — which is the opposite of what the PR is aiming for (generic errors to client, detail in logs). Could we log + sanitize the client message, something like:

```go
if err != nil {
    log.Warnf("bad request from user %s: %v", token.Subject(), err)
    reportError(http.StatusBadRequest, "Bad Request", w)
    return
}
```

Same thing at lines 219 and 227.

.github/integration/tests/sda/15_s3cmd-errorcodes_test.sh
```diff
@@ -25,10 +25,41 @@ cp s3cfg bads3cfg
...
+
+aws configure set aws_session_token "$(cat token)"
+
+listResponse=$(aws s3api list-objects --endpoint http://s3inbox:8000 --bucket testu_lifescience-ri.eu)
+if ! (( $(echo "$listResponse" | grep -c "test_dummy.org") == 0 )); then
+  echo "found other users files when listing towards s3inbox with ListObjects call"
+  exit 1
+fi
+
+listV2Response=$(aws s3api list-objects-v2 --endpoint http://s3inbox:8000 --bucket testu_lifescience-ri.eu)
```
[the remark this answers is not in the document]
KarlG-nbis: N/A

.github/integration/tests/sda/15_s3cmd-errorcodes_test.sh
```diff
@@ -25,10 +25,41 @@ cp s3cfg bads3cfg
...
+aws configure set aws_secret_access_key dummy # Value dont matter as we authenticate with the aws_session_token
+aws configure set aws_session_token "$(grep 'access_token' s3cfg | cut -d "=" -f 2 | xargs)"
+
+listResponse=$(aws s3api list-objects --endpoint http://s3inbox:8000 --bucket test_dummy.org)
+if (( $(echo "$listResponse" | grep -c "test_dummy.org") == 0 )); then
+  echo "found no files for users when listing towards s3inbox with ListObjects call"
+  exit 1
+fi
+
+listV2Response=$(aws s3api list-objects-v2 --endpoint http://s3inbox:8000 --bucket test_dummy.org)
```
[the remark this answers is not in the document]
KarlG-nbis: N/A

.github/integration/tests/sda/15_s3cmd-errorcodes_test.sh
```diff
@@ -25,10 +25,41 @@ cp s3cfg bads3cfg
...
+if ! [[ "$unathorized" =~ "Unauthorized" ]]; then
     echo "testing unathorized call failed"
     exit 1
 fi
 
+aws configure set aws_access_key_id dummy # Value dont matter as we authenticate with the aws_session_token
+aws configure set aws_secret_access_key dummy # Value dont matter as we authenticate with the aws_session_token
+aws configure set aws_session_token "$(grep 'access_token' s3cfg | cut -d "=" -f 2 | xargs)"
+
+listResponse=$(aws s3api list-objects --endpoint http://s3inbox:8000 --bucket test_dummy.org)
```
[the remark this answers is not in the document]
KarlG-nbis: N/A

sda/cmd/s3inbox/proxy.go
```diff
@@ -93,208 +90,241 @@ func NewProxy(s3conf config.S3InboxConf, s3Client *s3.Client, auth userauth.Auth
...
 	default:
-		log.Debugf("Unexpected request (%v) not allowed", r)
-		p.notAllowedResponse(w, r)
+		p.notAllowedResponse(w, fmt.Sprintf("user: %s, attempted to do not allowed request: method: %s, path: %s, query: %s", token.Subject(), r.Method, r.URL.Path, r.URL.RawQuery))
 	}
 }
 
 // Report 500 to the user, log the original error
-func (p *Proxy) internalServerError(w http.ResponseWriter, r *http.Request, err string) {
+func (p *Proxy) internalServerError(w http.ResponseWriter, err string) {
```
jhagberg: With `reportErrorToClient` keeping the client response generic, the log line is the one place ops has to work from when a 500 hits. Most callers include the operation (`"broker error: %v"`, `"storeObjectSizeInDB failed: %v"`) but none include the user or file path, so it's hard to tell whose upload failed in prod. Would it be too noisy to thread `username` and `filePath` into each call site's Sprintf, or have you considered and decided against?

KarlG-nbis: Yeah makes sense otherwise these logs will not mean much to the ops! Will update

sda/internal/database/db_functions.go
```diff
@@ -118,6 +118,38 @@ WHERE id_and_event.event = $3;`
...
+// as that means that ingestion has not been triggered and users are allowed continue uploading or reupload the file to the inbox
+// if no row is found does not return sql.ErrNoRows, just empty string in id return field
+func (dbs *SDAdb) GetFileIDInInbox(ctx context.Context, submissionUser, filePath string) (string, error) {
+	dbs.checkAndReconnectIfNeeded()
+	db := dbs.DB
+
+	const getFileID = `
+SELECT id_and_event.id
+FROM (
+    SELECT DISTINCT ON (f.id) f.id, fel.event FROM sda.files AS f
```
jhagberg, line 131: Think there's a subtle ordering issue here. `DISTINCT ON (f.id) ... ORDER BY f.id, fel.started_at DESC` picks the latest event per file (good), but `LIMIT 1` on the inner query then picks the file with the lowest `f.id` under the same `ORDER BY`. Since `files.id` is `gen_random_uuid()` (v4, random), that's essentially UUID roulette.

The `unique_ingested` constraint (`UNIQUE(submission_file_path, archive_file_path, submission_user)`) lets two rows coexist for the same (user, path) when `archive_file_path` differs. If an abandoned inbox row and a newer one both have `stable_id IS NULL`, we could return the older one.

Safest fix I can think of: carry `fel.started_at` out of the inner query and do the `LIMIT 1` on the outer, ordered by timestamp:

```sql
SELECT id_and_event.id
FROM (
    SELECT DISTINCT ON (f.id) f.id, fel.event, fel.started_at
    FROM sda.files AS f
    LEFT JOIN sda.file_event_log AS fel ON fel.file_id = f.id
    WHERE f.submission_user = $1
      AND f.submission_file_path = $2
      AND f.stable_id IS NULL
    ORDER BY f.id, fel.started_at DESC
) AS id_and_event
WHERE id_and_event.event IN ('registered', 'uploaded', 'disabled')
ORDER BY id_and_event.started_at DESC
LIMIT 1;
```

A test with two rows would lock the behavior in either way.

sda/internal/database/db_functions.go
```diff
@@ -118,6 +118,38 @@ WHERE id_and_event.event = $3;`
 	return fileID, nil
 }
 
+// GetFileIDInInbox gets the file id of a file which last known event is either 'registered', 'uploaded', or 'disabled'
+// as that means that ingestion has not been triggered and users are allowed continue uploading or reupload the file to the inbox
+// if no row is found does not return sql.ErrNoRows, just empty string in id return field
+func (dbs *SDAdb) GetFileIDInInbox(ctx context.Context, submissionUser, filePath string) (string, error) {
```
jhagberg, line 124: Every other DB helper in this file wraps its query in a retry loop (see `GetFileIDByUserPathAndStatus`). Without one here, a single transient connection glitch during upload turns into a 500 for the user. And since this now runs on every upload, a DB flap hits more requests than it used to. Worth matching the existing pattern?

sda/internal/database/db_functions_test.go
```diff
@@ -1904,3 +1904,26 @@ func (suite *DatabaseTests) TestSetBackedUp_FileID_Not_Exists() {
...
+	assert.NoError(suite.T(), err, "got %v when creating new connection", err)
+	defer db.Close()
+
+	fileID, err := db.RegisterFile(nil, "/inbox", "TestGetFileIDInInbox.c4gh", "testuser")
+	assert.NoError(suite.T(), err, "failed to register file in database")
+
+	fileIDFromDB, err := db.GetFileIDInInbox(context.TODO(), "testuser", "TestGetFileIDInInbox.c4gh")
+	assert.NoError(suite.T(), err)
+	assert.Equal(suite.T(), fileID, fileIDFromDB)
+}
```
jhagberg, line 1931: The two new cases cover happy path and not-found, but the filter is doing a lot of work that isn't exercised. A few cases I'd like to see, since `stable_id IS NULL` and the event-list filter are load-bearing:

- Latest event is `archived` / `ingested` → should return `""` (otherwise a user could re-upload an already-ingested file)
- `stable_id IS NOT NULL` → should return `""`
- Same path, different user → user-isolation check
- `disabled` event → should return the fileID (doc comment lists this case but nothing tests it)
- Two rows with (user, path, stable_id=NULL) → verifies DISTINCT ON / LIMIT 1 picks the right one (see my separate comment about ordering)

`TestGetFileIDByUserPathAndStatus` walks `registered` to `archived` as a model; this could follow the same shape.

jhagberg:
Checked this against ADR-0003 (#2319) and issue #2375:

- Map and (would-have-been) mutex removed. Good.
- ADR criterion "existing `proxy_test.go` tests pass unmodified": the test file is rewritten here (+44/-41), so we don't actually know whether the old tests would still pass against the new code.
- ADR criterion "concurrent upload test reproducing #2295": not added. This is the single regression guard that proves the crash can't come back. Around 30 lines with two goroutines and `sync.WaitGroup` would pin down the whole point of this PR.
- ADR specifically proposed `GetFileIDByUserPathAndStatus`, but we ended up with a new `GetFileIDInInbox` with different filter semantics. Probably the right call for this use case, but since the ADR is now accepted, a sentence in the PR description or a follow-up ADR amendment would help future-me trace the deviation.

sda/internal/database/db_functions.go
KarlG-nbis: A file which is "in the inbox" should never have stable_id == null, but yeah might be safer to update to `AND f.archive_file_path == ''` instead of `AND f.stable_id IS NULL`

sda/cmd/s3inbox/proxy.go
```diff
@@ -93,208 +90,241 @@ func NewProxy(s3conf config.S3InboxConf, s3Client *s3.Client, auth userauth.Auth
...
-	// to the s3cmd client.
-	// Writing 200 here breaks uploads though, and writing non-200 codes after
-	// the headers results in the error message always being
-	// "MD5 Sums don't match!".
-	if s3response.StatusCode < 200 || s3response.StatusCode > 299 {
-		w.WriteHeader(s3response.StatusCode)
-	}
+		if isReupload {
+			log.Infof("user: %s, reuploaded file: %s, with id: %s", username, filePath, fileID)
+			if err := p.sendMessageOnOverwrite(username, fileID, s3FilePath); err != nil {
```
jhagberg, line 303: Two questions on the new overwrite flow:

1. **Semantics**: the upload message above and the `InboxRemove` here both reference the same `fileID`. From the FEGA-portal consumer side, does `upload` followed by `remove` for the same fileID read as "version replaced" or "file deleted after upload"? Worth sanity-checking with whoever owns that consumer.

2. **Partial commit on failure**: if `sendMessageOnOverwrite` errors here, we return 500, but by then the S3 upload is committed, the upload MQ message has been sent, the size is persisted, and the `uploaded` event is written. The client retries, and now everything upstream sees duplicates. Could this be non-fatal (log and continue), or at least documented that a 500 at this point means the retry will produce duplicate events?

ADR-0003 flagged duplicate overwrite messages as out of scope, so if we're changing behavior here it might be cleaner as a separate PR with the decision visible in git history.

sda/internal/helper/helper.go
```diff
@@ -90,7 +90,12 @@ func NewAlwaysAllow() *AlwaysAllow {
...
-	return jwt.New(), nil
+	token := jwt.New()
+	if err := token.Set("sub", "dummy"); err != nil {
+		return nil, err
+	}
+
+	return token, nil
 }
 
 // AlwaysAllow is an Authenticator that always authenticates
```
jhagberg: Drive-by fix: godoc above says `AlwaysAllow` on the `AlwaysDeny` type. Pre-existing, but visible because you're editing the line just above.

```suggestion
// AlwaysDeny is an Authenticator that always denies authentication
```

sda/internal/database/db_functions.go
jbygdell: The `unique _ingested` is not working as intended since a file is not `locked` unless it is mapped to a dataset.
And the reason for this is du to how the FEGA portal handles things internally.

sda/cmd/s3inbox/proxy.go
```diff
@@ -93,208 +90,241 @@ func NewProxy(s3conf config.S3InboxConf, s3Client *s3.Client, auth userauth.Auth
...
+		p.internalServerError(w, fmt.Sprintf("forwarding error: %v", err))
 
 		return
 	}
+	defer func() {
+		_ = s3Response.Body.Close()
+	}()
 
 	// Send message to upstream and set file as uploaded in the database
 	// nolint: nestif // We need a nested if statement for checking whether fileId is persisted during possible reconnections
```
jhagberg: The "possible reconnections" rationale doesn't really fit anymore. That was about the old in-memory cache missing on a different pod, but with the DB lookup at the top of `handleUpload`, the nested block is there for "only do broker and event-log side effects on successful PutObject / CompleteMultipartUpload". A one-liner update would save a future reader some head-scratching.

sda/internal/database/db_functions.go
KarlG-nbis: Updated to use `f.archive_file_path = ''` as then we can rely on unique_ingested constraint to return a unique file which should be "in the inbox" based on submission_user and submission_file_path

sda/cmd/s3inbox/proxy.go
```diff
@@ -558,48 +533,34 @@ func (p *Proxy) requestInfo(ctx context.Context, fullPath string) (string, int64
...
 	result, err := p.s3Client.HeadObject(ctx, &s3.HeadObjectInput{
 		Bucket: aws.String(p.s3Conf.Bucket),
-		Key:    aws.String(filePath),
+		Key:    aws.String(s3FilePath),
 	})
 
 	if err != nil && strings.Contains(err.Error(), "StatusCode: 404") {
-		log.Debugf("s3 could not find file %s: %s", fullPath, err.Error())
-
 		return false, nil
```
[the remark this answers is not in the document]
KarlG-nbis: Out of scope of this PR

sda/internal/database/db_functions_test.go
KarlG-nbis: Updated to add additional checks when latest file event changes, but given update of `f.archive_file_path = ''`use in `GetFileIDInInbox` excluded the other suggestions

sda/internal/database/db_functions.go
KarlG-nbis: imo the currently retry logic is not very sound which is why I choose to exclude it, it would make sense to have some retry logic, but it would need improvement such that it only retries connection issues, etc, so it doesnt spend time retrying queries which will never succeed

sda/cmd/s3inbox/proxy.go
```diff
@@ -449,103 +403,123 @@ func (p *Proxy) resignHeader(r *http.Request) *http.Request {
...
+		return ListObjectsV2
+	case r.Method == http.MethodGet && isBucketPath && query.Has("uploads"):
+		return ListMultiPartUploads
+	case r.Method == http.MethodGet && isObjectPath && query.Has("uploadId"):
+		return ListParts
+	case r.Method == http.MethodGet && isBucketPath && query.Has("location"):
+		return GetBucketLocation
+	case r.Method == http.MethodGet && isBucketPath:
+		return ListObjects
+	case r.Method == http.MethodPut && isObjectPath && !query.Has("partNumber") && !query.Has("uploadId") && r.Header.Get("x-amz-copy-source") == "":
```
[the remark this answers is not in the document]
KarlG-nbis: That is the point of the first case to exclude any request containg query args that would indicate another s3 actions, as the ListObjects do not have any specific required query arg to be able to identify it.

sda/internal/database/db_functions.go
jhagberg: You're right, I went and read the retry pattern in `GetFileIDByUserPathAndStatus`. It retries every error except `sql.ErrNoRows`, so syntax errors, permission errors, and constraint violations all cost the user 62 seconds of backoff before the handler gives up. And it uses `time.Sleep` rather than being context-aware, so retries keep running after the client has disconnected. Your call to skip it makes sense.

Would it be worth opening a follow-up issue to track a proper rework of the retry pattern across this file? Matching `errors.Is(err, driver.ErrBadConn)` (or the pq equivalent) with context-awareness would let future helpers retry only the cases that actually benefit, without inheriting the current smell.

sda/cmd/s3inbox/proxy.go
```diff
@@ -315,123 +336,56 @@ func (p *Proxy) checkAndSendMessage(fileID string, jsonMessage []byte, r *http.R
...
+	nr, err := http.NewRequest(r.Method, p.s3Conf.Endpoint+r.URL.String(), r.Body) // #nosec G704 -- endpoint and port controlled by configuration
 	if err != nil {
-		log.Debug("error when redirecting the request")
-		log.Debug(err)
-
 		return nil, err
 	}
 	nr.Header = r.Header
 	contentLength, _ := strconv.ParseInt(r.Header.Get("content-length"), 10, 64)
 	nr.ContentLength = contentLength
```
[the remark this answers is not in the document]
jhagberg: This thread got marked resolved without a reply, just wanted to check the reasoning before letting it go. Is the 30s client timeout (line 84) already covering the cleanup case for you, or is context propagation planned for a follow-up? Asking because now that the logs include operation context, a context-cancelled error would actually surface differently than a 30s S3 timeout in the new log line, which seems useful for ops.

sda/cmd/s3inbox/proxy.go
```diff
@@ -93,208 +90,233 @@ func NewProxy(s3conf config.S3InboxConf, s3Client *s3.Client, auth userauth.Auth
...
 
-	// Redirect answer
-	log.Debug("redirect answer")
-	for header, values := range s3response.Header {
-		for _, value := range values {
-			w.Header().Add(header, value)
+				return
+			}
+		} else {
+			log.Infof("user: %s, uploaded file: %s, with id: %s", username, filePath, fileID)
```
jhagberg: Audit-trail follow-up: the old log line at this spot included the file's S3 checksum and the `pilot` private claim. The new `Infof` adds `fileID` (good) but drops both. For an SDA, the upload-success line is the non-repudiation record proving exactly which bytes landed at what time, and `pilot` tells us which project/tenant the user was acting on behalf of. Losing them in a refactor reads as a compliance regression worth a deliberate decision either way.

Both values are already in scope at this point. `message.Checksum[0].(Checksum).Value` carries the checksum from `CreateMessageFromRequest` on line 270 (which already did the HeadObject), and `pilot` reads off the token:

```go
pilot, _ := token.Get("pilot")
checksum := message.Checksum[0].(Checksum).Value
log.Infof("user: %s, pilot: %v, uploaded file: %s, with id: %s, checksum: %s",
    username, pilot, filePath, fileID, checksum)
```

The reupload log on line 302 has the same gap. Was the `pilot` removal deliberate?

sda/internal/database/db_functions.go
jbygdell: From the FEGA sie we have had this issue a few times.

The user uploads a file and triggers ingestion.
Something fails and the file is stuck in ex. archived state.
The user then uploads a new version of the file
If the file then gets a new ID in the DB the user sees `two` files in the FEGA portal. And the they have  problem picking the `correct version` of the file since only the path is visible

sda/cmd/s3inbox/proxy.go
KarlG-nbis: Ah yeah checksum would be worth keeping, but for pilot, mainly removed it since when checking the logs it was alwas logged as `pilot \\u003cnil\\u003e`

sda/internal/database/db_functions.go
KarlG-nbis: yeah created #2391 for this

sda/cmd/s3inbox/proxy.go
KarlG-nbis: 1. Yeah not sure, but my guess it would correlate to "file deleted after upload" as we publish a new "upload" message with the new checksum, etc when the new upload completes.

2. I think this is quite an edge case that could be worth capturing, but if `sendMessageOnOverwrite` failed most likely `checkAndSendMessage` would also fail which should be a error, but yeah in general these edge case scenarios and the recover / rollback logic could be worth looking at but imo quite low priority atm if we dont see it happening often

sda/internal/database/db_functions.go
jhagberg: Quick follow-up: I ran a direct-INSERT test against the constraint, and PostgreSQL rejected it with `pq: duplicate key value violates unique constraint "unique_ingested"`. So at the DB level the single-row assumption holds for (user, path, archive_file_path=''), and Karl's filter is safe in that sense.

Joakim's scenario is real and deserves its own ticket though. If a file is stuck in `archived` after failed ingestion and the user re-uploads, the new INSERT creates a fresh row with a new fileID and the FEGA portal ends up with two entries. That's an upstream UX/semantics question (should re-upload reuse the stuck file's ID, and if so what happens to its stable_id and dataset mappings?) that I don't think this PR should try to solve. Worth opening a separate issue so the trade-offs can be discussed properly.

sda/cmd/s3inbox/proxy.go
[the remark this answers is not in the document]
jhagberg: @KarlG-nbis ?

KarlG-nbis:
> Checked this against ADR-0003 (#2319) and issue #2375:
> 
>     * Map and (would-have-been) mutex removed. Good.
> 
>     * ADR criterion "existing `proxy_test.go` tests pass unmodified": the test file is rewritten here (+44/-41), so we don't actually know whether the old tests would still pass against the new code.

Yeah was updated to check the new s3 action detections

>     * ADR criterion "concurrent upload test reproducing [[s3inbox] port v3.1.0 hotfix for concurrent map read / write to main #2295](https://github.com/neicnordic/sensitive-data-archive/issues/2295)": not added. This is the single regression guard that proves the crash can't come back. Around 30 lines with two goroutines and `sync.WaitGroup` would pin down the whole point of this PR.

This was not added since constructing such a unit test that could consistently reproduce that issue is not feasible


>     * ADR specifically proposed `GetFileIDByUserPathAndStatus`, but we ended up with a new `GetFileIDInInbox` with different filter semantics. Probably the right call for this use case, but since the ADR is now accepted, a sentence in the PR description or a follow-up ADR amendment would help future-me trace the deviation.

Updated PR description to mention this

sda/cmd/s3inbox/proxy.go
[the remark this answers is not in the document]
KarlG-nbis: Main reasoning was that this felt out of scope of the PR. But yeah would make sense to address just not sure if this PR is the right place for it, ill create a follow up issue for it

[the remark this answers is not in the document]
KarlG-nbis: Created https://github.com/neicnordic/sensitive-data-archive/issues/2393

.github/integration/tests/sda/15_s3cmd-errorcodes_test.sh
```diff
@@ -25,10 +25,41 @@ cp s3cfg bads3cfg
 sed -i "s/access_token=.*/access_token=invalid/" bads3cfg
 
 unathorized=$(s3cmd -c bads3cfg -q ls s3://test_dummy.org/ 2>&1)
-if ! [[ "$unathorized" =~ "not authorized" ]]; then
+if ! [[ "$unathorized" =~ "Unauthorized" ]]; then
     echo "testing unathorized call failed"
     exit 1
```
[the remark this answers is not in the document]
KarlG-nbis: out of scope

sda/cmd/s3inbox/proxy.go
```diff
@@ -629,8 +590,7 @@ func formatUploadFilePath(filePath string) (string, error) {
 }
 
 // Write the error and its status code to the response
-func reportError(errorCode int, message string, w http.ResponseWriter) {
-	log.Error(message)
+func reportErrorToClient(errorCode int, message string, w http.ResponseWriter) {
 	errorResponse := ErrorResponse{
 		Code:    http.StatusText(errorCode),
 		Message: message,
```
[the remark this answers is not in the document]
KarlG-nbis: out of scope

sda/cmd/s3inbox/proxy_test.go
```diff
@@ -401,73 +442,65 @@ func (s *ProxyTests) TestServeHTTP_allowed() {
...
 	r.Method = "GET"
-	r.URL, _ = url.Parse("/dummy/file?delimiter=puppe")
+	r.URL, _ = url.Parse("/dummy?delimiter=puppe")
 	w = httptest.NewRecorder()
-	proxy.allowedResponse(w, r, s.token)
+	proxy.ServeHTTP(w, r)
 	assert.Equal(s.T(), 200, w.Result().StatusCode)
 	assert.Equal(s.T(), true, s.fakeServer.PingedAndRestore())
 
 	// Show multiparts uploads
```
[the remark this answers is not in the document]
KarlG-nbis: N/A

sda/cmd/s3inbox/proxy.go
jbygdell: The FEGA portal uses the message Correlation ID (DB file UUID) to keep track of things.
If a file is reuploaded during an ongoing submission the same Correlation IDs must be used.
The `sendMessageOnOverwrite` ensures that the FEGA portal sees a `remove` message before the new `uploaded` message. This is needed since otherwise the timestamps and stuff don't update in the portal.

As such a file can be reuploaded and reset to `uploaded` state as long as it has not been mapped to a dataset.

jhagberg (approved):
Approving. The refactor does what it set out to do: the in-memory `fileIDs` map is gone, the concurrent-upload test from 6e67bfef reproduces #2295 on main and passes cleanly here, and the supporting tests cover the `archived → cancelled → reupload` transitions.

A few things I appreciated along the way, for future reference rather than blockers:

- The `reportError` → `reportErrorToClient` rename was a cleaner split than what I originally suggested. It made the log/client boundary obvious at every call site.
- Adding user/method/path/query to `internalServerError` went further than my original nit asked for. 500 logs now have real operational context.
- Swapping to `archive_file_path = ''` used the existing unique constraint to guarantee single-row semantics without needing the ordering fix I was worried about.

Follow-ups tracked separately: #2391 (retry rework) and #2393 (forward client context to s3 request). Happy to see this ship. 🎉

jbygdell (approved):
Approving as is.

We might need to redo the `reupload` logic after testing in the FEGA scope. Since a file can be reuploaded and reset to uploaded state as long as it has not been mapped to a dataset according to the current working model there.

PR #2384, opened 2026-04-13
feat(chart): add v2 download service deployment alongside legacy

jhagberg:
## Related issue(s) and PR(s)
This PR closes #2345.

## Description

Adds Helm chart support for running the v2 download service in parallel with the legacy one during migration. The v2 service is opt-in (`global.downloadV2.enabled: false` by default) so existing deployments are unaffected.

**6 new templates** in `charts/sda-svc/templates/download-v2-*`: deployment, secrets, service, ingress, certificate, network policy. Each mirrors the structure of its legacy counterpart but adapted for v2's config format.

Key differences from legacy that the templates handle:
- Config uses flat pflag/viper keys (`api.port`) instead of nested YAML (`app.port`)
- Binary loaded via `--config-file` flag, not the `CONFIGFILE` env var
- Health probes on `/health/live` + `/health/ready`, rendered via `toYaml` from values
- TLS terminates at ingress — container always listens on port 8080
- Storage section mapped from `global.archive.*` into v2's `storage.archive.s3/posix` format
- gRPC host defaults to `{{ sda.fullname }}-reencrypt` (only needs override for split deployments)
- JWT public keys mounted via `global.oidc.jwtSecret` when set (reuses OIDC definition instead of a separate downloadV2-specific JWT block)
- No c4gh transient key volume (v2 removed decrypted serving)
- Network policy uses `sda.fullname` in podSelector (avoids an existing legacy bug with `sda.name`)
- Ingress hostname is `required` to prevent accidental catch-all rules
- DB credentials shared with legacy download (same scope and permissions)
- OIDC config read from `global.oidc`
- Helm-level guards: `required` on S3 credentials and service org-name/url, `fail` when `replicaCount > 1` without `pagination.hmacSecret`, `fail` when `visa.enabled` without `trustedIssuers`

Also includes:
- `@param` annotations on all new values for helm-docs compatibility (README update tracked in #2398)
- k3d smoke test script in `dev-tools/chart-smoke-test/` with 16-combination template render matrix
- Chart version bump to 3.3.0

### Follow-up issues

- #2387 — Support multiple storage endpoints per backend (values schema change)
- #2394 — Rework NetworkPolicy templates for CNI flexibility
- #2398 — Automate chart README generation via helm-docs

## ADR

## How to test

**Quick template validation:**
```bash
helm template test charts/sda-svc [your-values] \
  --set global.downloadV2.enabled=true \
  --set global.downloadV2.service.orgName=TestOrg \
  --set global.downloadV2.service.orgURL=http://test \
  --set global.ingress.hostName.downloadV2=dl-v2.example.com
```

**k3d smoke test (automated):**
```bash
./dev-tools/chart-smoke-test/test-download-v2.sh           # full run
./dev-tools/chart-smoke-test/test-download-v2.sh --no-build # skip image build
./dev-tools/chart-smoke-test/test-download-v2.sh --cleanup  # tear down
```

The smoke test first validates all 6 templates across 16 config combinations (TLS on/off × S3/POSIX × ingress × networkPolicy), then spins up k3d with postgres, rabbitmq, and minio, deploys the v2 download resources, and verifies `/health/ready` (DB + storage + gRPC ok), `/service-info`, and 401 on `/datasets` without a token.


charts/sda-svc/templates/download-v2-deploy.yaml
```diff
@@ -0,0 +1,152 @@
...
+          capabilities:
+            drop: ["ALL"]
+          runAsNonRoot: true
+          seccompProfile:
+            type: "RuntimeDefault"
+        command: ["sda-download"]
+        args: ["--config-file", "{{ include "confFile" . }}"]
+        ports:
+        - name: download-v2
+          containerPort: {{ ternary 8443 8080 .Values.global.tls.enabled }}
```
jbygdell: Given that we run on a non-privileged port and not run bot TLS and plain HTTP in parallel there is no need to have different ports.

```suggestion
          containerPort: 8080
```

charts/sda-svc/templates/download-v2-networkpolicy.yaml
```diff
@@ -0,0 +1,62 @@
...
+          podSelector:
+            matchLabels:
+              {{ .Values.global.networkPolicy.ingressLabelMatch | nindent 14 }}
+        ports:
+          - protocol: TCP
+            port: 8080
+          - protocol: TCP
+            port: 8443
+          - protocol: TCP
+            port: 443
```
jbygdell:
```suggestion
```

This only needs to match the open container ports

charts/sda-svc/templates/download-v2-secrets.yaml
```diff
@@ -0,0 +1,126 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ template "sda.fullname" . }}-download-v2
+type: Opaque
+stringData:
+  config.yaml: |-
+    api:
+      host: "0.0.0.0"
+      port: {{ ternary 8443 8080 .Values.global.tls.enabled }}
```
jbygdell:
```suggestion
      port: 8080
```

charts/sda-svc/templates/download-v2-secrets.yaml
```diff
@@ -0,0 +1,126 @@
...
+      password: {{ required "DB pass is required" (include "dbPassDownload" .) }}
+      database: {{ .Values.global.db.name }}
+      sslmode: {{ ternary .Values.global.db.sslMode "disable" .Values.global.tls.enabled }}
+      {{- if .Values.global.tls.enabled }}
+      cacert: {{ template "tlsPath" . }}/ca.crt
+      clientcert: {{ template "tlsPath" . }}/tls.crt
+      clientkey: {{ template "tlsPath" . }}/tls.key
+      {{- end }}
+    grpc:
+      host: {{ required "A valid grpc host is required" .Values.global.reencrypt.host }}
```
jbygdell: Technically this would only be needed if the reencrypt service is deployed in a different namespace. Otherwise the hostname will be defined by: `{{ template "sda.fullname" . }}-reencrypt`

charts/sda-svc/templates/download-v2-secrets.yaml
```diff
@@ -0,0 +1,126 @@
...
+          secret_key: {{ .Values.global.archive.s3SecretKey }}
+          bucket_prefix: {{ .Values.global.archive.s3BucketPrefix }}
+          region: {{ .Values.global.archive.s3Region }}
+          {{- if .Values.global.tls.enabled }}
+          ca_cert: {{ template "tlsPath" . }}/ca.crt
+          {{- end }}
+          {{- if .Values.global.archive.s3ChunkSize }}
+          chunk_size: {{ .Values.global.archive.s3ChunkSize }}
+          {{- end }}
+          disable_https: {{ not .Values.global.tls.enabled }}
```
jbygdell: This should support multiple archive URLs

charts/sda-svc/templates/download-v2-service.yaml
```diff
@@ -0,0 +1,18 @@
...
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ template "sda.fullname" . }}-download-v2
+  labels:
+    app: {{ template "sda.fullname" . }}-download-v2
+spec:
+  ports:
+  - name: download-v2
+    port: {{ ternary 443 80 .Values.global.tls.enabled }}
```
jbygdell: If this service is not intended to be accessed directly from the inside of the cluster this can be set to a static port number (ex. 8080) instead.

charts/sda-svc/templates/download-v2-ingress.yaml
```diff
@@ -0,0 +1,49 @@
...
+    {{- end }}
+spec:
+{{- if .Values.global.ingress.ingressClassName }}
+  ingressClassName: {{ .Values.global.ingress.ingressClassName }}
+{{- end }}
+  rules:
+  - host: {{ required "An ingress hostname is required for download-v2!" .Values.global.ingress.hostName.downloadV2 }}
+    http:
+      paths:
+      - pathType: Prefix
```
jbygdell, line 35: With a bit of clever path rewrite one could merge this into the existing download ingress.
But given that it requires controller specific settings it's better not to, for now.

jbygdell:
[chart version](https://github.com/neicnordic/sensitive-data-archive/blob/1950b2b4c7dd27186ea1870460b8641bc24527a6/charts/sda-svc/Chart.yaml#L3) should be given a minor bump. (3.3.0)

dev-tools/chart-smoke-test/test-download-v2.sh
```diff
@@ -0,0 +1,233 @@
...
+)
+
+# Render only v2 templates and apply directly
+for tmpl in download-v2-secrets download-v2-deploy download-v2-service; do
+    helm template pipeline charts/sda-svc "${HELM_ARGS[@]}" \
+        --show-only "templates/${tmpl}.yaml"
+done | kubectl apply -f -
+
+# Scale to 1 replica (single-node cluster can't satisfy topology spread for 2)
+kubectl scale deploy pipeline-sda-svc-download-v2 --replicas=1
```
jbygdell: Why not set this directly in the argument list? `--set downloadv2.replicaCount=1`

dev-tools/chart-smoke-test/test-download-v2.sh
```diff
@@ -0,0 +1,233 @@
...
+
+wait_for_pod "app=pipeline-sda-svc-download-v2" 60
+
+# -- 6. smoke tests --
+echo "=== Step 5: smoke tests ==="
+
+SVC="pipeline-sda-svc-download-v2:80"
```
jbygdell:
```suggestion
SVC="pipeline-sda-svc-download-v2"
```

Standard HTTP/S ports should not need to be set.

charts/sda-svc/templates/download-v2-secrets.yaml
```diff
@@ -0,0 +1,123 @@
...
+        {{- else }}
+        posix:
+        - path: {{ .Values.global.archive.volumePath }}
```
jbygdell, line 116: I now see that the `Values.yaml` doesn't support what the code requires.
In order to support multiple endpoints this block here should look more like this:
```suggestion
        archive:
        {{- with .Values.global.archive.posix }}
            posix:
        {{- toYaml . | nindent 8 }}
        {{- end }}
        {{- with .Values.global.archive.s3 }}
            s3:
        {{- toYaml . | nindent 8 }}
        {{- end }}
```

But then the `Values.yaml` need to look more like this:
```
global:
    archive:
#     s3:
#       -  endpoint: https://storage1.example.com
#           bucket_prefix: archive
#           region: us-east-1
#           chunk_size: 5mb  ## <-- only relevant for write operations
#           access_key: null
#           secret_key: null
#           port: 443               ## <-- should be optional
#           max_buckets: 10
#      -   endpoint: https://storage2.example.com
#           ...
#     posix:
#       - path: "/archive_1"
#         max_size: 123456     ## total bytes available, only for write operations
#         writer_disabled: bool  ## optional
#       - path: "/archive_2"

File: charts/sda-svc/templates/download-v2-deploy.yaml (hunk 1 above)
Comment (jhagberg) [reply charts/sda-svc/templates/download-v2-deploy.yaml]: Good point. Fixed in dc374da8 — the container now binds to a fixed 8080.

File: charts/sda-svc/templates/download-v2-networkpolicy.yaml (hunk 2 above)
Comment (jhagberg) [reply charts/sda-svc/templates/download-v2-networkpolicy.yaml]: Removed 8443 and 443 in dc374da8.

File: charts/sda-svc/templates/download-v2-secrets.yaml (hunk 3 above)
Comment (jhagberg) [reply charts/sda-svc/templates/download-v2-secrets.yaml]: Fixed in dc374da8. `api.port` is now hardcoded to 8080, and the `server-cert`/`server-key` block is gone — TLS terminates at the ingress.

File: charts/sda-svc/templates/download-v2-secrets.yaml (hunk 4 above)
Comment (jhagberg) [reply charts/sda-svc/templates/download-v2-secrets.yaml]: Applied in dc374da8. `grpc.host` now defaults to `{{ sda.fullname }}-reencrypt` and only needs an explicit value for split deployments.

File: charts/sda-svc/templates/download-v2-secrets.yaml (hunk 5 above)
Comment (jhagberg) [reply charts/sda-svc/templates/download-v2-secrets.yaml]: Agree. The attempt I first made (re-using `global.backupArchive`) was reverted in `7d9ee454` — backup is a separate storage namespace that download should never read from. See follow-up reply on line 110 for the proper path forward.

File: charts/sda-svc/templates/download-v2-service.yaml (hunk 6 above)
Comment (jhagberg) [reply charts/sda-svc/templates/download-v2-service.yaml]: Updated in 2b1eab5 to standard port 80 → `targetPort: 8080`, so in-cluster callers don't need to specify a non-standard port.

File: charts/sda-svc/templates/download-v2-ingress.yaml (hunk 7 above)
Comment (jhagberg) [reply charts/sda-svc/templates/download-v2-ingress.yaml:35]: Agreed, leaving as a separate ingress for now. If we converge on a single controller with path-rewrite support later, we can revisit.

File: dev-tools/chart-smoke-test/test-download-v2.sh (hunk 8 above)
Comment (jhagberg) [reply dev-tools/chart-smoke-test/test-download-v2.sh]: Good call. Switched to `--set downloadV2.replicaCount=1` in 2b1eab5b, dropped the separate `kubectl scale`.

File: dev-tools/chart-smoke-test/test-download-v2.sh (hunk 9 above)
Comment (jhagberg) [reply dev-tools/chart-smoke-test/test-download-v2.sh]: Applied in 2b1eab5 together with the service port change — the smoke test now uses `SVC="pipeline-sda-svc-download-v2"`.

File: charts/sda-svc/templates/download-v2-secrets.yaml (hunk 10 above)
Comment (jhagberg) [reply charts/sda-svc/templates/download-v2-secrets.yaml:116]: Thanks for the concrete proposal — that values shape is exactly where we should land. Two reasons I'd prefer to do it as a follow-up PR rather than in this one:

1. The same schema change needs to land in every archive-consuming chart (`finalize`, `verify`, `ingest`, `sync`…), not just download-v2. Doing it atomically across all services with one migration note makes the upgrade story much cleaner.
2. Landing this PR with the single-endpoint mapping unblocks the v2 rollout while we shape the schema proposal properly.

OK to merge this one with the single-endpoint mapping for now and open a follow-up issue for the values schema refactor?

**Comment by jhagberg** [comment]
Chart version bumped to `3.3.0` in `1fb7612e` per your comment.

File: charts/sda-svc/templates/download-v2-secrets.yaml (hunk 10 above)
Comment (jbygdell) [reply charts/sda-svc/templates/download-v2-secrets.yaml:116]: Yes, document and create a new issue.

Comment (jhagberg) [reply charts/sda-svc/templates/download-v2-secrets.yaml:116]: Opened #2387 with your proposed values shape as the starting point. Could you have a look and let me know if the scope, acceptance criteria, or proposed schema need adjustments before we pick it up?

**Review by jbygdell — DISMISSED** [review]

File: charts/sda-svc/templates/download-v2-deploy.yaml [hunk 11]
```diff
@@ -0,0 +1,152 @@
...
+          secret:
+            defaultMode: 0440
+            secretName: {{ template "sda.fullname" . }}-download-v2
+        {{- if .Values.global.downloadV2.visa.trustedIssuers }}
+        - name: iss
+          secret:
+            defaultMode: 0440
+            secretName: {{ template "sda.fullname" . }}-download-v2-iss
+        {{- end }}
+      {{- end }}
```
jbygdell, line 139:
```suggestion
        {{- end }}
        {{- if .Values.global.oidc.jwtSecret }}
        - name: jwt
          secret:
            defaultMode: 0440
            secretName: {{ .Values.global.oidc.jwtSecret  }}
        {{- end }}
      {{- end }}
```

charts/sda-svc/templates/download-v2-deploy.yaml
```diff
@@ -0,0 +1,152 @@
...
+{{ toYaml .Values.downloadV2.resources | trim | indent 10 }}
+        volumeMounts:
+      {{- if not .Values.global.vaultSecrets }}
+        - name: config
+          mountPath: {{ template "configPath" . }}
+        {{- if .Values.global.downloadV2.visa.trustedIssuers }}
+        - name: iss
+          mountPath: {{ template "trustedIssPath" . }}
+        {{- end }}
+      {{- end }}
```
jbygdell, line 99:
```suggestion
        {{- end }}
        {{- if .Values.global.oidc.jwtSecret }}
        - name: jwt
            mountPath: {{ template  jwtpath }}
        {{- end }}
      {{- end }}
```

charts/sda-svc/values.yaml
```diff
@@ -254,6 +256,102 @@ global:
...
+
+    app:
+      # @param global.downloadV2.app.environment set to "production" to enable production guards
+      environment: ""
+
+    jwt:
+      # @param global.downloadV2.jwt.pubkeyPath path to directory containing JWT public key PEM files
+      pubkeyPath: ""
+      # @param global.downloadV2.jwt.pubkeyURL URL to JWKS endpoint for JWT verification
+      pubkeyURL: ""
```
jbygdell, line 281: If this is for the user supplied access token why not reuse the OIDC definition.

charts/sda-svc/templates/download-v2-deploy.yaml
jhagberg: Applied in 978571da — when global.oidc.jwtSecret is set, the deployment now mounts that secret as a jwt volume.

charts/sda-svc/templates/download-v2-deploy.yaml
jhagberg: Applied in 978571da — corresponding volumeMount on jwtPath, gated on the same global.oidc.jwtSecret value.

charts/sda-svc/values.yaml
jhagberg: Good call. Reworked in 978571da: dropped the separate downloadV2.jwt.pubkeyPath knob, added global.oidc.jwtSecret as the canonical reference to the k8s secret holding the JWT public keys, and the deploy now mounts it on jwtPath. downloadV2.jwt.pubkeyURL is kept as the JWKS-endpoint alternative.

charts/sda-svc/templates/download-v2-secrets.yaml
```diff
@@ -0,0 +1,123 @@
...
+    auth:
+      allow-opaque: {{ .Values.global.downloadV2.auth.allowOpaque }}
+    {{- if .Values.global.downloadV2.visa.enabled }}
+    visa:
+      enabled: true
+      source: {{ .Values.global.downloadV2.visa.source }}
+      userinfo-url: {{ .Values.global.downloadV2.visa.userinfoURL }}
+      trusted-issuers-path: {{ template "trustedIssPath" . }}/iss.json
+      dataset-id-mode: {{ .Values.global.downloadV2.visa.datasetIdMode }}
+      identity:
```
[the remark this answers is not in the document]
jhagberg: Fixed in 9eeca4a — the secrets template now fails with a clear message when visa.enabled=true but trustedIssuers is empty, instead of letting the pod crashloop on a missing file.

charts/sda-svc/templates/download-v2-secrets.yaml
```diff
@@ -0,0 +1,123 @@
...
+    log:
+      level: {{ .Values.global.log.level }}
+    service:
+      id: {{ .Values.global.downloadV2.service.id }}
+      org-name: {{ .Values.global.downloadV2.service.orgName }}
+      org-url: {{ .Values.global.downloadV2.service.orgURL }}
+    jwt:
+      {{- if .Values.global.downloadV2.jwt.pubkeyPath }}
+      pubkey-path: {{ template "jwtPath" . }}/{{ .Values.global.downloadV2.jwt.pubkeyPath }}
+      {{- end }}
```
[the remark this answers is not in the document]
jhagberg: Resolved in 978571da via @jbygdell's review feedback: dropped the orphan global.downloadV2.jwt.pubkeyPath knob, added global.oidc.jwtSecret as the canonical k8s secret reference, and the deployment now mounts that secret at jwtPath when set.

dev-tools/chart-smoke-test/test-download-v2.sh
```diff
@@ -0,0 +1,232 @@
+#!/usr/bin/env bash
+# Local k3d smoke test for the v2 download service chart.
+#
+# Prerequisites: docker, k3d, kubectl, helm
+#
+# Usage:
+#   ./dev-tools/k3d/test-download-v2.sh              # full run (build + deploy + test)
+#   ./dev-tools/k3d/test-download-v2.sh --no-build    # skip image build (reuse existing)
+#   ./dev-tools/k3d/test-download-v2.sh --cleanup      # tear down cluster and exit
```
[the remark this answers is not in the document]
jhagberg: Fixed in 9eeca4a — the header usage block now references dev-tools/chart-smoke-test/ correctly. Stale paths from before the rename.

dev-tools/chart-smoke-test/test-download-v2.sh
```diff
@@ -0,0 +1,232 @@
...
+    matchLabels:
+      app: minio
+  template:
+    metadata:
+      labels:
+        app: minio
+    spec:
+      containers:
+      - name: minio
+        image: minio/minio:latest
```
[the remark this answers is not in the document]
jhagberg: Leaving minio:latest for now — this script is local-developer-only (not run in CI) so breakage is immediately visible, and the smoke test verifies functionality not version-specific behavior. Happy to pin if we ever wire this into CI.

charts/sda-svc/templates/download-v2-deploy.yaml
```diff
@@ -0,0 +1,152 @@
...
+          initialDelaySeconds: {{ .Values.downloadV2.livenessProbe.initialDelaySeconds | default 10 }}
+          periodSeconds: {{ .Values.downloadV2.livenessProbe.periodSeconds | default 30 }}
+          timeoutSeconds: {{ .Values.downloadV2.livenessProbe.timeoutSeconds | default 5 }}
+        readinessProbe:
+          httpGet:
+            path: /health/ready
+            port: download-v2
+            scheme: {{ ternary "HTTPS" "HTTP" .Values.global.tls.enabled }}
+          periodSeconds: {{ .Values.downloadV2.readinessProbe.periodSeconds | default 10 }}
+          timeoutSeconds: {{ .Values.downloadV2.readinessProbe.timeoutSeconds | default 9 }}
```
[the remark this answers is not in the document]
jhagberg: Good catch — fixed in 9eeca4a. Probe scheme is now hardcoded to HTTP since the API binary always serves plain HTTP on 8080 (TLS terminates at the ingress). Was a leftover ternary from before we removed api.server-cert/server-key.

charts/sda-svc/templates/download-v2-networkpolicy.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - namespaceSelector:
+        matchLabels:
+          name: {{ .Values.global.networkPolicy.internalNamespace }}
+    {{- end }}
+    - podSelector:
+        matchLabels:
+          role: reencrypt
+    ports:
+    - protocol: TCP
+      port: {{ .Values.global.reencrypt.port | int }}
```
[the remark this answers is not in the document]
jhagberg: Fixed in 9eeca4a — the NetworkPolicy egress port now uses the same ternary as the grpc.port in the secrets template, so TLS deployments can reach reencrypt on 50443.

charts/sda-svc/templates/download-v2-secrets.yaml
```diff
@@ -0,0 +1,123 @@
...
+    storage:
+      archive:
+        {{- if eq "s3" .Values.global.archive.storageType }}
+        {{- /* TODO: support multiple archive endpoints (v2 storage format is an array) */}}
+        s3:
+        - endpoint: {{ .Values.global.archive.s3Url }}:{{ .Values.global.archive.s3Port }}
+          access_key: {{ .Values.global.archive.s3AccessKey }}
+          secret_key: {{ .Values.global.archive.s3SecretKey }}
+          bucket_prefix: {{ .Values.global.archive.s3BucketPrefix }}
+          region: {{ .Values.global.archive.s3Region }}
```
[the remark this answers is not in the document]
jhagberg: Fixed in 9eeca4a — now matches the legacy download-secrets pattern: required guard on s3Url and the :port suffix is only appended when s3Port is set.

charts/sda-svc/templates/download-v2-secrets.yaml
```diff
@@ -0,0 +1,123 @@
...
+      cacert: {{ template "tlsPath" . }}/ca.crt
+      client-cert: {{ template "tlsPath" . }}/tls.crt
+      client-key: {{ template "tlsPath" . }}/tls.key
+      {{- end }}
+    log:
+      level: {{ .Values.global.log.level }}
+    service:
+      id: {{ .Values.global.downloadV2.service.id }}
+      org-name: {{ .Values.global.downloadV2.service.orgName }}
+      org-url: {{ .Values.global.downloadV2.service.orgURL }}
```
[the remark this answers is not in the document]
jhagberg: Fixed in 9eeca4a — applied your suggestion verbatim. Helm now fails fast on install/upgrade if either value is missing, instead of crashing the pod at startup.

charts/sda-svc/templates/download-v2-networkpolicy.yaml
```diff
@@ -0,0 +1,58 @@
...
+  egress:
+  - to:
+    {{- if .Values.global.networkPolicy.databaseNamespace }}
+    - namespaceSelector:
+        matchLabels:
+          name: {{ .Values.global.networkPolicy.databaseNamespace }}
+    {{- end }}
+    - podSelector:
+        matchLabels:
+          role: database
```
jbygdell, line 26: This value can not be hard coded.
Probably easiest to utilise the same method as below.

I would make use of a `toYaml` templating strategy so that all these external things can be defined manually.

charts/sda-svc/templates/download-v2-deploy.yaml
```diff
@@ -0,0 +1,162 @@
...
+          containerPort: 8080
+          protocol: TCP
+        livenessProbe:
+          httpGet:
+            path: /health/live
+            port: download-v2
+            scheme: HTTP
+          initialDelaySeconds: {{ .Values.downloadV2.livenessProbe.initialDelaySeconds | default 10 }}
+          periodSeconds: {{ .Values.downloadV2.livenessProbe.periodSeconds | default 30 }}
+          timeoutSeconds: {{ .Values.downloadV2.livenessProbe.timeoutSeconds | default 5 }}
```
jbygdell:
```suggestion
{{ toYaml .Values.downloadV2.livenessProbe | trim | indent 10 }}
```

Or remove all parts from the values file that are not templated here

charts/sda-svc/templates/download-v2-deploy.yaml
```diff
@@ -0,0 +1,162 @@
...
+        readinessProbe:
+          httpGet:
+            path: /health/ready
+            port: download-v2
+            scheme: HTTP
+          periodSeconds: {{ .Values.downloadV2.readinessProbe.periodSeconds | default 10 }}
+          timeoutSeconds: {{ .Values.downloadV2.readinessProbe.timeoutSeconds | default 9 }}
```
jbygdell:
```suggestion
{{ toYaml .Values.downloadV2.readinessProbe | trim | indent 10 }}
```

Or remove all parts from the values file that are not templated here

charts/sda-svc/values.yaml
```diff
@@ -525,6 +623,41 @@ download:
...
   minClientVersion: ""
 
+downloadV2:
+  name: download-v2
+  replicaCount: 2
+  livenessProbe:
+    httpGet:
+      path: /health/live
+      port: download-v2
+      # scheme set in deploy template via ternary on global.tls.enabled
```
jbygdell: These are not transferred to the deployment manifest.
either remove from here or update the deployment based on the comment in that file.

charts/sda-svc/values.yaml
```diff
@@ -525,6 +623,41 @@ download:
...
+    initialDelaySeconds: 10
+    periodSeconds: 30
+    timeoutSeconds: 5
+  readinessProbe:
+    httpGet:
+      path: /health/ready
+      port: download-v2
+      # scheme set in deploy template via ternary on global.tls.enabled
```
jbygdell: These are not transferred to the deployment manifest.
either remove from here or update the deployment based on the comment in that file.

charts/sda-svc/templates/download-v2-deploy.yaml
jhagberg: Done in 7bd19c96 — template now `toYaml`s the probe straight from values, same pattern as api/auth/download.

charts/sda-svc/templates/download-v2-deploy.yaml
jhagberg: Same for readiness, 7bd19c96.

charts/sda-svc/values.yaml
jhagberg: Fixed in 7bd19c96 — template just `toYaml`s the probe block so values maps 1:1 to what ends up on the pod.

charts/sda-svc/values.yaml
jhagberg: Same fix, see 7bd19c96.

charts/sda-svc/templates/download-v2-networkpolicy.yaml
jhagberg: Opened #2394 to track the NetworkPolicy refactor across all services (configurable selectors, CNI-specific caveats). Leaving v2 matching the legacy pattern in this PR.

charts/sda-svc/templates/download-v2-secrets.yaml
```diff
@@ -0,0 +1,129 @@
...
+        {{- end }}
+---
+{{- if .Values.global.downloadV2.visa.trustedIssuers }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ template "sda.fullname" . }}-download-v2-iss
+type: Opaque
+data:
+  iss.json: {{ .Values.global.downloadV2.visa.trustedIssuers | toJson | b64enc | quote }}
```
jbygdell, line 126:
```suggestion
  iss.json: {{ .Values.global.downloadV2.visa.trustedIssuers | toJson | quote }}
```

This would generate a file where the contents is base64 encoded rather than plain JSON.

dev-tools/chart-smoke-test/test-download-v2.sh
```diff
@@ -0,0 +1,319 @@
...
+check "/health/ready database ok" '"database":"ok"' "$health"
+check "/health/ready storage ok" '"storage":"ok"' "$health"
+
+svcinfo=$(run_curl curl-svcinfo curl -sf "http://$SVC/service-info")
+check "/service-info returns org" '"name":"TestOrg"' "$svcinfo"
+check "/service-info returns version" '"version":"2.0.0"' "$svcinfo"
+
+authresp=$(run_curl curl-auth curl -s -o /dev/null -w '%{http_code}' "http://$SVC/datasets")
+check "/datasets returns 401 without token" "401" "$authresp"
+
```
jbygdell, line 350: Include test for parsing config files.

ex. kubectl get secret NAME -o -o jsonpath='{.data.iss\.json}' | base64 -d | jq -r '.[0].iss' ==> compare against expected value.

Similarly for yaml files with the use of yq to extract a dynamic value like `global.downloadV2.service.orgName`

jhagberg: Done in 851d5962. Step 5a extracts config.yaml via yq (service.org-name, api.port, storage endpoint, db.host) and iss.json via base64+jq (iss, jku) before the HTTP tests. 14/14 passing locally.

Also swapped wait_for_pod for rollout status to avoid a race against the terminating pod between helm upgrades.

charts/sda-svc/values.yaml
```diff
@@ -38,6 +38,7 @@ global:
       auth: ""
       doa: ""
       download: ""
+      downloadV2: ""
```
KarlG-nbis, line 47: Ideally these new configurations should be documented in the sda-svc/README.md

charts/sda-svc/templates/download-v2-deploy.yaml
```diff
@@ -0,0 +1,151 @@
...
+        imagePullPolicy: {{ .Values.image.pullPolicy | quote }}
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop: ["ALL"]
+          runAsNonRoot: true
+          seccompProfile:
+            type: "RuntimeDefault"
+        command: ["sda-download"]
+        args: ["--config-file", "{{ include "confFile" . }}"]
```
KarlG-nbis: whats the benefit of passing `config-file` as an flag vs having it as the CONFIG_FILE env var similar as the other services?

charts/sda-svc/values.yaml
jhagberg: Agreed, follow-up in #2398 (helm-docs proposal).

charts/sda-svc/templates/download-v2-deploy.yaml
jhagberg: none really, both work via `viper.AutomaticEnv()` (config/v2/config.go:71-72). switched to `CONFIG_FILE` env in 6e12eef2.

charts/sda-svc/templates/download-v2-networkpolicy.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - from:
+        - namespaceSelector:
+            matchLabels:
+              name: {{ .Values.global.networkPolicy.ingressNamespace }}
+          podSelector:
+            matchLabels:
+              {{ .Values.global.networkPolicy.ingressLabelMatch | nindent 14 }}
+        ports:
+          - protocol: TCP
+            port: 8080
```
KarlG-nbis, line 53: Should this not be `80` to be aligned with the port the [ingress redirects to](https://github.com/neicnordic/sensitive-data-archive/pull/2384/changes#diff-9957054c40894f9c759ae5d9b859b70ff9f2d50ec47bffb678ca2f783b5452f4R41)? or does it not matter since 8080 is interpreted as 80?

charts/sda-svc/values.yaml
```diff
@@ -254,11 +256,107 @@ global:
...
+      # @param global.downloadV2.service.id GA4GH service-info identifier (reverse domain notation)
+      id: "neicnordic.sda.download"
+      # @param global.downloadV2.service.orgName organization name for GA4GH service-info
+      orgName: ""
+      # @param global.downloadV2.service.orgURL organization URL for GA4GH service-info
+      orgURL: ""
+
+    app:
+      # @param global.downloadV2.app.environment set to "production" to enable production guards
+      environment: ""
```
KarlG-nbis: Should the `environment` really be `downloadV2` specifc? imo makes sense if this was on the global scope(or perhaps even on the root scope) with "production" as default such that we in the future could just pass that to other apps as well if needed, since I do not see a use case with the need to configure the "environment" per application

charts/sda-svc/values.yaml
```diff
@@ -254,11 +256,107 @@ global:
       c4ghKeyFile: ""
       secretName: ""
 
+  downloadV2:
```
KarlG-nbis, line 267: These are currently under the `global` scope, but are application specific, imo that doesnt make much sense and should be moved to be [downloadV2](https://github.com/neicnordic/sensitive-data-archive/pull/2384/changes#diff-4194994f1d0dab730593906b694f9d002b0a8ad63273ed71356779280a9ec6e3R626) specific

charts/sda-svc/templates/download-v2-networkpolicy.yaml
jhagberg: 8080. NP filters at pod level so the service 80→8080 mapping isnt visible there, has to match what the container listens on.

charts/sda-svc/values.yaml
jhagberg: agreed, moved to `global.environment` in 9f5b8b62. v2 reads it as `app.environment` in the rendered config.

charts/sda-svc/values.yaml
jhagberg: the chart-wide convention is `global.X.*` for service config and top-level `X.*` for deployment shape (replicas, probes, resources). global.api, global.auth, global.download all follow it. moving v2 alone breaks the pattern, imo a chart-wide refactor should be its own change.

KarlG-nbis: My guess is the current structure with `global.X.*` for service config and top-level `X.*` for deployment shape was not the original intent, and imo it makes more sense for application level configurations (and deployment shape, etc) to be more related. And personally I dont see a big problem with breaking that pattern for the download v2 service as that is where we should be aiming. 

But yeah perhaps we could just have this as a follow up refactoring issue as well

jbygdell: Agreed, there are a number of things that we should change in the chart.
* better overall structure
* option to  include credentials from existing secrets
* option to use mTLS instead of basic auth

jhagberg: opened #2418 to track all three (structure, existingSecret, mTLS). lmk if scope or wording needs adjusting before someone picks it up.

KarlG-nbis (approved):

jbygdell (approved):

Issue #2365, opened 2026-03-27
fix(download): make reencrypt HealthCheck verify actual connectivity


jhagberg:
### Context

The reencrypt client's `HealthCheck()` method calls `connect()`, which uses
`grpc.NewClient` — but `grpc.NewClient` performs no I/O. This means the
health check only validates that the client can be constructed with valid
options, not that the remote reencrypt service is reachable.

The download `/health/ready` endpoint uses this method, so it can report
the gRPC connection as healthy when the reencrypt service is actually down.

Raised in PR #2187 review by @KarlG-nbis (`reencrypt/reencrypt.go:198`).
A TODO was added in commit 552e39dd.

### Proposed change

Make `HealthCheck` perform a real RPC to verify connectivity. Options:

1. **gRPC health protocol** — call the standard `grpc.health.v1.Health/Check`
   RPC if the reencrypt service implements it.
2. **Lightweight RPC** — send a no-op or echo request to the reencrypt service.
3. **Connection state check** — use `conn.GetState()` to check if the
   connection is in `READY` state, and trigger a connect if `IDLE`.

Option 1 is cleanest if the reencrypt service already serves gRPC health.

### Acceptance criteria

- [ ] `HealthCheck` verifies the reencrypt service is reachable
- [ ] `/health/ready` accurately reflects reencrypt availability
- [ ] Unit test covers the case where reencrypt is unreachable
- [ ] Remove the TODO comment from `reencrypt.go`

### Estimation of size

small

### Estimation of priority

medium — health readiness should accurately reflect dependency status

jhagberg:
closed by #2388

Issue #2364, opened 2026-03-27
feat(storage): add ping/healthcheck to storage/v2 Reader interface


jhagberg:
### Context

The download v2 service health readiness endpoint (`/health/ready`) checks
storage by verifying the reader is configured (non-nil), but cannot verify
actual connectivity to the storage backend.

Raised in PR #2187 review by @KarlG-nbis (`handlers/health.go:43`).

### Current state

`handlers/health.go` checks:
```go
if h.storageReader != nil {
    services["storage"] = "ok"
}
```

The `storage.Reader` interface (`internal/storage/v2/reader.go`) has no
ping or health method.

### Proposed change

Add a `Ping(ctx context.Context) error` method to the `storage.Reader`
interface. For the S3 backend, this could perform a `HeadBucket` call.
For POSIX, a stat on the root path.

The download health endpoint would then call `storageReader.Ping(ctx)` to
verify actual backend connectivity.

### Acceptance criteria

- [ ] `storage.Reader` interface gains a `Ping(ctx) error` method
- [ ] S3 reader implements Ping via HeadBucket or similar
- [ ] POSIX reader implements Ping via os.Stat on root path
- [ ] Download `/health/ready` uses Ping instead of nil-check
- [ ] Existing tests pass, new tests cover Ping

### Estimation of size

small

### Estimation of priority

low — nil-check is sufficient for most deployments; real health check is
a defence-in-depth improvement

jhagberg:
closed by #2388

PR #2343, opened 2026-03-23
refactor: health checks

zeidlitz:
## Related issue(s) and PR(s)
This PR addresses #2341 

## Description
We have reason to believe that the current CheckHealth handlers could leave idle database connections. This change aims at adding ideomatic kubernetes ready / liveness probes instead that make no new database connections removing the possibility of leaving idle database connections.

The endpoint /health is now replaced with /live and /ready meant to be [consumed by kubernetes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/)

PR #2388, opened 2026-04-14
fix: real health probes for v2 download service (#2364, #2365)

jhagberg:
## Summary

The `/health/ready` probe on the v2 download service reported "ok" for
storage and the reencrypt gRPC client based on nil-checks alone. Expired
S3 credentials or a dead reencrypt service would not trigger a
Kubernetes restart.

This PR closes both #2364 and #2365 by turning each dependency check
into a real round-trip.

- **Storage (#2364):** Added `Ping(ctx)` to the v2 `Reader` interface.
  S3 pings every configured endpoint with `ListBuckets` (matching the
  existing startup check). POSIX stats every path and re-verifies
  `IsDir`, mirroring the constructor's invariants. The wrapper
  delegates to both sub-readers and joins their errors.
- **Reencrypt (#2365):** Replaced the no-op `HealthCheck` (which only
  called `grpc.NewClient` and did no I/O) with a real
  `grpc_health_v1.Health/Check` RPC. The server side already registers
  the standard health service, so no server change is needed.
- **Context propagation:** `HealthCheck` now takes a `context.Context`
  and uses it directly without wrapping — the readiness handler's 5s
  deadline flows all the way through, matching how `db.Ping(ctx)` and
  `storageReader.Ping(ctx)` already work. This also follows the
  preference surfaced in #2343: let the caller (Kubernetes probe
  `timeoutSeconds`) bound health-check duration.

## Test plan

- [x] `go test -race ./internal/storage/v2/... ./cmd/download/...`
      passes (356 tests, 17 packages)
- [x] `go vet ./...` clean
- [x] New storage tests: S3 `Ping` OK + endpoint down, POSIX `Ping` OK
      + path gone + path is file
- [x] New reencrypt tests against a real `grpc_health_v1` server:
      SERVING, NOT_SERVING, server down
- [x] New handler tests using a real gRPC health server: all-healthy
      returns 200, reencrypt NOT_SERVING returns 503 with informative
      error
- [ ] Verify readiness probe behavior in a dev compose environment
- [ ] Confirm Kubernetes \`readinessProbe\` in the Helm chart still
      points to \`/health/ready\` after merge

### DIFF
--- a/.github/integration/scripts/make_download_credentials.sh
+++ b/.github/integration/scripts/make_download_credentials.sh
@@ -78,12 +78,16 @@ cat > "/shared/trusted-issuers.json" <<'EOF'
 ]
 EOF
 
-## create crypt4gh key
+## create crypt4gh keys
 if [ ! -f "/shared/c4gh.pub.pem" ]; then
     echo "downloading crypt4gh"
     latest_c4gh=$(curl --retry 100 -sL https://api.github.com/repos/neicnordic/crypt4gh/releases/latest | jq -r '.name')
     curl --retry 100 -sL "https://github.com/neicnordic/crypt4gh/releases/download/${latest_c4gh}/crypt4gh_linux_x86_64.tar.gz" | tar -xz -C /shared/
-    /shared/crypt4gh generate -n /shared/c4gh -p c4gh
+    # Generate all keys the shared reencrypt config references; passphrases
+    # must match sda/config.yaml.
+    /shared/crypt4gh generate -n /shared/c4gh -p c4ghpass
+    /shared/crypt4gh generate -n /shared/sync -p syncPass
+    /shared/crypt4gh generate -n /shared/rotatekey -p rotatekeyPass
 fi
 
 echo "download credentials setup complete"
--- a/sda/cmd/download/handlers/handlers_test.go
+++ b/sda/cmd/download/handlers/handlers_test.go
@@ -2,6 +2,8 @@ package handlers
 
 import (
 	"encoding/json"
+	"fmt"
+	"net"
 	"net/http"
 	"net/http/httptest"
 	"os"
@@ -10,9 +12,13 @@ import (
 
 	"github.com/gin-gonic/gin"
 	"github.com/neicnordic/sensitive-data-archive/cmd/download/middleware"
+	"github.com/neicnordic/sensitive-data-archive/cmd/download/reencrypt"
 	configv2 "github.com/neicnordic/sensitive-data-archive/internal/config/v2"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/health"
+	healthgrpc "google.golang.org/grpc/health/grpc_health_v1"
 )
 
 var loadConfigOnce sync.Once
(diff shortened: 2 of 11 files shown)


sda/cmd/download/handlers/health.go
```diff
@@ -40,15 +40,21 @@ func (h *Handlers) HealthReady(c *gin.Context) {
 
 	// Check storage reader
 	if h.storageReader != nil {
-		services["storage"] = "ok"
+		if err := h.storageReader.Ping(ctx); err != nil {
+			log.Warnf("health check: storage ping failed: %v", err)
+			services["storage"] = "error: " + err.Error()
```
jbygdell:
```suggestion
			services["storage"] = "error: backend unresponsive"
```

maybe not propagate the real error to the end users

sda/cmd/download/handlers/handlers_test.go
```diff
@@ -115,6 +121,135 @@ func TestHealthReady_Degraded(t *testing.T) {
...
+	assert.Equal(t, "degraded", response.Status)
+	assert.Equal(t, "ok", response.Services["database"])
+	assert.Equal(t, "ok", response.Services["storage"])
+	assert.Contains(t, response.Services["grpc"], "not configured")
+}
+
+func TestHealthReady_StorageUnhealthy(t *testing.T) {
+	router := gin.New()
+	mockDB := &mockDatabase{}
+	mockStorage := &mockStorageReader{pingErr: fmt.Errorf("connection refused")}
```
jbygdell:
```suggestion
	mockStorage := &mockStorageReader{pingErr: errors.New("connection refused")}
```

sda/cmd/download/handlers/handlers_test.go
```diff
@@ -2,6 +2,8 @@ package handlers
 
 import (
 	"encoding/json"
+	"fmt"
```
jbygdell:
```suggestion
	"errors"
	"fmt"
```

sda/cmd/download/handlers/health.go
jhagberg: Good point. The same pattern was in all three checks so I extended it to database and grpc as well. Real errors still end up in the warn log.

4bf7b201

sda/cmd/download/handlers/handlers_test.go
jhagberg: Applied in 4bf7b201.

sda/cmd/download/handlers/handlers_test.go
jhagberg: Done in 4bf7b201. fmt was only used for that one Errorf call, so I dropped it from the imports.

KarlG-nbis (approved):

jbygdell (approved):

.github/integration/scripts/make_download_credentials.sh
```diff
@@ -78,12 +78,16 @@ cat > "/shared/trusted-issuers.json" <<'EOF'
...
 
-## create crypt4gh key
+## create crypt4gh keys
 if [ ! -f "/shared/c4gh.pub.pem" ]; then
     echo "downloading crypt4gh"
     latest_c4gh=$(curl --retry 100 -sL https://api.github.com/repos/neicnordic/crypt4gh/releases/latest | jq -r '.name')
     curl --retry 100 -sL "https://github.com/neicnordic/crypt4gh/releases/download/${latest_c4gh}/crypt4gh_linux_x86_64.tar.gz" | tar -xz -C /shared/
-    /shared/crypt4gh generate -n /shared/c4gh -p c4gh
+    # Generate all keys the shared reencrypt config references; passphrases
+    # must match sda/config.yaml.
```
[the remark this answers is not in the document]
jhagberg: The "sda/config.yaml" reference here points to .github/integration/sda/config.yaml, which is the file the reencrypt container mounts in this compose setup. Not sda/config_local.yaml. Leaving the comment as is to avoid invalidating the existing approvals over wording.

sda/cmd/download/reencrypt/reencrypt.go
```diff
@@ -193,12 +194,27 @@ func (c *Client) ReencryptHeaderWithEditList(ctx context.Context, oldHeader []by
...
+	client := healthgrpc.NewHealthClient(c.conn)
+
+	resp, err := client.Check(ctx, &healthgrpc.HealthCheckRequest{})
+	if err != nil {
+		return fmt.Errorf("grpc health check failed: %w", err)
+	}
+
+	if resp.GetStatus() != healthgrpc.HealthCheckResponse_SERVING {
+		return fmt.Errorf("reencrypt service status: %s", resp.GetStatus().String())
+	}
```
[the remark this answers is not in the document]
jhagberg: Fair point. The reencrypt server currently sets the empty "" status and the Reencrypt_ServiceDesc.ServiceName status to SERVING in the same call, so an empty service field gives the same answer today. Worth tightening in a follow-up if per-service degradation is ever introduced. Leaving as is here to avoid invalidating the existing approvals.

sda/cmd/download/handlers/health.go
```diff
@@ -40,17 +40,23 @@ func (h *Handlers) HealthReady(c *gin.Context) {
...
 	}
 
 	// Check gRPC reencrypt client
 	if h.reencryptClient != nil {
-		if err := h.reencryptClient.HealthCheck(); err != nil {
+		if err := h.reencryptClient.HealthCheck(ctx); err != nil {
 			log.Warnf("health check: grpc connection failed: %v", err)
-			services["grpc"] = "error: " + err.Error()
+			services["grpc"] = "error: backend unresponsive"
 			allHealthy = false
```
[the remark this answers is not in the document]
jhagberg: This is intentional, per @jbygdell's review feedback in this PR (4bf7b201). The previous version exposed the raw backend error string which can leak infrastructure details. The detailed error is still logged at warn level for operators. PR description wording could be clearer about that, but the code is doing what we want.

sda/internal/storage/v2/posix/reader/reader.go
```diff
@@ -37,3 +38,18 @@ func NewReader(backendName string) (*Reader, error) {
...
 	return backend, nil
 }
+
+// Ping verifies all configured POSIX paths are accessible directories.
+func (r *Reader) Ping(_ context.Context) error {
+	for _, e := range r.configuredEndpoints {
+		fileInfo, err := os.Stat(e.Path)
+		if err != nil {
+			return fmt.Errorf("failed to ping POSIX path: %s, due to: %v", e.Path, err)
+		}
```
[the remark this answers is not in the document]
jhagberg: os.Stat is uninterruptible and the path list is small (one or two entries in practice), so adding ctx.Err() checks between iterations gives marginal value. Skipping to avoid invalidating the existing approvals.

Issue #1957, opened 2025-09-09
[database] Utilise prepared statements


KarlG-nbis:
Utilise [Prepared statements](https://go.dev/doc/database/prepared-statements) in the [database package](https://github.com/neicnordic/sensitive-data-archive/tree/main/sda/internal/database) where possible.

As prepared statements can help the code run a bit faster, reduces the repeated memory allocations needed when creating the query at each call and helps against SQL injection.

KarlG-nbis:
Will be implemented alongside #1956

KarlG-nbis:
Current WIP branch: https://github.com/neicnordic/sensitive-data-archive/tree/feature/database_interface

KarlG-nbis:
[PR](https://github.com/neicnordic/sensitive-data-archive/pull/2389) now ready for review

PR #2389, opened 2026-04-15
feature/database interface, prepared queries and exposed connection configuration

### LINKED ISSUE #1956: [database] interface for the database package
Create an interface for functions that the [database package](https://github.com/neicnordic/sensitive-data-archive/tree/main/sda/internal/database) supports and utilise interface instead of the [SDAdb](https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/internal/database/database.go#L28C6-L28C11)(which implements the interface) when interacting with the database.

Includes update all the function signatures to use the interface.

Benefits:
- Better clarity / readability of what db functions is supported
- Allows for easier mocking of interface in unit test (without the need for a postgresql instance)

KarlG-nbis:
## Related issue(s) and PR(s)
This PR closes #1956, #2366, and #1957 .

## Description

* Add database interface
  * Use dedicated database package struct for data transport (which should not be exposed directly by application to APIs, etc, just used for reading / writing to database layer)

* Remove database config from config package, as database config is now controlled by the database implementation packages and config/v2

* Add database transaction support

* Seperate db_functions.go into seperate files per method

* Update db functions to user pointers references to structs instead of copies of structs

* Utilise prepared statements, which are prepared during initilisation of database

* Move CheckKeyHash func from database to rotatekey, as database should not do application logic

* Update RegisterFile to not use function such that queries are controlled by application logic, and not dependent on whats in database

* Update MapFilesToDataset -> MapFileToDataset, so controll of each insert is done by application logic

* expose configuration for MaxIdleConns, MaxOpenConns, ConnMaxIdleTime, and ConnMaxLifetime

* Update naming of stableID -> accessionID in the go code

* Update api to use database interface
* Update auth to use database interface
* Update finalize to use database interface
* Update ingest to use database interface
* Update mapper to use database interface
   * Utilise transaction during message handling
* Update s3inbox to use database interface
* Update sync to use database interface
* Update verify to use database interface

## How to test
- Unit tests pass
- Integration test pass


sda/cmd/s3inbox/s3inbox_test.go
```diff
@@ -98,6 +103,14 @@ func TestMain(m *testing.M) {
...
 			fmt.Sprintf("%s/postgresql/initdb.d:/docker-entrypoint-initdb.d", rootDir),
 		},
+		PortBindings: map[docker.Port][]docker.PortBinding{
+			"5432/tcp": {
+				{
+					HostIP:   "0.0.0.0",
+					HostPort: "35431",
+				},
+			},
+		},
```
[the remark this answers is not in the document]
KarlG-nbis: This is needed since we want to simulate a network distubance by removing the container from the bridge network and for container to keep same port, will add comment stating reason.

sda/cmd/auth/main.go
```diff
@@ -265,7 +268,7 @@ func (auth AuthHandler) elixirLogin(ctx iris.Context) *OIDCData {
 
 		return nil
 	}
-	err = auth.Config.DB.UpdateUserInfo(idStruct.User, idStruct.Fullname, idStruct.Email, idStruct.EdupersonEntitlement)
+	err = auth.db.UpdateUserInfo(ctx, idStruct.User, idStruct.Fullname, idStruct.Email, idStruct.EdupersonEntitlement)
```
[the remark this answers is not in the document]
KarlG-nbis: It does compile and it does implement context and it essentially just forwards ctx.Request().Context()

sda/internal/database/postgres/method_map_file_to_dataset.go
```diff
@@ -0,0 +1,41 @@
...
+INSERT INTO sda.file_dataset (file_id, dataset_id)
+VALUES ($1, $2) ON CONFLICT DO NOTHING;
+`
+	queries[mapFileToDatasetInsertDatasetQuery] = `
+INSERT INTO sda.datasets (stable_id) 
+VALUES ($1) 
+ON CONFLICT (stable_id) DO 
+	UPDATE SET stable_id = EXCLUDED.stable_id
+RETURNING id;
+`
```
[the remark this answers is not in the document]
KarlG-nbis: That is preferable to having to do 2 queries insert + select

postgresql/initdb.d/04_grants.sql
```diff
@@ -111,6 +111,7 @@ CREATE ROLE mapper;
 GRANT USAGE ON SCHEMA sda TO mapper;
 GRANT INSERT ON sda.datasets TO mapper;
 GRANT SELECT ON sda.datasets TO mapper;
+GRANT UPDATE ON sda.datasets TO mapper;
```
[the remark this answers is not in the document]
KarlG-nbis: This is desired.

sda/cmd/sync/sync.go
```diff
@@ -274,7 +281,7 @@ func syncFiles(ctx context.Context, stableID string) error {
 	return nil
 }
 
-func buildSyncDatasetJSON(b []byte) ([]byte, error) {
+func buildSyncDatasetJSON(ctx context.Context, b []byte) ([]byte, error) {
 	var msg schema.DatasetMapping
 	_ = json.Unmarshal(b, &msg)
```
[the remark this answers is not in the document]
KarlG-nbis: Out of scope of PR

sda/internal/database/postgres/method_list_key_hashes.go
```diff
@@ -0,0 +1,49 @@
...
+
+	for rows.Next() {
+		h := &database.C4ghKeyHash{}
+		depr := sql.NullString{}
+		err := rows.Scan(&h.Hash, &h.Description, &h.CreatedAt, &depr)
+		if err != nil {
+			return nil, err
+		}
+		h.DeprecatedAt = depr.String
+
```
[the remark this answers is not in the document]
KarlG-nbis: N/A, this is the current implementation in main, which works

sda/internal/database/postgres/method_get_file_status.go
```diff
@@ -0,0 +1,29 @@
...
+
+const getFileStatusQuery = "getFileStatus"
+
+func init() {
+	queries[getFileStatusQuery] = `
+SELECT event 
+FROM sda.file_event_log 
+WHERE file_id = $1 
+ORDER BY id DESC LIMIT 1;
+`
```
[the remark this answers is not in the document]
KarlG-nbis: N/A, keeping existing behaviour

charts/sda-svc/templates/rotatekey-deploy.yaml
```diff
@@ -57,6 +57,8 @@ spec:
         env:
         - name: CONFIGFILE
           value: {{ include "confFile" . }}
+        - name: CONFIG_FILE
+          value: {{ include "confFile" . }}
```
jbygdell, line 61: Shouldn't this replace the ENV above?

KarlG-nbis: For now no, as the services are still using the "config" package as well as the "config/v2" package, and the "config" package expects CONFIGFILE, while "config/v2" expects CONFIG_FILE

So for now we need to set both, until we've refactored out the use of the "config" package.

sda/cmd/ingest/ingest_test.go
```diff
@@ -236,7 +250,7 @@ func (ts *TestSuite) SetupSuite() {
 		ts.FailNow("no private keys configured")
 	}
 
-	if err := ts.ingest.DB.AddKeyHash(hex.EncodeToString(publicKey[:]), "the test key"); err != nil {
+	if err := ts.ingest.db.AddKeyHash(context.TODO(), hex.EncodeToString(publicKey[:]), "the test key"); err != nil {
```
zeidlitz: does context.Background() signal intent clearer here? When I read contex.TODO() It tells me that there will be a context that makes sense here later that we have not figured out yet? Maybe that is the case and I have missed something.

sda/internal/config/config.go
```diff
@@ -487,22 +440,8 @@ func NewConfig(app string) (*Config, error) {
...
-			return nil, err
-		}
-		err = c.configDatabase()
-		if err != nil {
-			return nil, err
-		}
 
-		c.configSchemas()
-	case "intercept":
+	case "finalize", "ingest", "mapper", "verify", "intercept":
```
zeidlitz, line 444: So previously on the "intercept" case it just ran c.configBroker(), now it also runs the c.configSchemas(), is this just reduce the number of cases and that the "intercept" case can run c.configSchemas() but don't need to? Or was this a missing call before?

sda/internal/database/postgres/postgres.go
```diff
@@ -0,0 +1,329 @@
...
+	}, nil
+}
+
+func (db *pgDb) getPreparedStmt(tx *sql.Tx, queryName string) *sql.Stmt {
+	if tx == nil {
+		return db.preparedStatements[queryName]
+	}
+
+	return tx.Stmt(db.preparedStatements[queryName])
+}
```
zeidlitz: is it possible for getPreparedStmt to silently fail? Say that queryName does not exists in the map, will it return nil and could cause a caller to panic on the nill?

sda/internal/database/postgres/postgres.go
```diff
@@ -0,0 +1,329 @@
...
+	return db.db.PingContext(ctx)
+}
+
+// Close terminates the connection to the database
+func (db *pgDb) Close() error {
+	if db.db == nil {
+		return nil
+	}
+	for queryName, stmt := range db.preparedStatements {
+		if err := stmt.Close(); err != nil {
```
zeidlitz: is it a problem that this err is never surfaced if Close() returns an err?

sda/cmd/mapper/mapper.go
```diff
@@ -144,31 +150,52 @@ func handleMessage(ctx context.Context, delivered amqp.Delivery) {
...
 				continue
 			}
 
-			if fileMappingData == nil || fileMappingData.SubmissionLocation == "" {
-				log.Errorf("failed to find submission location for file with stable ID: %s, can not remove file from inbox", aID)
+			if fileMappingData == nil {
+				log.Errorf("could not find file with accession-id: %s, can not map file to dataset: %s", aID, mappings.DatasetID)
+
+				continue
+			}
```
jhagberg, line 195: I noticed that in the `mapping` case (mapper.go:175-184), if `GetMappingData` returns an error or `nil` for one of the accession IDs, we `continue` to the next ID, but the transaction is still committed and the message Ack'd at the end of the case (line 249, 259).

For a payload like `{accession_ids: [ID_1, ID_2, ID_3]}` where `ID_2` cannot be resolved, would not this mean the dataset still gets `registered` with only `ID_1` and `ID_3` mapped, and the message is permanently removed from the queue without any retry path?

The `MapFileToDataset` error branch a few lines below already does `return` + Nack on failure. Should we follow that same pattern for the `GetMappingData` error / nil cases as well, instead of `continue`?

zeidlitz (commented):
a few minor questions just out of curiosity, otherwise I can't see a reason not to add this. Did a few tests and runs locally and seems like there is no major behavior change that could cause issues from what I can see compared to main.

zeidlitz (dismissed):

sda/cmd/mapper/mapper.go
```diff
@@ -219,6 +246,16 @@ func handleMessage(ctx context.Context, delivered amqp.Delivery) {
 		return
 	}
 
+	if err := tx.Commit(); err != nil {
```
jhagberg, line 256: I noticed that `inboxWriter.RemoveFile` is called inside the per-accession-id loop (mapper.go:204), before `tx.Commit()` runs at the end of the case (line 249).

If the commit fails for some reason (transient DB error, deadlock, etc), the deferred `tx.Rollback()` will undo the mappings in `sda.file_dataset` and the dataset event log, but the inbox files have already been deleted from storage. Would not this leave us in a state where the database thinks nothing happened, while the files are gone from the inbox with no way to recover them?

Would it be safer to defer the inbox cleanup until after `tx.Commit()` has succeeded? For example, by collecting the paths to remove during the loop and only iterating over them once we know the transaction is committed?

sda/internal/database/postgres/method_register_file.go
```diff
@@ -0,0 +1,44 @@
...
+	if fileID != nil {
+		fileIDArg.Valid = true
+		fileIDArg.String = *fileID
+	}
+
+	if err := stmt.QueryRowContext(ctx, fileIDArg, inboxLocation, uploadPath, uploadUser).Scan(&createdFileID); err != nil {
+		return "", err
+	}
+
+	if _, err := logFileEventStmt.ExecContext(ctx, createdFileID, "registered", uploadUser, nil, nil); err != nil {
```
jhagberg, line 45: I noticed that the new `registerFile` (method_register_file.go:23-44) splits the file upsert and the `'registered'` event log insert into two separate `ExecContext` calls, and the public `RegisterFile` (`postgres.go:131`) invokes it with `tx=nil`. The previous implementation used the `sda.register_file` PL/pgSQL function (`postgresql/initdb.d/02_functions.sql`) which performed both operations atomically in a single stored procedure.

If the second `ExecContext` fails, or the connection drops between the two calls, would not this leave a row in `sda.files` without its corresponding `'registered'` entry in `sda.file_event_log`? A subsequent `RegisterFile` call from the same user/path would eventually self-heal via `ON CONFLICT ON CONSTRAINT unique_ingested DO UPDATE` and then insert the missing event, but until that happens the invariant is broken, and `GetUserFiles` can return the orphaned row via its `LEFT JOIN sda.file_event_log` and then fail scanning `NULL fel.event` into `SubmissionFileInfo.Status` (string).

Both public callers (`sda/cmd/ingest/ingest.go:336` and `sda/cmd/s3inbox/proxy.go:237`) currently go through the non-tx path.

Would it be safer to wrap only the non-transactional public `RegisterFile` path in an internal transaction, while leaving `pgTx.RegisterFile` to use the caller's existing tx? That keeps the original atomicity guarantee without re-introducing the stored procedure.

sda/internal/database/postgres/method_cancel_file.go
```diff
@@ -0,0 +1,46 @@
...
+		return fmt.Errorf("failed to unset file data (file-id: %s): %v", fileID, err)
+	}
+
+	if _, err := deleteChecksumsStmt.ExecContext(ctx, fileID); err != nil {
+		return fmt.Errorf("failed to delete checksums (file-id: %s): %v", fileID, err)
+	}
+
+	if _, err := logFileEventStmt.ExecContext(ctx, fileID, "disabled", "system", "{\"reason\": \"file cancelled\"}", message); err != nil {
+		return fmt.Errorf("failed to log cancel file event (file-id: %s): %v", fileID, err)
+	}
```
jhagberg, line 52: I noticed that `cancelFile` (method_cancel_file.go:26-46) executes three sequential `ExecContext` calls — clearing the archive fields on `sda.files`, deleting the checksums, and inserting the `'disabled'` file event — and the public `CancelFile` (`postgres.go:163`) invokes it with `tx=nil`. The previous implementation in `db_functions.go` wrapped all three mutations in a `BeginTx`/`Commit` block, making the operation atomic.

If the second `ExecContext` (delete checksums) or the third (`'disabled'` event log insert) fails, would not this leave the file in a partial state — archive metadata cleared on `sda.files`, but with leftover rows in `sda.checksums` and/or no `'disabled'` entry in `sda.file_event_log`?

The single caller (`sda/cmd/ingest/ingest.go:299`) runs after any applicable archive and backup removal has already succeeded (lines 283-297), so a partial cancellation could leave DB state that no longer matches storage. If the archive fields are cleared but a later statement fails, the handler returns `nack`, but a retry may stop after `GetArchived` returns nil (lines 270-280) because the file no longer appears archived, leaving leftover checksum rows and/or no `'disabled'` audit event.

Following the same pattern proposed for `RegisterFile`, would it be safer to wrap only the non-transactional public `CancelFile` path in an internal transaction, while leaving `pgTx.CancelFile` to use the caller's existing tx?

postgresql/initdb.d/04_grants.sql
jhagberg, line 114: I noticed that this PR adds `GRANT UPDATE ON sda.datasets TO mapper;` to `postgresql/initdb.d/04_grants.sql:114`, to support the new `mapFileToDatasetInsertDataset` query in `method_map_file_to_dataset.go:21-25` which uses `ON CONFLICT (stable_id) DO UPDATE SET stable_id = EXCLUDED.stable_id RETURNING id;`.

The `initdb.d/` scripts only run on freshly initialized databases. Running `git diff main...HEAD --stat -- postgresql/` shows that no corresponding migration was added under `postgresql/migratedb.d/` for this grant (last existing migration is `23_expand_files_table_with_storage_locations.sql`).

Would not this mean that on existing installations the `mapper` role keeps its previous permissions (no `UPDATE` on `sda.datasets`), and the first mapping message after deployment would fail with a `permission denied for table datasets` error from PostgreSQL?

Should we add a new migration step, e.g. `postgresql/migratedb.d/24_grant_update_on_datasets_to_mapper.sql`, that grants `UPDATE` on `sda.datasets` to `mapper` for upgraded databases, and add the corresponding schema version 24 entry to `initdb.d/01_main.sql`?

sda/cmd/api/api.go
```diff
@@ -1165,7 +1242,17 @@ func reVerify(c *gin.Context, accessionID string) (*gin.Context, error) {
...
+	reVerify := schema.IngestionVerification{
+		User:        reverificationData.SubmissionUser,
+		FilePath:    reverificationData.SubmissionFilePath,
+		FileID:      reverificationData.FileID,
+		ArchivePath: reverificationData.ArchiveFilePath,
+		EncryptedChecksums: []schema.Checksums{{
+			Type:  reverificationData.ArchivedCheckSumType,
+			Value: reverificationData.ArchivedCheckSum,
+		}},
+		ReVerify: false,
```
jhagberg: I noticed that this PR moves the `IngestionVerification` message construction from `GetReVerificationData` (`db_functions.go:1421` on main, which set `ReVerify: true`) to inline code in `reVerify` (api.go:1245-1255), where the `ReVerify` field is now hardcoded to `false`.

Looking at how the verify service consumes this message, `verify.go:364-365` switches on `message.ReVerify`: the `case message.ReVerify:` branch only does a checksum comparison and Ack/Nack (lines 365-409), while the `default:` branch (line 410+) treats the message as an initial verification: it may call `SetVerified` if the stored checksum differs (lines 468-476), always appends a new `'verified'` file event (line 481), and emits a fresh accession request to the verified queue (line 491).

Would not this mean that `PUT /file/verify/:accession` and `PUT /dataset/verify/*dataset` now hit the initial-verification path instead of the re-verify path, rewriting verification state and emitting a fresh accession request for files that already have an accession ID?

Should `ReVerify` be set to `true` here, the way `GetReVerificationData` did before, to keep the existing endpoint semantics?

sda/internal/config/config.go
KarlG-nbis: Yeah its just to minimize the amount of duplicate switch cases, as after the removal of `configDatabase` they were identical ( [intercept](https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/internal/config/config.go#L511) did to c.configSchemas() as well prior)

sda/internal/database/postgres/postgres.go
KarlG-nbis: In theory it could happen, but should never in practice if code is written correct, but yeah could add a check for that

sda/internal/database/postgres/postgres.go
KarlG-nbis: Not really as db should mainly be closed just before an application shutsdown, and there is not much for an application to react to if there is an error here, but will refactor it such that we will return such errors an let application react/log it instead

sda/cmd/ingest/ingest_test.go
KarlG-nbis: Ah yeah maybe context.Background() does make more sense, had mixed it up and though they recommended context.TODO() for tests, etc

sda/cmd/mapper/mapper.go
KarlG-nbis: Yeah updated to follow similar pattern, with nack and requeue if error, but just nack without requeue if not found

sda/cmd/mapper/mapper.go
KarlG-nbis: Yeah makes sense to act upon that after the tx commit, will update!

sda/cmd/api/api.go
KarlG-nbis: Yes good catch!

postgresql/initdb.d/04_grants.sql
KarlG-nbis: As per prior conversations with Jocke, role management should not be part of the schema, which was the reason to not introduce a new schema migration verison, and this grant has to happen manually as part of the deployment procedure

sda/internal/database/postgres/method_register_file.go
KarlG-nbis: Yeah I wanted to introduce tx handling in more services, but felt like that could cause this PR to grow out of scope, but will atleast add tx just around these kinds of calls to maintain current functionality

sda/internal/database/postgres/method_cancel_file.go
KarlG-nbis: Yeah updated to add tx just around `CancelFile` for now

sda/internal/database/postgres/method_get_key_hash.go
```diff
@@ -0,0 +1,30 @@
...
+func (db *pgDb) getKeyHash(ctx context.Context, tx *sql.Tx, fileID string) (string, error) {
+	stmt, err := db.getPreparedStmt(tx, getKeyHashQuery)
+	if err != nil {
+		return "", err
+	}
+
+	var keyHash string
+	if err := stmt.QueryRowContext(ctx, fileID).Scan(&keyHash); err != nil {
+		return "", err
+	}
```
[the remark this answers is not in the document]
KarlG-nbis: Out of scope, keeping current behaviour

sda/cmd/mapper/mapper.go
```diff
@@ -144,42 +150,70 @@ func handleMessage(ctx context.Context, delivered amqp.Delivery) {
...
+		}
+
+		return
+	}
+	defer func() {
+		if err := tx.Rollback(); err != nil {
+			log.Errorf("failed to rollback database transaction, due to: %v", err)
+		}
+	}()
+
```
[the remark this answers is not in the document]
KarlG-nbis: N/A, has been updated to commit before removal

charts/sda-svc/templates/rotatekey-secrets.yaml
```diff
@@ -33,17 +33,17 @@ stringData:
...
+      client_key: {{ template "tlsPath" . }}/tls.key
     {{- end }}
       host: {{ .Values.global.db.host }}
-      database: {{ .Values.global.db.name }}
+      name: {{ .Values.global.db.name }}
       password: {{ required "DB password is required" (include "dbPassRotate" .) }}
       port: {{ .Values.global.db.port }}
-      sslmode: {{ ternary .Values.global.db.sslMode "disable" .Values.global.tls.enabled }}
+      ssl_mode: {{ ternary .Values.global.db.sslMode "disable" .Values.global.tls.enabled }}
       user: {{ required "DB user is required" (include "dbUserRotate" .) }}
```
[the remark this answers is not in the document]
KarlG-nbis: fixed

sda/internal/storage/v2/s3/writer/writer_test.go
```diff
@@ -368,20 +368,207 @@ func (ts *WriterTestSuite) TestWriteFile_NoMockLocationBroker_FirstBucketFull()
 type notImplementedDatabase struct {
 }
 
-func (m *notImplementedDatabase) GetSizeAndObjectCountOfLocation(_ context.Context, _ string) (uint64, uint64, error) {
+func (m *notImplementedDatabase) BeginTransaction(_ context.Context) (database.Transaction, error) {
 	panic("function not expected to be called in unit tests")
 }
```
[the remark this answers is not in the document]
KarlG-nbis: Out of scope

sda/internal/storage/v2/posix/writer/writer_test.go
```diff
@@ -353,20 +353,207 @@ func (ts *WriterTestSuite) TestWriteFile_NoMockLocationBroker_FirstDirFull() {
 type notImplementedDatabase struct {
 }
 
-func (m *notImplementedDatabase) GetSizeAndObjectCountOfLocation(_ context.Context, _ string) (uint64, uint64, error) {
+func (m *notImplementedDatabase) BeginTransaction(_ context.Context) (database.Transaction, error) {
 	panic("function not expected to be called in unit tests")
 }
```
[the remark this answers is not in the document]
KarlG-nbis: Out of scope

postgresql/initdb.d/04_grants.sql
jhagberg: Aha makes sense! Sorry, missed the earlier copilot comment and your discussion with Jocke on this one!

sda/internal/database/postgres/config.go
```diff
@@ -0,0 +1,305 @@
...
+			Required: false,
+			AssignFunc: func(flagName string) {
+				globalConf.schema = viper.GetString(flagName)
+			},
+		},
+		&config.Flag{
+			Name: "database.max_idle_connections",
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.Int(flagName, globalConf.maxIdleConnections, "Sets the maximum number of connections in the idle connection pool, set to <= 0 for unlimited")
+			},
```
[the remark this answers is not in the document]
KarlG-nbis: fixed

charts/sda-svc/values.yaml
```diff
@@ -217,6 +217,14 @@ global:
     password: ""
     port: 5432
     sslMode: "verify-full"
+    ## @param maxIdleConnections, Sets the maximum number of connections in the idle connection pool, set to <= 0 for unlimited
```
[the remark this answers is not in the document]
KarlG-nbis: fixed

charts/sda-svc/README.md
```diff
@@ -145,6 +145,10 @@ The following table lists the configurable parameters of the `sda-svc` chart and
...
 | `global.db.passOutgest`                          | Password used for `data out` services.                                                                                                                                                                                                                                                         | `""`                                            |
 | `global.db.port`                                 | Port that the database is listening on.                                                                                                                                                                                                                                                        | `5432`                                          |
+| `global.db.maxIdleConnections`                   | Sets the maximum number of connections in the idle connection pool, set to <= 0 for unlimited                                                                                                                                                                                                  | `2`                                             |
```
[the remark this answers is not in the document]
KarlG-nbis: fixed

sda/cmd/ingest/ingest.go
```diff
@@ -325,17 +348,33 @@ func (app *Ingest) ingestFile(ctx context.Context, fileID string, message schema
...
+				log.Errorf("failed to rollback RegisterFile transaction, reason: %v", err)
+			}
+
+			return "nack"
+		}
+		if err := tx.Commit(); err != nil {
+			log.Errorf("failed to commit RegisterFile transaction, reason: %v", err)
 
 			return "nack"
 		}
```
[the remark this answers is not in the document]
KarlG-nbis: fixed

sda/cmd/ingest/ingest.go
```diff
@@ -289,24 +296,40 @@ func (app *Ingest) cancelFile(ctx context.Context, fileID string, message schema
...
+			log.Errorf("failed to rollback CancelFile transaction, reason: %v", err)
+		}
+
+		return "nack"
+	}
+	if err := tx.Commit(); err != nil {
+		log.Errorf("failed to commit CancelFile transaction, reason: %v", err)
+
 		return "nack"
 	}
```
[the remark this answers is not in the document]
KarlG-nbis: fixed

sda/cmd/s3inbox/proxy.go
```diff
@@ -233,10 +233,25 @@ func (p *Proxy) handleUpload(s3RequestType S3RequestType, w http.ResponseWriter,
...
+				log.Errorf("failed to rollback RegisterFile transaction, reason: %v", err)
+			}
+
+			return
+		}
+		if err := tx.Commit(); err != nil {
+			p.internalServerError(w, token.Subject(), r.Method, r.URL.Path, r.URL.RawQuery, fmt.Sprintf("failed to commit RegisterFile transaction, reason: %v", err))
 
 			return
 		}
```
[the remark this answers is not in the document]
KarlG-nbis: fixed

pahatz (dismissed):

sda/internal/database/postgres/config.go
```diff
@@ -0,0 +1,305 @@
...
+// Initialize globalConf with default values
+var globalConf = &dbConfig{
+	host:                  "", // No default, needs to be provided by config / NewPostgresSQLDatabase options
+	port:                  0,  // No default, needs to be provided by config / NewPostgresSQLDatabase options
+	user:                  "", // No default, needs to be provided by config / NewPostgresSQLDatabase options
+	password:              "", // No default, needs to be provided by config / NewPostgresSQLDatabase options
+	databaseName:          "sda",
+	schema:                "sda",
+	cACert:                "",
+	sslMode:               "disable",
```
nanjiangshu: It's a bit risky to define the default value of sslMode as `disable`. Better to set it as `verify-full` or `verify-ca`.

sda/internal/database/postgres/config.go
```diff
@@ -0,0 +1,305 @@
...
+		maxIdleConnections:    c.maxIdleConnections,
+		maxOpenConnections:    c.maxOpenConnections,
+		connectionMaxIdleTime: c.connectionMaxIdleTime,
+		connectionMaxLifeTime: c.connectionMaxLifeTime,
+	}
+}
+
+// dataSourceName builds a postgresql data source string to use with sql.Open().
+func (c *dbConfig) dataSourceName() string {
+	connInfo := fmt.Sprintf("host=%s port=%d user=%s password=%s dbname=%s sslmode=%s search_path=%s",
```
nanjiangshu: If a password contains special characters like `'`, the DSN may break or behave unexpectedly. it is better to escape special characters, or single quote the string.

sda/internal/database/postgres/config.go
```diff
@@ -0,0 +1,305 @@
...
+	if c.clientCert != "" {
+		connInfo = fmt.Sprintf("%s sslcert=%s", connInfo, c.clientCert)
+	}
+
+	if c.clientKey != "" {
+		connInfo = fmt.Sprintf("%s sslkey=%s", connInfo, c.clientKey)
+	}
+
+	return connInfo
+}
```
nanjiangshu, line 298: In addition, right now invalid combinations of SSL configuration are allowed. For example 
`client_cert without client_key` or `sslmode=verify-full without ca_cert`

Better to add a validation function e.g. 

```golang
func (c *dbConfig) validate() error
```
to fail the startup if the configuration is invalid.

sda/internal/database/postgres/config.go
```diff
@@ -0,0 +1,305 @@
...
+	user:                  "", // No default, needs to be provided by config / NewPostgresSQLDatabase options
+	password:              "", // No default, needs to be provided by config / NewPostgresSQLDatabase options
+	databaseName:          "sda",
+	schema:                "sda",
+	cACert:                "",
+	sslMode:               "disable",
+	clientCert:            "",
+	clientKey:             "",
+	maxIdleConnections:    2,
+	maxOpenConnections:    0,
```
nanjiangshu: This means unlimited open connections by default, shall we set a safer limit by default, e.g. 20?

sda/internal/database/postgres/config.go
nanjiangshu: I'm not quite sure where search_path will come from. If it is from user import, better to be sanitized?

nanjiangshu (commented):
The PR looks great and the refactoring is definitely an improvement compared to having all database methods in a single large file. 

However, with the current approach we now have quite a large number of `method_*.go` files in the same directory. I wonder if it could eventually become difficult to navigate as more DB methods are added.

Have you considered grouping related methods into domain-oriented files instead, for example:
- archive.go
- datasets.go
- headers.go
- users.go
- ...
so that closely related queries and methods stay together?

sda/internal/database/postgres/config.go
KarlG-nbis: Yeah makes sense, mainly kept it as `disable` as that was the current default, but will update

sda/internal/database/postgres/config.go
KarlG-nbis: Well it would fail the startup when application failes to connect to the database as well, do no think we should add to much costom validation, and instead rely on the connection to fail if configuration is incorrect

sda/internal/database/postgres/config.go
KarlG-nbis: Good point, will update!

sda/internal/database/postgres/config.go
KarlG-nbis: Yeah, will update to use `sql.OpenDB` with `pq.Config` instead to reduce risk of this

KarlG-nbis: it originates from the configuration with "sda" as default, what kind of sanitization where you thinking?

sda/internal/database/postgres/config.go
KarlG-nbis: Updated to 

maxIdleConnections:    2,
maxOpenConnections:    20,
connectionMaxIdleTime: 5 * time.Second,
connectionMaxLifeTime: 0,

which should be more sensible defaults for us

sda/internal/database/postgres/postgres.go
```diff
@@ -0,0 +1,344 @@
...
+	for queryName, query := range queries {
+		preparedStmt, err := pg.db.Prepare(query)
+		if err != nil {
+			log.Errorf("failed to prepare query: %s, due to: %v", queryName, err)
+			_ = pg.Close()
+
+			return nil, fmt.Errorf("failed to prepare query: %s, due to: %w", queryName, err)
+		}
+		pg.preparedStatements[queryName] = preparedStmt
+	}
```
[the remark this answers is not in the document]
KarlG-nbis: future problem, and we've discussed that user grants should not be controlled by the schema

sda/internal/database/postgres/config.go
```diff
@@ -0,0 +1,298 @@
...
+	password:              "", // No default, needs to be provided by config / NewPostgresSQLDatabase options
+	databaseName:          "sda",
+	schema:                "sda",
+	cACert:                "",
+	sslMode:               "verify-full",
+	clientCert:            "",
+	clientKey:             "",
+	maxIdleConnections:    2,
+	maxOpenConnections:    20,
+	connectionMaxIdleTime: 5 * time.Second,
```
jhagberg: I noticed that this commit sets `connectionMaxIdleTime` to `5 * time.Second` (config.go:44). The previous default was `0` (no idle timeout), and similar pools elsewhere tend to use much longer intervals — pgxpool defaults to 30 minutes, for example.

Was the intent to recover quickly from stale connections after transient DB hiccups, or to keep the pool small under low traffic? Just want to make sure we are not going to see noticeable connection churn on services with bursty request patterns.

jhagberg (dismissed):
Thanks for the quick turnaround on this! The fixes look good and the tx-wrapping around `RegisterFile` and `CancelFile` reads well. Approving from my side!

sda/internal/database/postgres/config.go
KarlG-nbis: ooh wops, meant to set it to 5 min actually! As on the server side we limit idle connections to 10min

sda/internal/database/postgres/config.go
nanjiangshu: Aha, if `search_path` is from the configuration, then it's safer. But still better to wrap the entire value in single quotes.

sda/internal/database/postgres/config.go
```diff
@@ -0,0 +1,298 @@
...
+		Host:        c.host,
+		Port:        c.port,
+		Database:    c.databaseName,
+		User:        c.user,
+		Password:    c.password,
+		SSLMode:     pq.SSLMode(c.sslMode),
+		SSLCert:     c.clientCert,
+		SSLKey:      c.clientKey,
+		SSLRootCert: c.cACert,
+		Options:     fmt.Sprintf("-c search_path=%s", c.schema),
```
nanjiangshu:
```suggestion
		Options:     fmt.Sprintf("-c search_path='%s'", c.schema),
```

While YAML is usually clean, single-quoting search_path is a defensive best practice. it ensures lib/pq treats the value as a literal string and prevents connection errors if spaces are introduced via environment overrides.

nanjiangshu (commented):
Great work. I don't have any more comments except for quoting of search_path string.

nanjiangshu (approved):

jhagberg (approved):

sda/internal/database/postgres/config.go
jbygdell: This shouldn't be required, if omitted the library defaults to the standard port.

jbygdell: Setting this to `verify-full` means that client certificates will be required. PG defaults to `Prefer`

KarlG-nbis: Yeah, updated this to `verify-full` based on [comment](https://github.com/neicnordic/sensitive-data-archive/pull/2389#discussion_r3219122898) from @nanjiangshu, but can update to `prefer` if you think that makes more sense for the sda stack

jbygdell: Using non TLS defaults simplifies testing since we don't have to generate certificates for the integration tests

sda/cmd/auth/main.go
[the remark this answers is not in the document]
KarlG-nbis: N/A, `iris.Context` should pass the `ctx.Request().Context()` underneath

sda/internal/database/postgres/config.go
KarlG-nbis: updated to use `prefer`

sda/cmd/auth/main.go
```diff
@@ -265,7 +268,7 @@ func (auth AuthHandler) elixirLogin(ctx iris.Context) *OIDCData {
...
 	if err != nil {
```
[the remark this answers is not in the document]
KarlG-nbis: N/A, iris.Context passes the ctx.Request().Context()

sda/cmd/auth/main.go
```diff
@@ -265,7 +268,7 @@ func (auth AuthHandler) elixirLogin(ctx iris.Context) *OIDCData {
...
 		log.Warn("Could not log user info.")
```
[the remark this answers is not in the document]
KarlG-nbis: N/A iris.Context passes ctx.Request().Context()

sda/internal/database/postgres/method_store_header.go
```diff
@@ -0,0 +1,35 @@
...
+		return err
+	}
+
+	result, err := stmt.ExecContext(ctx, hex.EncodeToString(header), id)
+	if err != nil {
+		return err
+	}
+	if rowsAffected, _ := result.RowsAffected(); rowsAffected == 0 {
+		return errors.New("something went wrong with the query zero rows were changed")
+	}
```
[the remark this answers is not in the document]
KarlG-nbis: N/A

sda/internal/database/postgres/method_set_accession_id.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	result, err := stmt.ExecContext(ctx, accessionID, fileID)
+	if err != nil {
+		return err
+	}
+
+	if rowsAffected, _ := result.RowsAffected(); rowsAffected == 0 {
+		return errors.New("something went wrong with the query zero rows were changed")
+	}
```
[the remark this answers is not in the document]
KarlG-nbis: N/A

sda/internal/database/postgres/method_set_key_hash.go
```diff
@@ -0,0 +1,33 @@
...
+		return err
+	}
+
+	result, err := stmt.ExecContext(ctx, keyHash, fileID)
+	if err != nil {
+		return err
+	}
+	if rowsAffected, _ := result.RowsAffected(); rowsAffected == 0 {
+		return errors.New("something went wrong with the query, zero rows were changed")
+	}
```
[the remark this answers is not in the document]
KarlG-nbis: N/A

PR #2417, opened 2026-04-30
fix(s3/writer): sort buckets numerically so 11th bucket is created

### LINKED ISSUE #2413: [storage-v2] s3 writer - only 10 buckets can automatically be created
# Describe the bug
In the storage-v2 when configured to support above 10 buckets, when creating the 11th bucket the 2nd bucket will be attempted to be created again, and code will return error saying bucket already exists.

This is due to how the bucket names are sorted, and its not using natural sort

In the [code](https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/internal/storage/v2/s3/writer/config.go#L216)
We use 

```go
slices.SortFunc(bucketsWithPrefix, strings.Compare)
```
which results in
```
[bucket-1 bucket-10 bucket-2 bucket-3 bucket-4 bucket-5 bucket-6 bucket-7 bucket-8 bucket-9]
```
when having 10 buckets, with "bucket-" as the prefix.

But
```go
[bucket-1 bucket-2 bucket-3 bucket-4 bucket-5 bucket-6 bucket-7 bucket-8 bucket-9 bucket-10]
```
 is expected

# Steps to reproduce
* Configure archive storage with 1 maxObjects, and 11 maxBuckets
* Ingest 11 files
* observe that ingest fails with BucketAlreadyExists error

# Expected behavior
- [ ] Expected bucket is created and used

# Additional context
Using a sort func something similar as the following should resolve the issue
```go
slices.SortFunc(test, func(a, b string) int {
	aInc, aerr := strconv.Atoi(strings.TrimPrefix(a, prefix))
	bInc, berr := strconv.Atoi(strings.TrimPrefix(b, prefix))
	if aerr != nil || berr != nil {
		return 0
	}
	return aInc - bInc
})
```

# Estimation of size
small/medium/big

# Estimation of priority
low/medium/high

SAY-5:
## Related issue(s) and PR(s)
This PR closes #2413.

## Description

`findActiveBucket` sorts the existing prefix-matched buckets with `strings.Compare` and then uses the last element as the previous increment. With ten buckets named `bucket-1` through `bucket-10` the lexical order yields:

```
[bucket-1 bucket-10 bucket-2 bucket-3 bucket-4 bucket-5 bucket-6 bucket-7 bucket-8 bucket-9]
```

so the "last" bucket becomes `bucket-9` and the next increment is `bucket-10`, which the S3 API rejects with `BucketAlreadyExists`. Ingesting the 11th file therefore fails when `max_buckets > 10` — the exact reproducer in #2413.

The fix is the suggestion from the issue body, lifted into a small named helper `sortBucketsNumerically(buckets, prefix)` that:

1. parses the integer suffix after `prefix` and orders by that value, so `bucket-10` lands after `bucket-9`;
2. falls back to lexical comparison for any pair where either side doesn't match `<prefix><integer>`, so the comparator stays total and Go's slice-sort contract isn't violated.

## ADR

## How to test

Three unit tests added under `internal/storage/v2/s3/writer`:

* `TestSortBucketsNumerically_OrderingFollowsSuffix` — ten-bucket regression case from the issue body; pins that the highest-numbered bucket is last so the next allocation increments it correctly.
* `TestSortBucketsNumerically_NumericFallback` — non-numeric suffix (e.g. `bucket-mixed`) doesn't get dropped or panic.
* `TestSortBucketsNumerically_EmptyAndSingle` — degenerate inputs (n=0, n=1).

```
$ go test -count=1 -run TestSortBucketsNumerically -v ./internal/storage/v2/s3/writer/...
=== RUN   TestSortBucketsNumerically_OrderingFollowsSuffix
--- PASS: TestSortBucketsNumerically_OrderingFollowsSuffix (0.00s)
=== RUN   TestSortBucketsNumerically_NumericFallback
--- PASS: TestSortBucketsNumerically_NumericFallback (0.00s)
=== RUN   TestSortBucketsNumerically_EmptyAndSingle
--- PASS: TestSortBucketsNumerically_EmptyAndSingle (0.00s)
PASS
ok  	github.com/neicnordic/sensitive-data-archive/internal/storage/v2/s3/writer	1.048s
```

`go build`, `go vet`, `gofmt` all clean on the modified files.

### DIFF
--- a/sda/internal/storage/v2/s3/writer/config.go
+++ b/sda/internal/storage/v2/s3/writer/config.go
@@ -213,7 +213,7 @@ func (endpointConf *endpointConfig) findActiveBucket(ctx context.Context, backen
 		return activeBucket, nil
 	}
 
-	slices.SortFunc(bucketsWithPrefix, strings.Compare)
+	sortBucketsNumerically(bucketsWithPrefix, endpointConf.BucketPrefix)
 
 	// find first bucket with available object count and size
 	for _, bucket := range bucketsWithPrefix {
@@ -253,4 +253,22 @@ func (endpointConf *endpointConfig) findActiveBucket(ctx context.Context, backen
 	}
 
 	return activeBucket, nil
+}
+
+// sortBucketsNumerically sorts buckets in place by the integer suffix that
+// follows prefix so `<prefix>10` comes after `<prefix>9`, not after
+// `<prefix>1`. With plain lexical ordering the 11th bucket lookup picked
+// `<prefix>9` as the "latest" entry and tried to recreate `<prefix>2`,
+// which the S3 API then rejected with BucketAlreadyExists (#2413). If a
+// bucket name can't be parsed as `<prefix><number>` we fall back to
+// lexical ordering between that pair so the function stays total.
+func sortBucketsNumerically(buckets []string, prefix string) {
+	slices.SortFunc(buckets, func(a, b string) int {
+		aInc, aErr := strconv.Atoi(strings.TrimPrefix(a, prefix))
+		bInc, bErr := strconv.Atoi(strings.TrimPrefix(b, prefix))
+		if aErr != nil || bErr != nil {
+			return strings.Compare(a, b)
+		}
+		return aInc - bInc
+	})
 }
--- a/sda/internal/storage/v2/s3/writer/config.go
+++ b/sda/internal/storage/v2/s3/writer/config.go
@@ -269,6 +269,7 @@ func sortBucketsNumerically(buckets []string, prefix string) {
 		if aErr != nil || bErr != nil {
 			return strings.Compare(a, b)
 		}
+
 		return aInc - bInc
 	})
 }
--- a/sda/internal/storage/v2/s3/writer/config_test.go
+++ b/sda/internal/storage/v2/s3/writer/config_test.go
@@ -1,0 +1,64 @@
+package writer
+
+import (
+	"testing"
+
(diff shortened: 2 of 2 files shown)


KarlG-nbis (dismissed):
Thanks for the contribution!

sda/internal/storage/v2/s3/writer/config.go
```diff
@@ -254,3 +254,21 @@ func (endpointConf *endpointConfig) findActiveBucket(ctx context.Context, backen
...
+// bucket name can't be parsed as `<prefix><number>` we fall back to
+// lexical ordering between that pair so the function stays total.
+func sortBucketsNumerically(buckets []string, prefix string) {
+	slices.SortFunc(buckets, func(a, b string) int {
+		aInc, aErr := strconv.Atoi(strings.TrimPrefix(a, prefix))
+		bInc, bErr := strconv.Atoi(strings.TrimPrefix(b, prefix))
+		if aErr != nil || bErr != nil {
+			return strings.Compare(a, b)
+		}
+		return aInc - bInc
```
jbygdell, line 273:
```suggestion

		return aInc - bInc
```

Missing blank line before return.

SAY-5:
Fixed nlreturn lint finding in config.go. Note: the gosec G124 in sda/cmd/auth/main.go is pre-existing on main and outside this diff.

KarlG-nbis (approved):

zeidlitz (approved):

Issue #1151, opened 2024-11-28
[sda-admin] extend delete feature


MalinAhlberg:
**Please describe the feature**
As an admin, I want to be able to delete files so that I don't have to do it manually

**Acceptance criteria**
- [ ] Delete files from the archive?
- [ ] Delete multiple files at once?
- [ ] Have a good strategy to keep us from removing the wrong files and being unable to restore them...

**Additional context**

**Estimation of size**: small/medium/big

**Estimation of priority**: low/medium/high

Created from discussions in https://scilifelab.atlassian.net/wiki/spaces/GS/pages/3294756906/Cleaning+the+inbox

jbygdell:
Maybe not link to our internal stuff in a public repo..

PR #2425, opened 2026-05-11
feat(api): extend admin api add file_events

zeidlitz:
## Related issue(s) and PR(s)
This PR closes issue #1151

## Description
Extends admin api with the endpoints `GET /file/events/:fileid` and `POST /file/events/:fileid/:event` which should address the following needs described in the issue:

- [x] List file events with sda-admin.
- [x] Create a new file-event log, with the admin-user as the triggering user (probably the token subject).
- [x] Add reason in the file event log as well so we know why this was done.
- [x] It should not be possible to delete events, just add new ones to take precedence.

## How to test

upload a file, call `GET /file/events/:fileid` to see the event associated with it. Call `POST /file/events/:fileid/:event` with a new event to change it.

### DIFF
--- a/sda-admin/main.go
+++ b/sda-admin/main.go
@@ -32,6 +32,10 @@ Commands:
                                 Assign accession ID to a file.
   file rotatekey -file-id FILEUUID
                                 Rotate encryption key for a specific file.
+  file get-event -file-id FILEUUID
+                                Retrieve the event / status for a given file
+  file update-event -file-id FILEUUID -event EVENT
+                                Update the event / status for a given file
   dataset create -user SUBMISSION_USER -dataset-id DATASET_ID accessionID [accessionID ...]
                                 Create a dataset from a list of accession IDs and a dataset ID.
   dataset release -dataset-id DATASET_ID
@@ -105,6 +109,23 @@ Options:
   -user USERNAME       Specify the username associated with the file.
   -fileid FILEUUID     Specify the file ID of the file to assign the accession ID.
   -accession-id ID     Specify the accession ID to assign to the file.`
+
+var getFileEventUsage = `Usage with file path and user: sda-admin file get-event -fileid <FILE_ID>
+Usage with file ID: sda-admin file get-event -fileid <FILE_ID>
+
+	Get the latest event associated with a file by providing the file id
+
+Options:
+  -fileid FILEUUID     Specify the file ID of the file to assign the accession ID.`
+
+var updateFileEventUsage = `Usage with fileid and event: sda-admin file update-event -fileid <FILE_ID> -event <EVENT>
+Usage with file ID: sda-admin file update-event -fileid <FILE_ID> -event <EVENT>
+
+	Append the file_event_log with a new event for the given fileid
+
+Options:
+  -fileid <FILE_ID>     Specify the file ID of the file to assign the accession ID.
+	-event <EVENT>        Specity the event to update the file with`
 
 var fileRotateKeyUsage = `Usage: sda-admin file rotatekey -file-id FILEUUID
   Rotate encryption key for a specific file.
@@ -282,6 +303,10 @@ func handleHelpFile() error {
 		_, _ = fmt.Println(fileAccessionUsage)
 	case flag.Arg(2) == "rotatekey":
 		_, _ = fmt.Println(fileRotateKeyUsage)
+	case flag.Arg(2) == "get-event":
+		_, _ = fmt.Println(getFileEventUsage)
+	case flag.Arg(2) == "update-event":
+		_, _ = fmt.Println(updateFileEventUsage)
 	default:
 		return fmt.Errorf("unknown subcommand '%s' for '%s'.\n%s", flag.Arg(2), flag.Arg(1), fileUsage)
 	}
@@ -367,6 +392,14 @@ func handleFileCommand() error {
 		}
 	case "rotatekey":
 		if err := handleFileRotateKeyCommand(); err != nil {
+			return err
+		}
+	case "get-event":
+		if err := handleFileGetEvent(); err != nil {
+			return err
+		}
+	case "update-event":
+		if err := handleFileUpdateEvent(); err != nil {
 			return err
 		}
 	default:
@@ -460,6 +493,46 @@ func handleFileRotateKeyCommand() error {
 	if err != nil {
 		return fmt.Errorf("error: failed to rotate key for file, reason: %v", err)
 	}
+
+	return nil
+}
+
+func handleFileGetEvent() error {
+	var fileID string
+	fileGetEventCmd := flag.NewFlagSet("get-event", flag.ExitOnError)
+	fileGetEventCmd.StringVar(&fileID, "file-id", "", "File ID (UUID) to rotate key for")
+
+	if err := fileGetEventCmd.Parse(flag.Args()[2:]); err != nil {
+		return fmt.Errorf("error: failed to parse command line arguments, reason: %v", err)
+	}
+
+	body, err := file.GetEvent(apiURI, token, fileID)
+	if err != nil {
+		return fmt.Errorf("could not get event for file %s, reason: %v", fileID, err)
+	}
+
+	fmt.Printf("response: %s", body)
+
+	return nil
+}
+
+func handleFileUpdateEvent() error {
+	var fileID, event, reason string
+	fileGetEventCmd := flag.NewFlagSet("update-event", flag.ExitOnError)
+	fileGetEventCmd.StringVar(&fileID, "file-id", "", "File ID (UUID) to append event to")
+	fileGetEventCmd.StringVar(&event, "event", "", "Event to append")
+	fileGetEventCmd.StringVar(&reason, "reason", "", "Reason for changing the event_log")
+
+	if err := fileGetEventCmd.Parse(flag.Args()[2:]); err != nil {
+		return fmt.Errorf("error: failed to parse command line arguments, reason: %v", err)
+	}
+
+	_, err := file.PostEvent(apiURI, token, fileID, event, reason)
+	if err != nil {
+		return fmt.Errorf("could not get event for file %s, reason: %v", fileID, err)
+	}
+
+	fmt.Printf("appended event '%s' for file '%s'", event, fileID)
 
 	return nil
 }
--- a/sda-admin/main.go
+++ b/sda-admin/main.go
@@ -32,7 +32,7 @@ Commands:
                                 Assign accession ID to a file.
   file rotatekey -file-id FILEUUID
                                 Rotate encryption key for a specific file.
-  file get-event -file-id FILEUUID
+  file get-events -file-id FILEUUID
                                 Retrieve the event / status for a given file
   file update-event -file-id FILEUUID -event EVENT
                                 Update the event / status for a given file
@@ -110,8 +110,8 @@ Options:
   -fileid FILEUUID     Specify the file ID of the file to assign the accession ID.
   -accession-id ID     Specify the accession ID to assign to the file.`
 
-var getFileEventUsage = `Usage with file path and user: sda-admin file get-event -fileid <FILE_ID>
-Usage with file ID: sda-admin file get-event -fileid <FILE_ID>
+var getFileEventUsage = `Usage with file path and user: sda-admin file get-events -fileid <FILE_ID>
+Usage with file ID: sda-admin file get-events -fileid <FILE_ID>
 
 	Get the latest event associated with a file by providing the file id
 
@@ -303,7 +303,7 @@ func handleHelpFile() error {
 		_, _ = fmt.Println(fileAccessionUsage)
 	case flag.Arg(2) == "rotatekey":
 		_, _ = fmt.Println(fileRotateKeyUsage)
-	case flag.Arg(2) == "get-event":
+	case flag.Arg(2) == "get-events":
 		_, _ = fmt.Println(getFileEventUsage)
 	case flag.Arg(2) == "update-event":
 		_, _ = fmt.Println(updateFileEventUsage)
@@ -394,7 +394,7 @@ func handleFileCommand() error {
 		if err := handleFileRotateKeyCommand(); err != nil {
 			return err
 		}
-	case "get-event":
+	case "get-events":
 		if err := handleFileGetEvent(); err != nil {
 			return err
 		}
@@ -499,7 +499,7 @@ func handleFileRotateKeyCommand() error {
 
 func handleFileGetEvent() error {
 	var fileID string
-	fileGetEventCmd := flag.NewFlagSet("get-event", flag.ExitOnError)
+	fileGetEventCmd := flag.NewFlagSet("get-events", flag.ExitOnError)
 	fileGetEventCmd.StringVar(&fileID, "file-id", "", "File ID (UUID) to rotate key for")
 
 	if err := fileGetEventCmd.Parse(flag.Args()[2:]); err != nil {
--- a/sda-admin/main.go
+++ b/sda-admin/main.go
@@ -529,7 +529,7 @@ func handleFileUpdateEvent() error {
 
 	_, err := file.PostEvent(apiURI, token, fileID, event, reason)
 	if err != nil {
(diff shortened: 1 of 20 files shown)


sda/cmd/api/api.go
```diff
@@ -202,6 +202,8 @@ func setup(conf *config.Config) (*http.Server, error) {
 	r.DELETE("/file/:username/:fileid", rbac(e), deleteFile)            // Delete a file from inbox
 	// submission endpoints below here
 	r.POST("/file/ingest", rbac(e), ingestFile)                      // start ingestion of a file
+	r.GET("/file/events/:fileid", rbac(e), getFileEvent)             // get file events associated with a file
+	r.POST("/file/events/:fileid/:event", rbac(e), updateFileEvent)  // get file events associated with a file
```
KarlG-nbis: These new APIs should be described in the cmd/apI/swagger.yaml, and ideally also in the cmd/apI/api.md

zeidlitz: Right! Added

KarlG-nbis: minor: comment `get file events associated with a file` does not align with what updateFileEvent does

sda/cmd/api/swagger_v1.yml
```diff
@@ -229,6 +229,51 @@ paths:
...
+    get:
+      description: Get events associated with a given fileID
+      parameters:
+      - in: path
+        name: fileID
+        schema:
+          type: string
+        required: true
+      responses:
+        "200":
```
KarlG-nbis: Do we not expect to get any data back on this API? ie is something like
```
        "200":
          content:
            application/json:
              schema:
                properties:
                  event:
                    type: string
                    example: uploaded
                  ...
```
missing?

sda/cmd/api/swagger_v1.yml
```diff
@@ -229,6 +229,51 @@ paths:
...
+        "401":
+          description: Authentication failure.
+        "404":
+          description: Bad user file ID combination
+        "500":
+          description: Internal application error.
+  /file/events/{fileID}/{event}:
+    post:
+      description: Append a event to the event log for the file associated with the given fileID
+      parameters:
```
KarlG-nbis, line 254: missing description of the request body that sets the file event log description, ie something like
``` 
    post:
      ...
      requestBody:
        content:
          application/json:
            properties:
              reason:
                    type: string
                    required: true
                    description: Describe the reason for updating the file event
                    example: "File event changed to X due to Y"
```

sda/cmd/api/api.go
```diff
@@ -1184,6 +1186,59 @@ func reVerify(c *gin.Context, accessionID string) (*gin.Context, error) {
...
+		c.AbortWithStatusJSON(
+			http.StatusBadRequest,
+			gin.H{
+				"error":  "json decoding : " + err.Error(),
+				"status": http.StatusBadRequest,
+			},
+		)
+	}
+
+	token, err := auth.Authenticate(c.Request)
```
KarlG-nbis: minor: auth should probably be done before ` c.BindJSON(&updateEvent)`

sda/cmd/api/api.go
```diff
@@ -1184,6 +1186,59 @@ func reVerify(c *gin.Context, accessionID string) (*gin.Context, error) {
...
+	if err != nil {
+		c.JSON(http.StatusUnauthorized, err.Error())
+
+		return
+	}
+
+	detailsMap := map[string]string{"reason": updateEvent.Reason}
```
KarlG-nbis: minor: would be good to validate that `updateEvent.Reason != ""`

KarlG-nbis: Also do we need to create this `detailsMap` would it be enough to just marshal updateEvent?

sda/cmd/api/api.go
```diff
@@ -1184,6 +1186,59 @@ func reVerify(c *gin.Context, accessionID string) (*gin.Context, error) {
...
+	}
+
+	err = Conf.API.DB.UpdateFileEventLog(fileID, event, token.Subject(), string(details), "{}")
+	if err != nil {
+		c.AbortWithStatusJSON(http.StatusInternalServerError, err.Error())
+
+		return
+	}
+
+	c.JSON(http.StatusOK, gin.H{"fileID": fileID, "event": event, "reason": string(details)})
```
KarlG-nbis: This response is not described in the swagger

KarlG-nbis: (if we event need it since this is an action, so 200 response should indicate that the requested action was done)

sda/cmd/api/swagger_v1.yml
```diff
@@ -229,6 +229,51 @@ paths:
           description: Authentication failure.
         "500":
           description: Internal application error.
+  /file/events/{fileID}:
+    get:
+      description: Get events associated with a given fileID
```
KarlG-nbis: Current behaviour of the API implementation is just returning the latest file event so description is abit misleading atm

sda/cmd/api/swagger_v1.yml
zeidlitz: done!

sda/cmd/api/api.go
zeidlitz: ah right, looks like a copy paste. Addressed

sda/cmd/api/swagger_v1.yml
KarlG-nbis: And according to #1151 we should have an API that "List file events with sda-admin." So to me makes sense to update the API impl, and swagger response to return all file events for a file id including the event, user_id, details, message, and timestamp

sda/cmd/api/api.go
zeidlitz: agreed, will adjust

sda/internal/schema/schema.go
```diff
@@ -153,6 +153,10 @@ type IngestionUserError struct {
 	Reason   string `json:"reason"`
 }
 
+type UpdateEvent struct {
```
KarlG-nbis: given this is just used to parse the request body of the `updateFileEvent` API impl dont think we need it here, could just be defined in the `updateFileEvent` impl, eg

``` 
func updateFileEvent(c *gin.Context) {
...
	type updateFileEventBody struct {
		Reason string `json:"reason"`	
	}
	updateEvent := &updateFileEventBody{}
...
```

sda/cmd/api/api.go
zeidlitz: good input, will adjust

sda/cmd/api/api.go
zeidlitz: sure, will adjust

sda/cmd/api/swagger_v1.yml
zeidlitz: agreed, will adjust

sda/internal/schema/schema.go
zeidlitz: sure! will adjust

sda/cmd/auth/main.go
```diff
@@ -222,7 +222,7 @@ func (auth AuthHandler) getEGAConf(ctx iris.Context) {
 // getOIDC redirects to the oidc page defined in auth.Config
 func (auth AuthHandler) getOIDC(ctx iris.Context) {
 	state := uuid.New()
-	ctx.SetCookie(&http.Cookie{Name: "state", Value: state.String(), Secure: true})
+	ctx.SetCookie(&http.Cookie{Name: "state", Value: state.String(), Secure: true, HttpOnly: true, SameSite: http.SameSiteLaxMode})
```
jbygdell, line 228: Not related to this PR

sda/cmd/api/api.go
```diff
@@ -1184,6 +1186,65 @@ func reVerify(c *gin.Context, accessionID string) (*gin.Context, error) {
 	return c, nil
 }
 
+func updateFileEvent(c *gin.Context) {
+	fileID := strings.TrimPrefix(c.Param("fileid"), "/")
+	event := strings.TrimPrefix(c.Param("event"), "/")
```
jbygdell, line 1338: event needs to be validated against the allowed states or do you rely on the DB error in case of a typo here?

sda/cmd/api/api.go
```diff
@@ -1184,6 +1186,65 @@ func reVerify(c *gin.Context, accessionID string) (*gin.Context, error) {
 	return c, nil
 }
 
+func updateFileEvent(c *gin.Context) {
```
jbygdell, line 1336: This function needs a test so that we verify correct response for all expected outcomes:
* All OK
* Bad file ID (typo)
* Bad event (typo)
* Missing message

jbygdell (dismissed):

sda/cmd/api/api.go
zeidlitz: Sure. Added.

sda/cmd/api/api.go
zeidlitz: My thinking was that since this is a admin endpoint it would make sense to get the postgres error returned but I can add validation of the events in `sda.file_events`

kostas-kou (commented):
Good job. I just have one comment, I think the issue mentions to be able to register an event by using the `sda-admin` tool [here](https://github.com/neicnordic/sensitive-data-archive/tree/main/sda-admin). Of course the changes in the api should be there but the tool needs to be extended as well. At least this is what I understand from the ticket. Just ignore this comment if I am wrong.

sda/cmd/api/swagger_v1.yml
```diff
@@ -229,6 +229,116 @@ paths:
...
+                error: "json decoding : unexpected end of JSON input"
+                status: 400
+        '401':
+          description: Unauthorized — missing or invalid Bearer token.
+          content:
+            application/json:
+              schema:
+                type: string
+              example: "token is expired"
+        '500':
```
jbygdell:
```suggestion
              example: "token is expired"
        '404':
          description: Bad request — file ID not found.
          content:
            application/json:
              schema:
                type: object
                properties:
                  error:
                    type: string
                  status:
                    type: integer
              example:
                error: ""file: e3b82a4a-b6da-4d7b-a795-27138e24f25a not found""
                status: 404
        '500':
```

All statuses we return should be in the swagger.

sda-admin/main.go
```diff
@@ -464,6 +497,46 @@ func handleFileRotateKeyCommand() error {
...
+	fileGetEventCmd.StringVar(&event, "event", "", "Event to append")
+	fileGetEventCmd.StringVar(&reason, "reason", "", "Reason for changing the event_log")
+
+	if err := fileGetEventCmd.Parse(flag.Args()[2:]); err != nil {
+		return fmt.Errorf("error: failed to parse command line arguments, reason: %v", err)
+	}
+
+	_, err := file.PostEvent(apiURI, token, fileID, event, reason)
+	if err != nil {
+		return fmt.Errorf("could not get event for file %s, reason: %v", fileID, err)
```
jbygdell:
```suggestion
		return fmt.Errorf("could not set event for file %s, reason: %v", fileID, err)
```

sda/cmd/api/api.go
```diff
@@ -1328,6 +1332,83 @@ func reVerify(c *gin.Context, accessionID string) (*gin.Context, error) {
...
+	fileID := strings.TrimPrefix(c.Param("fileid"), "/")
+	event := strings.TrimPrefix(c.Param("event"), "/")
+
+	token, err := auth.Authenticate(c.Request)
+	if err != nil {
+		c.JSON(http.StatusUnauthorized, err.Error())
+
+		return
+	}
```
jbygdell, line 1345:
```suggestion
```

The RBAC middleware should handle this part.

sda/internal/database/postgres/method_get_file_status_history.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/neicnordic/sensitive-data-archive/internal/database"
+)
+
+const getFileStatusHistoryQuery = "getFileStatus"
+
+func init() {
+	queries[getFileStatusHistoryQuery] = `
+SELECT event, user_id, details, message, started_at
+FROM sda.file_event_log 
+WHERE file_id = $1
```
jbygdell, line 16:
```suggestion
	queries[getFileStatusHistoryQuery] = `SELECT event, user_id, details, message, started_at
FROM sda.file_event_log 
WHERE file_id = $1 
```

Try not to have the the queries start with a `\n` since the logs becomes much harder to read when the first line is blanc.

sda/cmd/api/api.go
zeidlitz: How do we extract the subject without this? `UpdateFileEventLog` on [L1384](https://github.com/neicnordic/sensitive-data-archive/pull/2425/changes#diff-c32669eadfdf3640193ca09dc1398b703587f20a7f1fa58c969275581d80dfacR1384) needs it

sda/internal/database/postgres/method_get_file_status_history.go
zeidlitz: This is a non issue imo. All the other queries in the db function files have this formatting so I would just rather keep it aligned. This should have been raised on the db interface PR if It's issue worth blocking a change for.

sda-admin/main.go
zeidlitz: adjusted

pahatz (approved):
Well done!

kostas-kou (commented):
I know that this is becoming a bit too much and this issue should be split in two but I have a couple of more comments. Maybe some unit tests for the `sda-admin` part and instructions in the `sda-admin/README` how you update an event with the tool.
And maybe an integration test about the endpoint? Unless I missed it.

I am sorry for the second round of comments but most of them are about the `sda-admin` which was not there before

sda/cmd/api/api.go
```diff
@@ -1328,6 +1332,83 @@ func reVerify(c *gin.Context, accessionID string) (*gin.Context, error) {
...
+	fileID := strings.TrimPrefix(c.Param("fileid"), "/")
+
+	statusHistory, err := db.GetFileStatusHistory(c, fileID)
+	if err != nil {
+		c.AbortWithStatusJSON(http.StatusInternalServerError, err.Error())
+
+		return
+	}
+
+	c.JSON(http.StatusOK, statusHistory)
```
KarlG-nbis, line 1410: minor: would be good to have seperate struct (with json tags) for exposing on the api response, as the `models.FileStatus` belongs to the database layer, not api, etc (and currently it doesnt have json tags)

sda/internal/database/postgres/method_get_file_status_history.go
```diff
@@ -0,0 +1,45 @@
...
+`
+}
+
+func (db *pgDb) getFileStatusHistory(ctx context.Context, tx *sql.Tx, fileID string) ([]database.FileStatus, error) {
+	stmt, err := db.getPreparedStmt(tx, getFileStatusHistoryQuery)
+	if err != nil {
+		return nil, err
+	}
+
+	rows, err := stmt.QueryContext(ctx, fileID)
```
KarlG-nbis, line 26: missing closing of rows

sda/internal/database/postgres/method_get_file_events.go
```diff
@@ -0,0 +1,39 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	rows, err := stmt.QueryContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if rows.Err() != nil {
```
KarlG-nbis: `rows.Err()` should be checked after the `rows.Next()` loop, and to me a little cleaner with the following

```go
if err := rows.Err(); err != nil {
  return nil, err
}
``` 

(also missing closing of the rows)

```go
defer func() {
	_ = rows.Close()
}()
```

kostas-kou (approved):

sda/internal/database/postgres/method_get_file_status_history.go
zeidlitz: Good catch! Adjusted

sda/internal/database/postgres/method_get_file_events.go
zeidlitz: Adjusted!

PR #2427, opened 2026-05-19
feat(sda-svc)!: chart 4.0 multi-endpoint storage

### LINKED ISSUE #2387: feat(chart): support multiple storage endpoints per backend in sda-svc values

As a chart operator, I want to configure multiple S3 or POSIX endpoints per
storage backend (archive, inbox, sync, backup) in `global.*` values, so that
deployments with sharded, multi-region, or migrating storage can be expressed
without schema workarounds.

### Context

The storage-v2 package (introduced in #2244) accepts a list of endpoint
configs per backend:

```yaml
storage:
  archive:
    s3:
    - endpoint: https://storage1.example.com
      bucket_prefix: archive
      ...
    - endpoint: https://storage2.example.com
      ...
    posix:
    - path: /archive_1
    - path: /archive_2
```

The current `sda-svc` chart can only express one endpoint per backend, using
flat scalar keys:

```yaml
# Current values.yaml — flat scalars, single endpoint only
global:
  archive:
    storageType: s3
    s3Url: https://storage.example.com
    s3Port: 443
    s3AccessKey: ""
    s3SecretKey: ""
    s3BucketPrefix: archive
    s3Region: us-east-1
    s3ChunkSize: 5MB
    volumePath: /archive   # posix alternative
```

There is no way to add a second endpoint without repeating keys, which YAML
does not allow. The chart blocks operators from using the multi-endpoint
capability that the service code already supports.

@jbygdell raised this during review of #2384 (v2 download chart templates).

### Proposed values shape

Replace the flat per-backend keys with nested `s3:` and `posix:` lists, where
each list item carries the fields that used to be flat scalars:

```yaml
global:
  archive:
    s3:
    - endpoint: https://storage1.example.com
      bucket_prefix: archive
      region: us-east-1
      access_key: ""
      secret_key: ""
      chunk_size: 5MB
      max_buckets: 10
    - endpoint: https://storage2.example.com
      ...
    posix:
    - path: /archive_1
      max_size: 0
    - path: /archive_2
      writer_disabled: false
```

The same shape applies to `global.backupArchive`, `global.inbox`, and
`global.sync.destination`.

The shift is scalar → list of objects. The old keys (`s3Url`,
`s3BucketPrefix`, `s3Region`, `s3AccessKey`, `s3SecretKey`, `s3Port`,
`s3ChunkSize`, `volumePath`) go away. Their values move into list items under
`s3:` or `posix:`.

### Impact

All charts that emit storage config need updating:

- `download-v2-secrets.yaml` (added in #2384)
- `finalize-secrets.yaml`
- `verify-secrets.yaml`
- `ingest-secrets.yaml`
- `sync-secrets.yaml`
- Any others that reference `global.archive.*`, `global.backupArchive.*`,
  `global.inbox.*`, `global.sync.destination.*`

This is a breaking change to the chart values schema. Existing deployments
will need to migrate their `values.yaml`. We should add a migration note and
bump the chart to 4.0.0.

## Acceptance criteria

- [ ] `global.archive`, `global.backupArchive`, `global.inbox`, and
  `global.sync.destination` accept nested `s3: []` and `posix: []` lists
- [ ] Templates iterate over all configured endpoints and emit one config
  entry per list item
- [ ] Single-endpoint deployments remain expressible (list with one entry)
- [ ] Migration documented in `CHANGELOG.md` and chart README
- [ ] Chart version bumped to reflect breaking change
- [ ] Integration tests cover at least one multi-endpoint archive scenario

## Additional context

- Related PR: #2384 (adds v2 download chart with single-endpoint mapping as a
  temporary bridge)
- Related PR: #2244 (service-side multi-endpoint implementation)
- Related docs: `sda/internal/storage/v2/README.md`

## Estimation of size

medium

## Estimation of priority

medium. The service code already supports multi-endpoint, but nothing in
production is blocked today.

jhagberg:
## Related issue(s) and PR(s)
Closes #2387.

## Description

`charts/sda-svc` couldn't express multi-endpoint storage. The 3.x chart had flat scalar keys (`s3Url`, `storageType`, `volumePath`, ...) on each backend, but storage-v2 has expected nested `s3:`/`posix:` lists since PR #2244. This PR migrates `global.archive`, `global.backupArchive`, `global.inbox`, and `global.sync.destination` to the list shape.

Render-time fail-fasts so operators don't ship broken configs:

- A 3.x scalar key still set points at `MIGRATION-4.0.md`.
- Mixed `inbox.s3` + `inbox.posix` is rejected. The chart already deploys two separate inbox services, one for each protocol, so the combination has no clean meaning.
- Writer-enabled entries on both s3 and posix sides of a writer backend (archive, backupArchive, sync.destination) get blocked. storage-v2 returns `ErrorMultipleWritersNotSupported` at runtime; the chart names the values path and reminds you to set `writerDisabled: true` on the reader side.

DOA stays single-endpoint. Multi-endpoint archive + `doa.enabled` fails at render time. URL/port/secure get normalized for the Java MinIO client so a 3.x `archive.s3Url: minio.minio` (no scheme) still produces the right env vars.

Breaking change. `charts/sda-svc/MIGRATION-4.0.md` covers the scalar → list conversion, volume renames (`archive` → `archive-0`, `backup` → `backup-archive-0`, ...), and multi-endpoint examples.

## ADR

## How to test

```bash
helm lint charts/sda-svc
bash dev-tools/chart-smoke-test/test-helpers.sh          # 93 assertions
bash dev-tools/chart-smoke-test/test-multi-endpoint.sh   # 14 render scenarios
bash dev-tools/chart-smoke-test/test-download-v2.sh      # 16 template-matrix combos
```

`MIGRATION-4.0.md` has worked examples (two-S3, mixed s3+posix, two-NFS).


charts/sda-svc/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: sda-svc
-version: 3.4.1
+version: 4.0.0
 appVersion: v3.1.37
```
jbygdell:
```suggestion
appVersion: v3.1.69
```

charts/sda-svc/MIGRATION-4.0.md
```diff
@@ -0,0 +1,253 @@
...
+  archive:
+    locationBrokerCacheTTL: ""    # backend-level, unchanged
+    s3:
+    - endpoint: https://storage.example.com
+      port: 443
+      bucketPrefix: archive
+      region: us-east-1
+      accessKey: AKIA...
+      secretKey: "..."
+      chunkSize: 50MB
```
jbygdell, line 46: Chunk size should probably be defined on the top level  since it is not endpoint dependent.

jhagberg: I noticed that storage-v2 actually reads `chunk_size` per-endpoint, not at the backend level. It sits on the `endpointConfig` struct in both the reader (`sda/internal/storage/v2/s3/reader/config.go:26`) and the writer (`sda/internal/storage/v2/s3/writer/config.go:31`), and is validated per entry. So the per-endpoint shape in the chart mirrors what the service config expects.

That said, in practice the chunk size operators want is probably the same across endpoints. Would it be worth adding a backend-level convenience like `global.archive.chunkSize` that the helper applies to every endpoint unless an entry overrides it? Or is the 1:1 mirror fine as is, since `maxBuckets`, `maxObjects`, and `maxSize` are already shaped the same way?

jbygdell: The `maxBuckets` and sych are quotad per endpoint so having them separate for each one is a must.

The `Chunk size` should have application defaults and would be globally overridden based on performance testing. Since it is an optional configuration maybe we shouldn't spend too much time on it.

jhagberg: Makes sense, leaving it as is then. Thanks!

charts/sda-svc/CHANGELOG.md
```diff
@@ -0,0 +1,24 @@
+# sda-svc chart changelog
```
KarlG-nbis: Nice with a changelog! But as per agreement, we should follow https://keepachangelog.com/en/1.1.0/ 

so some small edits should be done to align with that.

eg: start file with

```
# Changelog - sda-svc Helm Chart

All notable changes to this project will be documented in this file.

The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
```

mark version entry with date
```
## [4.0.0] - 2023-05-27
```

etc,

but makes sense to keep the `Closes #2387.` in some format, and the other information as well

jhagberg: Thanks for pointing that out! Updated to follow keepachangelog.com with the preamble, `[version] - date` header, and Changed/Added/Removed sections.

kostas-kou (approved):

charts/sda-svc/CHANGELOG.md
jhagberg: Sorry I forgot to push the changes...

jbygdell (approved):

charts/sda-svc/templates/doa-deploy.yaml
```diff
@@ -163,36 +173,56 @@ spec:
...
-          value: {{ .Values.global.archive.s3Region }}
         - name: S3_SECURE
-          value: "true"
-        - name: S3_BUCKET
-          value: {{ .Values.global.archive.s3BucketPrefix | quote}}
-        {{- if .Values.global.archive.s3CaFile }}
+          value: {{ $doaSecure | quote }}
+        {{- if and (eq $doaSecure "true") .Values.global.tls.enabled }}
         - name: S3_ROOT_CERT_PATH
           value: "/etc/ssl/certs/java/ca.crt"
```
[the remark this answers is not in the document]
jbygdell: Ignore

charts/sda-svc/templates/_helpers.yaml
```diff
@@ -343,3 +343,260 @@ rotatekey credentials
...
+  {{- $writerBackends = append $writerBackends (dict "name" "sync.destination" "cfg" .Values.global.sync.destination) -}}
+{{- end -}}
+{{- range $writerBackends -}}
+  {{- $name := .name -}}{{- $cfg := .cfg -}}
+  {{- $s3w := 0 -}}{{- $posixW := 0 -}}
+  {{- range $cfg.s3 -}}{{- if not .writerDisabled -}}{{- $s3w = add $s3w 1 -}}{{- end -}}{{- end -}}
+  {{- range $cfg.posix -}}{{- if not .writerDisabled -}}{{- $posixW = add $posixW 1 -}}{{- end -}}{{- end -}}
+  {{- if and (gt $s3w 0) (gt $posixW 0) -}}
+    {{- fail (printf "global.%s has writer-enabled entries in both s3 and posix. storage-v2 supports only one writer implementation per backend (ErrorMultipleWritersNotSupported). Set writerDisabled: true on all but one writer; readers are unaffected." $name) -}}
+  {{- end -}}
```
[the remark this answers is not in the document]
jhagberg: Good catch, and this one turned out to be a real bug rather than a docs problem. I reproduced it with a test and fixed it in #2545, now merged as a26546ef. `loadConfig` in both `s3/writer/config.go` and `posix/writer/config.go` drops `writer_disabled` entries before the writer sees them, so the workflow the chart documents works at runtime.

The chart still pins `appVersion: v3.1.76`, which predates the fix, so the bump comes with the next sda release as usual.

PR #2441, opened 2026-05-26
[SDA-download] Improve calls of header reencrypt

pahatz:
Addresses failed downloads that were traced back to the rpc deadline exceeded. Includes a configuration change and some small improvements.

- increase reencrypt timeout
- stream now uses request context and exits on cancellation
- stream loop  guard against no-progress reads/writes by adding max consecutive empty-reads limit and writes returning zero bytes
- add stream flushes

## Related issue(s) and PR(s)

## Description
Addresses failed downloads that were traced back to the rpc deadline exceeded. Includes a configuration change and some small improvements.


sda-download/api/sda/sda.go
zeidlitz: nit: I think we can improve the comments here, something like:

```suggestion
    // EOF with no bytes read means we've consumed the stream cleanly.
    // If r > 0 we fall through to write the last chunk before exiting.
    if err == io.EOF && r == 0 {
        return nil
    }
```

sda-download/api/sda/sda.go
```diff
@@ -459,14 +459,22 @@ func adjustToStartPosition(fileStream io.Reader, start, end int64) (int64, int64
...
 	// Calculate how much we should read (if given)
 	togo := end - start
+	consecutiveEmptyReads := 0
+	const maxConsecutiveEmptyReads = 10
+
+	flusher, canFlush := writer.(http.Flusher)
 
 	buf := make([]byte, 4096)
 
 	// Loop until we've read what we should (if no/faulty end given, that's EOF)
```
zeidlitz: nit: I think we can improve this comment while we are here

maybe something like:

```suggestion
// Read until the target byte count is reached. If end is 0 (unknown/unset) read until EOF
```

zeidlitz (approved):
LGTM

KarlG-nbis (approved):

nanjiangshu (approved):
Looks good to me!

PR #2263, opened 2026-02-23
docs(rfc): RFC-0001 explore standardizing on accessionID everywhere

jhagberg:
## Related issue(s) and PR(s)

Originally opened as ADR-0001 following discussion in PR #2232 by @KarlG-nbis.
Converted to RFC-0001 per [ADR-0005](https://github.com/neicnordic/sensitive-data-archive/blob/main/docs/decisions/0005-introduce-rfcs-as-upstream-exploration-phase.md)
because the rollout questions haven't converged yet.

## Description

Explores standardising on `accessionID` / `accession_id` across the entire
stack — Go application code, API responses, and database schema.

The current codebase uses `stableID` in DB-layer Go code and `accessionID` in
API/message-facing code, while the DB columns use `stable_id`. This RFC
explores unifying on the bioinformatics domain term everywhere.

Reuses the MADR structure (now placed under `docs/rfcs/`): four considered
options, codebase survey (~232 `accessionID` vs ~187 `stableID` occurrences),
implementation guidance for the favoured direction, and a new **Open
Questions** section listing what hasn't been resolved.

**Direction currently favoured:** option 1, `accessionID` everywhere. Status
in the front matter is `exploring`.

## What changed in the ADR → RFC conversion

- Moved file to `docs/rfcs/0001-standardize-on-accessionid.md`
- `status: proposed` → `status: exploring`; `decision-makers:` → `authors:`
- Removed `## Decision Outcome` and `### Confirmation`
- Added `## Open Questions` (DB migration strategy, service sequencing,
  external consumers, reversibility window, migration script placement)
- Added `### Direction currently favoured` in More Information (preserves
  the earlier "Chosen option" rationale and the naming convention table)
- Added a row in `docs/rfcs/README.md` index

## How to test

- Review the RFC at `docs/rfcs/0001-standardize-on-accessionid.md`
- Verify the Open Questions reflect what is actually unresolved on the team
- Markdown renders correctly on GitHub
- Promoted to an ADR per ADR-0005's procedure once the Open Questions land

PR #2449, opened 2026-05-28
docs(rfc): RFC-0004 Plan future releases

jbygdell:
## Description

Given that the application stack is being used in several production grade projects by a number of other organizations it is time to start planning future releases a bit better.
* The scope of the release (feature addition, library updates or bigger changes)
* Which Issues that will be part of a specific release
* All PRs related to a release should be merged into a feature branch and only merged into main when the release is created.

### DIFF
--- a/docs/rfcs/0004-plan-releases.md
+++ b/docs/rfcs/0004-plan-releases.md
@@ -1,0 +1,32 @@
+---
+status: exploring
+date: "2026-05-28"
+authors:
+   - "@jbygdell"
+consulted: []
+informed: []
+---
+
+# Plan future releases
+
+## Context and Problem Statement
+
+Currently releases are created on every merge to the main branch without proper release notes or changelog.
+
+## Decision Drivers
+
+* **Structured releases** - Features should be planned in advance so that the codebase doesn't contain half completed features. For example implementation of "v2" shared components should be rolled out across the board so that application config stays consistent for all apps.
+
+## Considered Options
+
+* **Make use of Epics** - Use epics to group a set of Issues into a future release. The PRs for the Issues should not be merged into main on completion but rather into a branch dedicated to the upcoming release.
+* **Use release branches** - Start using branches for each upcoming release with the name format `Release-X.Y.Z`. Where `X.Y.Z` is the semantic version to be released.
+* **Be specific on the version** - The version tag for an upcoming release should reflect what it contains.
+  * Patch - Small bug fixes, Dependabot library updates and such - This is a transparent update that has no effect on configuration of a running application.
+  * Minor - Feature additions and larger bug fixes that can affect how an app is configured - Might require changes to a running application's configuration.
+  * Major - Breaking changes where a running application can not be rolled back to a previous version.
+
+## Open Questions
+
+* Decide on a future release strategy
+* Rework the RFC into an ADR


docs/rfcs/0004-plan-releases.md
```diff
@@ -0,0 +1,32 @@
...
+
+# Plan future releases
+
+## Context and Problem Statement
+
+Currently releases are created on every merge to the main branch without proper release notes or changelog.
+
+## Decision Drivers
+
+* **Structured releases** - Features should be planned in advance so that the codebase doesn't contain half completed features. For example implementation of "v2" shared components should be rolled out across the board so that applicaiton config stays consistent for all apps.
```
nanjiangshu:
```suggestion
* **Structured releases** - Features should be planned in advance so that the codebase doesn't contain half completed features. For example implementation of "v2" shared components should be rolled out across the board so that application config stays consistent for all apps.
```

docs/rfcs/0004-plan-releases.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Considered Options
+
+* **Make use of Epics** - Use epics to group a set if Issues into a future release. The PRs for the Issues should not be merged into main on completion but rather into a branch dedicated to the upcoming release.
```
nanjiangshu:
```suggestion
* **Make use of Epics** - Use epics to group a set of Issues into a future release. The PRs for the Issues should not be merged into main on completion but rather into a branch dedicated to the upcoming release.
```

docs/rfcs/0004-plan-releases.md
```diff
@@ -0,0 +1,32 @@
...
+* **Use release branches** - Start using branches for each upcoming relase with the name format `Release-X.Y.Z`. Where `X.Y.Z` is the sematic version to be released.
```
nanjiangshu:
```suggestion
* **Use release branches** - Start using branches for each upcoming release with the name format `Release-X.Y.Z`. Where `X.Y.Z` is the semantic version to be released.
```

docs/rfcs/0004-plan-releases.md
```diff
@@ -0,0 +1,32 @@
...
+* **Be specific on the version** - The version tag for an upcoming release should reflect what it contains.
+  * Patch - Small bug fixes, dependabot library updates and such - This is a transparent update that have no effect on configuration of a running application.
```
nanjiangshu:
```suggestion
  * Patch - Small bug fixes, Dependabot library updates and such - This is a transparent update that has no effect on configuration of a running application.
```

docs/rfcs/0004-plan-releases.md
```diff
@@ -0,0 +1,32 @@
...
+  * Minor - Feature additions and larger bug fixes that can affect how an app is configured - Might require changes to a running applications configuration.
```
nanjiangshu:
```suggestion
  * Minor - Feature additions and larger bug fixes that can affect how an app is configured - Might require changes to a running application's configuration.
```

docs/rfcs/0004-plan-releases.md
```diff
@@ -0,0 +1,32 @@
...
+  * Major - Breaking changes where a running application can not be rolled back to a previos version.
```
nanjiangshu:
```suggestion
  * Major - Breaking changes where a running application can not be rolled back to a previous version.
```

docs/rfcs/0004-plan-releases.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Open Questions
+
+* Descide on a future release strategy
```
nanjiangshu:
```suggestion
* Decide on a future release strategy
```

nanjiangshu (approved):

zeidlitz (approved):

docs/rfcs/0004-plan-releases.md
```diff
@@ -0,0 +1,32 @@
...
+   - "@jbygdell"
+consulted: []
+informed: []
+---
+
+# Plan future releases
+
+## Context and Problem Statement
+
+Currently releases are created on every merge to the main branch without proper release notes or changelog.
```
jhagberg: I am probably missing some context here, but I was not sure how to read this against #2439, which merged on the morning of the 28th and moved the release workflows to `workflow_dispatch` with an explicit version input, and #2448 the following day, which made the pipeline validate the changelog rather than write it (`release_sda.yml:47-57` now bails out if it cannot parse a previous version from `sda/CHANGELOG.md`).

Was the intent that this sentence describes the situation that motivated that work, rather than where we are now? If so, would it be worth saying so explicitly, so a later reader does not take it as a description of the current pipeline?

What still seems open to me is the part you describe in the PR body, deciding in advance what actually goes into a release, which #2439 does not really address. Would it be worth reframing the context around that?

docs/rfcs/0004-plan-releases.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Decision Drivers
+
+* **Structured releases** - Features should be planned in advance so that the codebase doesn't contain half completed features. For example implementation of "v2" shared components should be rolled out across the board so that application config stays consistent for all apps.
+
+## Considered Options
+
+* **Make use of Epics** - Use epics to group a set of Issues into a future release. The PRs for the Issues should not be merged into main on completion but rather into a branch dedicated to the upcoming release.
```
jhagberg: I might be reading these wrong, but they felt to me like parts of one proposal rather than alternatives to each other, since "Make use of Epics" already prescribes the dedicated branch, and the patch/minor/major definitions below seem to hold whichever branching model we pick.

Would it be worth listing a couple of alternatives alongside them? Staying trunk based with Milestones for the scope and tags cut from `main` would be one. Something like [release-please](https://github.com/googleapis/release-please-action) could be another, although our commit subjects are only partly conventional today, so that might be a bigger change than it first looks. Even if we end up where the RFC already points, would having the rejected options written down make the eventual ADR easier to write?

jhagberg (commented):
Thanks for writing this one up. I have been reading it alongside the release workflows and left a few questions inline, none of them blocking, and `exploring` seems like a fine state to merge in.

One that is not inline, since the file is not in the diff: should we add a row to the index table in `docs/rfcs/README.md`? I noticed RFC-0001 (#2263) and RFC-0002 (#2320) both add theirs in the same PR that adds the RFC file, while this one and RFC-0003 (#2453) do not, so I may be missing whether the plan was to fill the index in separately. Those two both replace the same placeholder line, so they will probably conflict with each other whichever merges first.

Two smaller ones. Would it be worth filling in `consulted` and `informed`, given that the PR body mentions other organisations running this in production? And would a `More Information` section help, pointing at the existing `release_v1` branch and ADR-0002 as prior art we already have in the repo, and at `CONTRIBUTING.md`, since this would touch how we branch and when we merge to `main`?

docs/rfcs/0004-plan-releases.md
```diff
@@ -0,0 +1,32 @@
...
+* **Use release branches** - Start using branches for each upcoming release with the name format `Release-X.Y.Z`. Where `X.Y.Z` is the semantic version to be released.
```
jhagberg: I was not sure which version `X.Y.Z` refers to here. `release_sda.yml:152-180` seems to ship seven images under one tag driven by `sda/CHANGELOG.md`, covering sda, validator orchestrator, download, doa, postgres, rabbitmq and sftp inbox, while `sda-admin` and `charts/sda-svc` have their own workflows and changelogs.

Was the intent that a release branch covers only that train, or that the chart and the CLI would follow it too? I also noticed the branch we already have is called `release_v1` rather than a full `X.Y.Z`, so maybe the naming here is still open?

There is also `sda-cli`, which a release already makes a statement about through `app.minimalcliversion` (`sda-download/internal/config/config.go:327`). I have a vague memory of a discussion about moving it into this repo. If that happened, would it join the same train, and either way should the RFC say how the minimum client version is picked for a release?

docs/rfcs/0004-plan-releases.md
```diff
@@ -0,0 +1,32 @@
...
+* **Be specific on the version** - The version tag for an upcoming release should reflect what it contains.
+  * Patch - Small bug fixes, Dependabot library updates and such - This is a transparent update that has no effect on configuration of a running application.
```
jhagberg: I could not quite tell from this whether Dependabot PRs are meant to go to the release branch, to `main`, or to both while a branch is open. Since they sit in the patch category here, I assume they should keep flowing?

I noticed we already have something close to this in `release_v1`, where `.github/dependabot.yaml` carries a second full set of package managers with `target-branch: release_v1`, from line 64 to the end of the file. If a branch is named `Release-X.Y.Z` and so changes name every release, would not that config need editing each cycle, unlike the stable `release_v1` name?

More generally, would it be worth a short pros and cons section? I was mostly looking for what a release branch costs us, since a branch held open for a while also tends to drift and integration only really gets tested at merge time.

docs/rfcs/0004-plan-releases.md
```diff
@@ -0,0 +1,32 @@
...
+  * Minor - Feature additions and larger bug fixes that can affect how an app is configured - Might require changes to a running application's configuration.
+  * Major - Breaking changes where a running application can not be rolled back to a previous version.
+
+## Open Questions
+
+* Decide on a future release strategy
+* Rework the RFC into an ADR
```
jhagberg: Since the README describes this section as the main output of an RFC, I wondered whether these two are meant as placeholders for now. "Decide on a future release strategy" reads to me as close to the title, and "Rework the RFC into an ADR" is the promotion step, so I was not sure what we would be converging on.

Would it help to put the things we do not agree on yet here instead? I was thinking of ones like: where does a hotfix branch from while a release branch is open (`release_sda.yml:22-33` already allows `hotfix/*`), how long can a release branch reasonably live, who decides the scope and who cuts the release, whether `main` stays releasable while the branch stabilises, and whether the three approvals from three partners rule applies again to the release branch merge. Or perhaps some of those are already settled and I have just missed the discussion?

docs/rfcs/0004-plan-releases.md
jbygdell: This RFC where written before the changes to how the releases are done now and simply not updated.
That change in the release handling is probably in part due to the RFC.

docs/rfcs/0004-plan-releases.md
jbygdell: The X.Y.Z would refer to  the targeted semantic version, in most cases Z would be `0` since feature additions warrant a minor release bump.

docs/rfcs/0004-plan-releases.md
jbygdell: Given that we are thinking of deprecating Dependabot or at least run it very seldom it's more that kind of changes that dependabot does currently i.e. updates libraries of applications that does no see active development.

docs/rfcs/0004-plan-releases.md
jbygdell: fixed

jhagberg (approved):
Looks good to me, approving as `exploring`. One small thing you could fold in now or later: your note in the thread that Dependabot may be deprecated or run rarely feels like it belongs in the file, since it changes what a release branch has to carry. And if you want Open Questions to be more than the title, a couple of the concrete ones from my earlier comment could go there, but that can equally land in a follow-up PR.

PR #2453, opened 2026-05-29
docs(rfc): RFC-0003 Rename the `api` app

jbygdell:
## Description

The API used to perform administrative tasks is simply called `api`. That's not a very apt name considering that all other apps are named by the tasks they perform.
Changing the name to `admin` is not a good way to go since it then will clash with the CLI tool `sda-admin`.

The api app should be renamed with some key considerations taken into account. 
* Name should be descriptive 
* Name should not easily be confused with another app or tool.

### DIFF
--- a/docs/rfcs/0003-rename-api-app.md
+++ b/docs/rfcs/0003-rename-api-app.md
@@ -1,0 +1,27 @@
+---
+status: exploring
+date: "2026-05-28"
+authors:
+   - "@jbygdell"
+consulted: []
+informed: []
+---
+
+# Rename the API app
+
+## Context and Problem Statement
+
+Currently the management API is located under `sda/cmd/api` and when built the container command becomes `sda-api`. Since it can perform administrative tasks it is easy to refer to it as `sda-admin`, somthing that conflicts with the CLI tool with the same name.
+
+## Decision Drivers
+
+* **Consistency** - applications should have a name that is related to the tasks they perform
+
+## Considered Options
+
+* **`admin-api`** - Not a good choice since through RBAC restrictions submitters can also use the app.
+* **`manager` or `management`** - Possible new name that is better describes the application.
+
+## Open Questions
+
+* Decide on renaming the API app


docs/rfcs/0003-rename-api-app.md
```diff
@@ -0,0 +1,27 @@
+---
+status: exploring
+date: "2026-05-28"
+authors:
+   - "@jbygdell"
+consulted: []
+informed: []
+---
+
+# Rename the APi app
```
nanjiangshu:
```suggestion
# Rename the API app
```

docs/rfcs/0003-rename-api-app.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Context and Problem Statement
+
+Currently the management API is locates unser `sda/cmd/api` and when built the comntainer command becomes `sda-api`. Since it can perform administrative tasks it is easy to refer to it as `sda-admin`, somthing that conflicts with the CLI tool with the same name.
```
nanjiangshu:
```suggestion
Currently the management API is located under `sda/cmd/api` and when built the container command becomes `sda-api`. Since it can perform administrative tasks it is easy to refer to it as `sda-admin`, something that conflicts with the CLI tool with the same name.
```

docs/rfcs/0003-rename-api-app.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Decision Drivers
+
+* **Consistency** - applications should have a name that is related to the tasks they perform
+
+## Considered Options
+
+* **`admin-api`** - Not a good choise since through RBAC restrictions submitters can also use the app.
```
nanjiangshu:
```suggestion
* **`admin-api`** - Not a good choice since through RBAC restrictions submitters can also use the app.
```

docs/rfcs/0003-rename-api-app.md
```diff
@@ -0,0 +1,27 @@
...
+* **`manager` or `management`** - Posible new name that is better in describing the application.
```
nanjiangshu:
```suggestion
* **`manager` or `management`** - Possible new name that is better describes the application.
```

docs/rfcs/0003-rename-api-app.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Open Questions
+
+* Descide on renaming the api app
```
nanjiangshu:
```suggestion
* Decide on renaming the API app
```

nanjiangshu (approved):

zeidlitz (dismissed):
In my opinion the name `api` is perfectly fine and very descriptive of what the service / application is doing. I think the problem described here would best be solved by just removing the `sda-admin` tool entirely. In my opinion the `sda-admin` tool don't serve a good enough purpose to justify renaming the `api` app to something less descriptive. 

My proposed solution would be:

keep the name `api`, remove the `sda-admin` tool . The benefits would be less overhead in the repository. Less code to maintain and overall just less complexity. 

Even if we wan't to keep the `sda-admin` cli tool. I don't think the risk of confusion is a problem worth renaming the `api` service for.

KarlG-nbis (approved):
Agreed, the "API" is a bit generic and not very descriptive since we have other services which also has APIs(eg, download), and possibly worth separating out the "user APIs" and the "admin APIs" to separate services, etc

jhagberg:
I like that this is an RFC rather than an ADR, the naming question is clearly still open.

Reading back through the thread, the substance of the discussion has not made it into the file yet. @zeidlitz argued for keeping `api` and removing the `sda-admin` tool instead, @KarlG-nbis suggested splitting the user APIs and the admin APIs into separate services, and the file is still the single commit from May with the two options it started with. That is not really on you, the RFC README never said what an author should do with review feedback. I have a PR open in #2553 that fills that gap, and it would make this concrete: a new option goes under `## Considered Options`, an objection to an option goes into its Pros and Cons, and anything still unresolved goes into `## Open Questions`.

Applying that here, would it be worth folding both of those in before this merges? @zeidlitz's point in particular goes at the premise rather than the wording, and I see the review was dismissed as AWOL, which unblocks the merge but leaves the argument unanswered. Step 6 in #2553 suggests that an objection still live at merge time goes in as a named Open Question and gets taken up at the next SDA-Devs meet-up. Would that work here? Someone reading the file in a year should not have to open this thread to learn that we considered keeping `api`.

A few smaller things.

@nanjiangshu's suggestion on line 14 (the `somthing` typo) is still unapplied. The other four look byte-identical to the current text, so I think those can just be resolved.

`docs/rfcs/README.md` has no index row for 0003. #2449 is missing one too, and the README's step list never mentioned it, which is probably why. #2553 moves that into step 3.

On the options themselves, `manager` or `management` sits close to `sda/cmd/orchestrate` and the separate `sda-validator/orchestrator`. Would three control-plane-sounding names just move the confusion rather than remove it? And the rename is a breaking change for deployers, `charts/sda-svc/templates/api-deploy.yaml:75` and `.github/integration/sda-s3-integration.yml:215` both pin `sda-api` as the container command. Maybe a decision driver about migration cost?

Also `## Open Questions` currently only has "Decide on renaming the API app", which is the title restated. Could it hold what we genuinely cannot answer yet, such as which name, whether the churn is worth it, and whether the binary name has to keep tracking the directory name (`sda/Dockerfile:6` derives it, so a directory rename changes `sda-api` on its own)?

None of this needs a conclusion before merging, `exploring` is fine as an end state.

docs/rfcs/0003-rename-api-app.md
```diff
@@ -0,0 +1,30 @@
...
+Currently the management API is located under `sda/cmd/api` and when built the container command becomes `sda-api`. Since it can perform administrative tasks it is easy to refer to it as `sda-admin`, something that conflicts with the CLI tool with the same name.
+
+## Decision Drivers
+
+* **Consistency** - applications should have a name that is related to the tasks they perform
+
+## Considered Options
+
+* **`admin-api`** - Not a good choice since through RBAC restrictions submitters can also use the app.
+* **`manager` or `management`** - Possible new name that is better describes the application.
```
MalinAhlberg:
```suggestion
* **`manager` or `management`** - Possible new name that better describes the application.
```

docs/rfcs/0003-rename-api-app.md
```diff
@@ -0,0 +1,30 @@
...
+* Keeping the name **`api`** - Based on the fact that we have other APIs that are task specific and thus have matching names this should be no different.
```
MalinAhlberg:
```suggestion
* Keeping the name **`api`** - Based on the fact that we have other APIs that are task specific and thus have matching names this should be no different.
* Keeping the name **`api`** and removing `sda-admin` tool

```

docs/rfcs/0003-rename-api-app.md
```diff
@@ -0,0 +1,30 @@
...
+* Splitting the code into `submitter` and `admin/management` parts together with new names for the apps.
+
+## Open Questions
+
+* Decide on renaming the API app
+* Split the code into `submitter` and `admin/management` parts
```
MalinAhlberg:
```suggestion
* Split the code into `submitter` and `admin/management` parts

## Pros and Cons
### Keeping the name api

* Good, because it avoid additional work

### Keeping the name api and removing the sda-admin tool

* Good, because gives less overhead in the repository; less code to maintain and overall just less complexity.
```

MalinAhlberg:
```suggestion
* Split the code into `submitter` and `admin/management` parts

## Pros and Cons
### Keeping the name api
* Good, because we would avoid additional work

### Keeping the name api and removing the sda-admin tool
* Good, because gives less overhead in the repository; less code to maintain and overall just less complexity.
```

MalinAhlberg (changes requested):
Good points and discussion!
I found a minor typo, and I agree that the different options/ideas from the discussion thread should be added in the file, as @jhagberg suggests.

Just to get the hang of how you are working nowadays, I tried to add Erik's suggestions to the document. Please let me know if and how it could be improved :)

MalinAhlberg (changes requested):
Good points and discussion!
I found a minor typo, and I agree that the different options/ideas from the discussion thread should be added in the file, as @jhagberg suggests.

Just to get the hang of how you are working nowadays, I tried to add Erik's suggestions to the document. Please let me know if and how it could be improved :)

MalinAhlberg (approved):

jhagberg:
Thanks both, nice to see the options from the thread in the file now.

I pushed a small commit on top. The "Decision records and RFCs" check was failing on MD022, `## Pros and Cons` needed a blank line before the `###` heading. And now that #2553 is on main, step 4 asks for the PR URL in a `discussion:` front-matter field, so I added that too, this is the first RFC to carry one. Hope that is OK, shout if you would rather I had left it as a suggestion.

Two things I did not touch since they are wording rather than mechanics. Line 39 says "it avoid" and line 43 says "because gives", so "it avoids" and "because it gives", or @MalinAhlberg's "we would avoid" from the still open thread. Applying that suggestion as-is would drop the blank lines again though, so probably just the words.

And the Pros and Cons only have Good entries so far. Would it be fair to add that keeping `api` leaves the `sda-admin` confusion in place, and that removing `sda-admin` drops a tool the top-level README documents and CI builds? The rest of my earlier comment can wait for a follow-up PR, step 6 allows that.

PR #2460, opened 2026-06-03
test(reencrypt): pin re-encrypted header length-stable, bytes-unstable invariant

jhagberg:
## Related issue(s) and PR(s)
Guards the re-encrypted header stability contract documented in `sda/cmd/download/swagger_v2.yml` for the download-v2 `/files/{fileId}/header` endpoint. No dedicated issue — arose from a client (sda-download-ui) question about whether header bytes are stable per `(fileId, recipientKey)`.

# Description
Adds a regression test, `TestReencryptHeader_LengthStableBytesDiffer`, to the reencrypt service suite.

The download-v2 API documents that re-encrypted headers are recipient-specific and that byte-stability across requests is implementation-defined. In practice the reencrypt service (via crypt4gh `ReEncryptHeader`) draws a fresh ephemeral writer keypair and nonce on every call, so for the same `(header, recipient key)`:

- the header **length** is stable — clients can trust `Content-Length`;
- the header **bytes differ** on each call — clients must not cache or byte-compare headers; resume belongs on `/files/{fileId}/content`.

The test pins both halves of that contract, and additionally asserts both headers are valid Crypt4GH, still decrypt the body, and neither carries a `DataEditList` packet on the simple (no-edit-list) path — the packet structure that keeps the length stable. This stops a future change to the re-encryption path from silently breaking the client contract.

Test-only; no production code changes.

### DIFF
--- a/sda/cmd/reencrypt/reencrypt_test.go
+++ b/sda/cmd/reencrypt/reencrypt_test.go
@@ -7,6 +7,7 @@ import (
 	"crypto/x509"
 	"encoding/base64"
 	"encoding/hex"
+	"errors"
 	"io"
 	"net"
 	"os"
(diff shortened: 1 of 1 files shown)


jbygdell (approved):

nanjiangshu (approved):
Looks good to me!

PR #2464, opened 2026-06-04
feat(sda): config-driven projectCode inbox-path reconstruction

yasinmiran:
## Problem (relates to #2429)

In the FEGA-Norway deployment, inbox files land under a project-code-prefixed per-user directory — `p11-dummy@elixir-europe.org/files/x.raw.enc` (TSD upload, **not** s3inbox). Stock SDA's read-side `UnanonymizeFilepath` assumes the per-user directory is exactly the username, so it mis-resolves the path and ingest/mapper can't find or clean the file.

## Approach — config-driven reconstruction (additive, default == stock)

The FEGA-Norway proxy anonymizes the path it publishes to `/files/<file>` (project code + username stripped — see ELIXIR-NO/FEGA-Norway#819), so the message stays a clean, node-agnostic path. SDA reconstructs the physical inbox path from config:

- `storage.inbox.projectCode` — optional per-user-directory prefix (e.g. `p11`); empty (the default) ⇒ stock layout.
- `storage.inbox.projectCodeDelimiter` — separator between the project code and the username (e.g. `-`).

`ResolveInboxPath(filePath, user, InboxConfig)` reconstructs the per-user directory and joins the path under it:

- **No project code** → defers to the stock `UnanonymizeFilepath` (normalize the username, prepend `<user>/`), so existing deployments are byte-for-byte unaffected.
- **A project code** → rebuilds `<projectCode><delimiter><username>/<path>` using the raw username. Whether to normalize is **derived from the project code** rather than a separate toggle: a project code denotes a TSD-style inbox that stores the raw username verbatim. (No current deployment needs a project code *with* normalization; an explicit toggle can be added if one ever does.)

`helper.go` is **purely additive** — only `InboxConfig` + `ResolveInboxPath` are added; `AnonymizeFilepath`/`UnanonymizeFilepath` are left untouched. Config flows through a single `config.LoadInboxConfig` read (mapper/api via `NewConfig`; ingest directly, since on `main` it loads through `config/v2` but shares the same viper instance). Wired into every inbox-path site: ingest (find + read), mapper (delete), api (delete + read). FEGA-Norway sets `projectCode=p11`, `projectCodeDelimiter="-"`.

## Verification

`go build`/`vet`/`test` green, including unit tests for `ResolveInboxPath` and a `LoadInboxConfig` seam test (nested `storage.inbox.*` YAML → the dotted keys it reads). End-to-end in the FEGA-Norway stack: the full pipeline ran green (9/9) on the `v3.1.61` images we deploy — exercising ingest's reconstruction and the mapper's inbox cleanup — and ingest's reconstruction is re-confirmed on the `v3.1.72` images. Design rationale + the stock-behavior trace: ELIXIR-NO/FEGA-Norway#820.

### DIFF
--- a/sda/internal/config/config.go
+++ b/sda/internal/config/config.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/neicnordic/crypt4gh/keys"
 	"github.com/neicnordic/sensitive-data-archive/internal/broker"
+	"github.com/neicnordic/sensitive-data-archive/internal/helper"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
@@ -40,6 +41,7 @@ type Config struct {
 	ReEncrypt    ReEncConfig
 	Auth         AuthConf
 	RotateKey    RotateKeyConf
+	Inbox        helper.InboxConfig
 }
 
 type Grpc struct {
@@ -376,6 +378,7 @@ func NewConfig(app string) (*Config, error) {
 			return nil, err
 		}
 		c.configSchemas()
+		c.configInbox()
 
 		c.API.Grpc, err = configReEncryptClient()
 		if err != nil {
@@ -451,6 +454,7 @@ func NewConfig(app string) (*Config, error) {
 		}
 
 		c.configSchemas()
+		c.configInbox()
 	case "notify":
 		c.configSMTP()
 
@@ -766,6 +770,24 @@ func (c *Config) configReEncryptServer() (err error) {
 	}
 
 	return nil
+}
+
+// LoadInboxConfig reads the per-user inbox directory layout from the shared viper instance. An
+// empty (absent) project code yields stock SDA behavior, so deployments that omit the section are
+// unaffected. This is the single source of the storage.inbox.* keys, read both by NewConfig
+// (mapper, api) and directly by the ingest service, which loads through config/v2 but populates
+// the same viper instance.
+func LoadInboxConfig() helper.InboxConfig {
+	return helper.InboxConfig{
+		ProjectCode:          viper.GetString("storage.inbox.projectCode"),
+		ProjectCodeDelimiter: viper.GetString("storage.inbox.projectCodeDelimiter"),
+	}
+}
+
+// configInbox populates the inbox layout config for services that resolve anonymized submission
+// paths back to their physical inbox path (mapper, api).
+func (c *Config) configInbox() {
+	c.Inbox = LoadInboxConfig()
 }
 
 // configSchemas configures the schemas to load depending on
--- a/sda/internal/helper/helper.go
+++ b/sda/internal/helper/helper.go
@@ -464,3 +464,39 @@ func UnanonymizeFilepath(fp string, username string) string {
 
 	return filepath.Join(strings.Replace(username, "@", "_", 1), fp)
 }
+
+// InboxConfig describes how a deployment names its per-user inbox directories, so an
+// anonymized submission path (the user prefix stripped) can be resolved back to the physical
+// inbox-relative path. An empty ProjectCode selects stock SDA behavior.
+type InboxConfig struct {
+	// ProjectCode optionally prefixes the per-user inbox directory (e.g. "p11"). Empty keeps the
+	// stock layout where the directory is the normalized username (see UnanonymizeFilepath).
+	ProjectCode string
+	// ProjectCodeDelimiter separates ProjectCode from the username (e.g. "-"). Unused when
+	// ProjectCode is empty.
+	ProjectCodeDelimiter string
+}
+
+// ResolveInboxPath reconstructs the physical inbox-relative path for an anonymized submission
+// filePath. An already-resolved path (e.g. on reprocessing) is returned unchanged.
+//
+// With no ProjectCode it is the stock round-trip, deferring to UnanonymizeFilepath (normalize the
+// username — first "@"->"_" — and prepend "<user>/"), so existing deployments are unaffected.
+//
+// With a ProjectCode it rebuilds "<ProjectCode><delimiter><username>/<filePath>" using the RAW
+// username. Whether to normalize is derived from the project code rather than configured
+// separately: a project code denotes a TSD-style inbox namespaced by project (e.g. FEGA-Norway's
+// "p11-dummy@elixir-europe.org/files/..."), which stores the username verbatim. No current
+// deployment needs a project code together with normalization; add an explicit toggle if one does.
+func ResolveInboxPath(filePath, username string, cfg InboxConfig) string {
+	if cfg.ProjectCode == "" {
+		return UnanonymizeFilepath(filePath, username)
+	}
+
+	userDir := cfg.ProjectCode + cfg.ProjectCodeDelimiter + username
+	if strings.HasPrefix(filePath, userDir) {
+		return filePath
+	}
+
+	return filepath.Join(userDir, filePath)
+}
--- a/sda/cmd/api/api.go
+++ b/sda/cmd/api/api.go
@@ -544,7 +544,7 @@ func deleteFile(c *gin.Context) {
 		return
 	}
 
-	filePath = helper.UnanonymizeFilepath(filePath, submissionUser)
+	filePath = helper.ResolveInboxPath(filePath, submissionUser, Conf.Inbox)
 	for count := 1; count <= 5; count++ {
 		err = inboxWriter.RemoveFile(c, location, filePath)
 		if err == nil {
@@ -639,9 +639,10 @@ func downloadFile(c *gin.Context) {
 
 	// Get inbox file handle #noqa
 	file, err := inboxReader.NewFileReader(c, location,
-		helper.UnanonymizeFilepath(
+		helper.ResolveInboxPath(
 			filePath,
 			strings.TrimPrefix(c.Param("username"), "/"),
+			Conf.Inbox,
 		),
 	)
 	if err != nil {
--- a/sda/cmd/ingest/ingest.go
+++ b/sda/cmd/ingest/ingest.go
@@ -43,6 +43,7 @@ type Ingest struct {
 	ArchiveKeyList []*[32]byte
 	db             database.Database
 	InboxReader    storage.Reader
+	InboxConfig    helper.InboxConfig
 	Broker         brokerv2.Broker
 }
 
@@ -69,6 +70,7 @@ func run() error {
 	if err = configv2.Load(); err != nil {
 		return fmt.Errorf("failed to load config: %v", err)
 	}
+	app.InboxConfig = config.LoadInboxConfig()
 
 	app.Broker, err = rabbitmq.NewRabbitMQBroker(context.Background())
 	if err != nil {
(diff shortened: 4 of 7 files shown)


zeidlitz (approved):
Looks pretty straight forward to me! No comments on my side

sda/internal/config/config.go
```diff
@@ -768,6 +899,19 @@ func (c *Config) configReEncryptServer() (err error) {
 	return nil
 }
 
+// configInbox reads the per-user inbox directory layout settings used to resolve an anonymized
+// submission path back to its physical inbox path. The defaults (no project code, normalize the
+// username) reproduce stock SDA behavior, so existing deployments are unaffected.
+func (c *Config) configInbox() {
+	viper.SetDefault("storage.inbox.projectCode", "")
+	viper.SetDefault("storage.inbox.projectCodeDelimiter", "")
```
jbygdell:
```suggestion
```

Default values should not be needed for these since the null value of a string variable is `""`

sda/internal/config/config.go
```diff
@@ -768,6 +899,19 @@ func (c *Config) configReEncryptServer() (err error) {
...
+	viper.SetDefault("storage.inbox.normalizeUsername", true)
```
jbygdell: Consider flipping this one around to that when enabled it uses the project prefixed name instead.

sda/internal/helper/helper.go
```diff
@@ -453,14 +453,76 @@ func CreatePrivateKeyFile(keyFile string, passphrase string) ([32]byte, error) {
 	return publicKey, nil
 }
 
+// AnonymizeFilepath strips the leading "<normalized-user>/" directory from an inbox path, so the
+// stored and published path is independent of the submitting user. It is the write-side inverse of
+// UnanonymizeFilepath; the username is normalized (see normalizeUsername) to match the on-disk
+// inbox directory name.
 func AnonymizeFilepath(fp string, username string) string {
-	return strings.ReplaceAll(fp, strings.Replace(username, "@", "_", 1)+"/", "")
+	return strings.ReplaceAll(fp, normalizeUsername(username)+"/", "")
```
jbygdell: This is variable is a path, never concatenate those manually use the filepath library already present in the code to join the parts.

sda/internal/helper/helper.go
```diff
@@ -453,14 +453,76 @@ func CreatePrivateKeyFile(keyFile string, passphrase string) ([32]byte, error) {
...
+// inbox-relative path.
+type InboxConfig struct {
+	// ProjectCode is an optional prefix on the per-user directory (e.g. "p11"). Empty means
+	// the directory is just the username, matching stock SDA.
+	ProjectCode string
+	// ProjectCodeDelimiter separates ProjectCode from the username (e.g. "-"). Unused when
+	// ProjectCode is empty.
+	ProjectCodeDelimiter string
+	// NormalizeUsername replaces the first "@" in the username with "_" (stock SDA behavior).
+	NormalizeUsername bool
```
jbygdell: This shouldn't be needed, the presence of the project code prefix should be enough

jbygdell (changes requested):
Rebase on main before you open up the PR for proper review.

Although maybe outside of your control it would be better to have the project ID as a  subpath prefix since that would solve more than just your issue here.
I.e that the file path becomes `VOLUME_ROOT/PROJECT-ID/USER-ID/PATH/TO/FILES`

sda/internal/helper/helper.go
```diff
@@ -453,14 +453,76 @@ func CreatePrivateKeyFile(keyFile string, passphrase string) ([32]byte, error) {
...
 }
 
+// AnonymizeFilepath strips the leading "<normalized-user>/" directory from an inbox path, so the
+// stored and published path is independent of the submitting user. It is the write-side inverse of
+// UnanonymizeFilepath; the username is normalized (see normalizeUsername) to match the on-disk
+// inbox directory name.
 func AnonymizeFilepath(fp string, username string) string {
-	return strings.ReplaceAll(fp, strings.Replace(username, "@", "_", 1)+"/", "")
+	return strings.ReplaceAll(fp, normalizeUsername(username)+"/", "")
 }
```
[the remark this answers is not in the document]
yasinmiran: This PR no longer modifies `AnonymizeFilepath` — it's reverted to be byte-identical to upstream `main`, so the `ReplaceAll`-vs-`TrimPrefix` behavior is pre-existing and out of scope for this inbox-path change. Not touching it here.

sda/internal/helper/helper.go
```diff
@@ -453,14 +453,76 @@ func CreatePrivateKeyFile(keyFile string, passphrase string) ([32]byte, error) {
...
+// already starts with that directory. The username is normalized (see normalizeUsername).
 func UnanonymizeFilepath(fp string, username string) string {
-	if strings.HasPrefix(fp, strings.Replace(username, "@", "_", 1)) {
+	normalizedUser := normalizeUsername(username)
+	if strings.HasPrefix(fp, normalizedUser) {
 		return fp
 	}
 
-	return filepath.Join(strings.Replace(username, "@", "_", 1), fp)
+	return filepath.Join(normalizedUser, fp)
```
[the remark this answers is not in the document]
yasinmiran: Same here — `UnanonymizeFilepath` is unchanged from upstream in this PR, so it's out of scope. For the record, the absolute-path concern doesn't apply: `filepath.Join` cleans a leading separator on its second argument, so `Join("user", "/files/x")` → `"user/files/x"`, not an absolute path.

sda/internal/helper/helper.go
```diff
@@ -453,14 +453,76 @@ func CreatePrivateKeyFile(keyFile string, passphrase string) ([32]byte, error) {
...
+	if cfg.ProjectCode != "" {
+		userDir = cfg.ProjectCode + cfg.ProjectCodeDelimiter + user
+	}
+
+	if strings.HasPrefix(filePath, userDir) {
+		return filePath
+	}
+
+	return filepath.Join(userDir, filePath)
+}
```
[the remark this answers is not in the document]
yasinmiran: This one is on our new function, so addressing it directly:

1. `filepath.Join(userDir, "/files/x")` does **not** drop `userDir`. `Join` cleans and joins; a leading `/` on the second arg is not treated as absolute. `TestResolveInboxPath_projectCode_reconstructsRawUserDir` feeds exactly `/files/x.raw.enc` and asserts `p11-dummy@elixir-europe.org/files/x.raw.enc`.
2. The directory-boundary case (`userDir="bob"`, `filePath="bobby/..."`) isn't reachable in this flow: the anonymized submission path is always `/files/<f>` and never begins with the user dir, so the `HasPrefix` branch only fires on reprocessing where `filePath` already **is** the resolved dir. Leaving as-is.

sda/internal/config/config.go
yasinmiran: Done — removed. The inbox config now reads through `LoadInboxConfig`, which calls `viper.GetString` with no `SetDefault`; an absent `storage.inbox.*` section yields the zero value (`""`), i.e. stock behavior.

sda/internal/config/config.go
yasinmiran: Resolved by removing the toggle entirely. Whether to normalize is now derived from the project code: no project code → stock normalize (defer to `UnanonymizeFilepath`); a project code → raw username. There's no `normalizeUsername` flag to flip anymore.

sda/internal/helper/helper.go
yasinmiran: The line you flagged (`AnonymizeFilepath`) is reverted — that function is untouched/stock now. The only remaining manual concatenation is in `ResolveInboxPath`: `cfg.ProjectCode + cfg.ProjectCodeDelimiter + username`. That builds a single directory **name** (`p11-dummy@elixir-europe.org`), where the delimiter `-` is part of the name, not a path separator — so `filepath.Join` would be wrong here (it would insert a `/`). Where there genuinely is a separator — joining that dir onto the file path — we do use `filepath.Join(userDir, filePath)`.

sda/internal/helper/helper.go
yasinmiran: Agreed — dropped it. `InboxConfig` is now just `{ProjectCode, ProjectCodeDelimiter}`; normalization is derived from whether a project code is set, exactly as you suggested (the presence of the project code is enough).

sda/internal/helper/helper.go
```diff
@@ -453,14 +453,67 @@ func CreatePrivateKeyFile(keyFile string, passphrase string) ([32]byte, error) {
...
+		return UnanonymizeFilepath(filePath, username)
+	}
+
+	userDir := cfg.ProjectCode + cfg.ProjectCodeDelimiter + username
+	if strings.HasPrefix(filePath, userDir) {
+		return filePath
+	}
+
+	return filepath.Join(userDir, filePath)
+}
```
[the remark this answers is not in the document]
yasinmiran: `filepath.Join` does not treat a leading `/` on its second argument as absolute — it cleans and joins, so `Join("p11-dummy@elixir-europe.org", "/files/x.raw.enc")` → `"p11-dummy@elixir-europe.org/files/x.raw.enc"`. `TestResolveInboxPath_projectCode_reconstructsRawUserDir` feeds exactly `/files/x.raw.enc` and asserts the prefixed result; it passes. So `userDir` is not dropped and the resolved path is correct.

sda/internal/helper/helper.go
```diff
@@ -453,14 +453,67 @@ func CreatePrivateKeyFile(keyFile string, passphrase string) ([32]byte, error) {
 	return publicKey, nil
 }
 
+// AnonymizeFilepath strips the leading "<normalized-user>/" directory from an inbox path, so the
+// stored and published path is independent of the submitting user. It is the write-side inverse of
+// UnanonymizeFilepath; the username is normalized (see normalizeUsername) to match the on-disk
+// inbox directory name.
 func AnonymizeFilepath(fp string, username string) string {
-	return strings.ReplaceAll(fp, strings.Replace(username, "@", "_", 1)+"/", "")
+	return strings.ReplaceAll(fp, normalizeUsername(username)+"/", "")
```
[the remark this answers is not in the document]
yasinmiran: Reverted — `AnonymizeFilepath` is now byte-identical to upstream `main`, so this is pre-existing behavior and out of scope for this PR.

sda/internal/helper/helper.go
```diff
@@ -453,14 +453,67 @@ func CreatePrivateKeyFile(keyFile string, passphrase string) ([32]byte, error) {
...
+// already starts with that directory. The username is normalized (see normalizeUsername).
 func UnanonymizeFilepath(fp string, username string) string {
-	if strings.HasPrefix(fp, strings.Replace(username, "@", "_", 1)) {
+	normalizedUser := normalizeUsername(username)
+	if strings.HasPrefix(fp, normalizedUser) {
 		return fp
 	}
 
-	return filepath.Join(strings.Replace(username, "@", "_", 1), fp)
+	return filepath.Join(normalizedUser, fp)
```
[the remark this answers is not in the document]
yasinmiran: Reverted — `UnanonymizeFilepath` is now byte-identical to upstream `main`; pre-existing and out of scope for this PR.

sda/internal/config/config.go
```diff
@@ -768,6 +772,24 @@ func (c *Config) configReEncryptServer() (err error) {
...
+// LoadInboxConfig reads the per-user inbox directory layout from the shared viper instance. An
+// empty (absent) project code yields stock SDA behavior, so deployments that omit the section are
+// unaffected. This is the single source of the storage.inbox.* keys, read both by NewConfig
+// (mapper, api) and directly by the ingest service, which loads through config/v2 but populates
+// the same viper instance.
+func LoadInboxConfig() helper.InboxConfig {
+	return helper.InboxConfig{
+		ProjectCode:          viper.GetString("storage.inbox.projectCode"),
+		ProjectCodeDelimiter: viper.GetString("storage.inbox.projectCodeDelimiter"),
+	}
```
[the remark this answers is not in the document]
yasinmiran: Good catch — the `storage.inbox.normalizeUsername` toggle was intentionally removed during review. Normalization is now derived from the project code (no project code → stock normalize; project code → raw username), so the only keys are `projectCode` + `projectCodeDelimiter`. I've updated the PR description to match the implemented behavior.

PR #2466, opened 2026-06-08
feat(sda): config-driven projectCode inbox paths + non-s3inbox ingest catch-all fix

yasinmiran:
> **Supersedes #2464.** Moved to an upstream branch so the Docker image-build and integration-test CI can run (a fork PR can't push to `ghcr.io/neicnordic`). The design was reviewed on #2464; review continues here.

This PR makes two related changes on the read/ingest side of the inbox path, both surfaced by the FEGA-Norway deployment (TSD upload, **not** s3inbox):

1. config-driven projectCode inbox-path reconstruction (additive, default == stock), relates to #2429
2. fix for the non-s3inbox ingest catch-all regression, see #2467

---

## 1. projectCode inbox-path reconstruction

In the FEGA-Norway deployment, inbox files land under a project-code-prefixed per-user directory: `p11-dummy@elixir-europe.org/files/x.raw.enc`. Stock SDA's read-side `UnanonymizeFilepath` assumes the per-user directory is exactly the username, so it mis-resolves the path and ingest/mapper can't find or clean the file.

The FEGA-Norway proxy anonymizes the path it publishes to `/files/<file>` (project code + username stripped, see ELIXIR-NO/FEGA-Norway#819), so the message stays a clean, node-agnostic path. SDA reconstructs the physical inbox path from config:

- `storage.inbox.projectCode`: optional per-user-directory prefix (e.g. `p11`); empty (the default) means stock layout.
- `storage.inbox.projectCodeDelimiter`: separator between the project code and the username (e.g. `-`).

The loader validates the pair at startup: setting only one of the two keys is an error, the delimiter is restricted to an allowlist (`-`, `_`, `.`), and the code must not contain path separators or whitespace, so a misconfiguration fails the service instead of producing garbage paths at runtime. Both keys absent (the default) skips validation entirely.

`ResolveInboxPath(filePath, user, InboxProjectConfig)` reconstructs the per-user directory and joins the path under it:

- **No project code** falls back to the stock `UnanonymizeFilepath` (normalize the username, prepend `<user>/`), so existing deployments are byte-for-byte unaffected.
- **A project code** rebuilds `<projectCode><delimiter><username>/<path>` using the raw username. Whether to normalize is **derived from the project code** rather than a separate toggle: a project code denotes a TSD-style inbox that stores the raw username verbatim.

`helper.go` is **purely additive**: only `InboxProjectConfig` + `ResolveInboxPath` are added; `AnonymizeFilepath`/`UnanonymizeFilepath` are untouched. Config flows through a single `config.LoadInboxProjectConfig` read (mapper/api via `NewConfig`; ingest directly, since on `main` it loads through `config/v2` but shares the same viper instance). Wired into every inbox-path site: ingest (find + read), mapper (delete), api (delete + read). FEGA-Norway sets `projectCode=p11`, `projectCodeDelimiter="-"`.

## 2. non-s3inbox ingest catch-all fix

The status `""` catch-all, where ingest is the **first** registration point because the upload side does not pre-register via s3inbox/sftp-inbox, regressed in the v2 migration. It located and registered the file using `message.Key` (the broker correlation-id) instead of the trigger filepath, and returned after `RegisterFile` without archiving, so a non-registered upload parked at `registered` and verify never ran. Restored to use the trigger filepath for `FindFile`/`RegisterFile` and to fall through to read + decrypt + archive in one pass, the same way a pre-registered `uploaded` file is handled. Full root cause in #2467.

This path is invisible to the existing SDA CI because every integration flow uploads via s3inbox (which pre-registers), so nothing enters the catch-all. Added a regression test that drives status `""` through to `archived`.

## Verification

`go build`/`vet`/`test` green, including unit tests for `ResolveInboxPath`, a `LoadInboxProjectConfig` seam test (nested `storage.inbox.*` YAML to the dotted keys it reads), validation tests (half-configured keys, delimiter allowlist, project-code denylist, error propagation out of `NewConfig`), and a new ingest regression test driving status `""` to registered then archived under the message correlation-id. End-to-end in the FEGA-Norway stack on the `v3.1.72` images: the full pipeline ran **green (9/9)** and a file walked `registered -> submitted -> archived -> verified -> ready`, exercising the projectCode reconstruction, the catch-all fix, and the mapper inbox cleanup. Design rationale + stock-behavior trace: ELIXIR-NO/FEGA-Norway#820.

## Commits (kept separate so the regression fix is reviewable apart from the additive work)

- `feat(sda): config-driven projectCode inbox-path reconstruction`, the additive feature
- `fix(helper): harden ResolveInboxPath for leading separator and segment boundary`, leading `/`, path-segment boundary
- `fix(ingest): register and archive non-s3inbox uploads in one pass`, the catch-all regression fix (#2467)
- `refactor: rename InboxConfig to InboxProjectConfig per review`

The remaining commits are review follow-ups: hardening tests, startup validation of the new config keys, and doc/typo fixes. See the commit list for the full set.

---

Fixes #2429
Fixes #2467

### DIFF
--- a/sda/cmd/ingest/ingest.go
+++ b/sda/cmd/ingest/ingest.go
@@ -70,7 +70,10 @@ func run() error {
 	if err = configv2.Load(); err != nil {
 		return fmt.Errorf("failed to load config: %v", err)
 	}
-	app.InboxProjectConfig = config.LoadInboxProjectConfig()
+	app.InboxProjectConfig, err = config.LoadInboxProjectConfig()
+	if err != nil {
+		return fmt.Errorf("failed to load inbox project config: %v", err)
+	}
 
 	app.Broker, err = rabbitmq.NewRabbitMQBroker(context.Background())
 	if err != nil {
--- a/sda/cmd/ingest/ingest.go
+++ b/sda/cmd/ingest/ingest.go
@@ -43,6 +43,7 @@ type Ingest struct {
 	ArchiveKeyList []*[32]byte
 	db             database.Database
 	InboxReader    storage.Reader
+	InboxConfig    helper.InboxConfig
 	Broker         brokerv2.Broker
 }
 
@@ -69,6 +70,7 @@ func run() error {
 	if err = configv2.Load(); err != nil {
 		return fmt.Errorf("failed to load config: %v", err)
 	}
+	app.InboxConfig = config.LoadInboxConfig()
 
 	app.Broker, err = rabbitmq.NewRabbitMQBroker(context.Background())
 	if err != nil {
@@ -297,7 +299,8 @@ func (app *Ingest) ingestFile(ctx context.Context, fileID, filePath, user, archi
 		// Catch all for implementations inbox uploading that does not register the file in the DB, e.g. for those not using S3inbox or sftpInbox
 		// Since we dont have the submission location in storage, we need to look through all configured storage locations.
 		var findFileErr, registerErr error
-		submissionLocation, findFileErr = app.InboxReader.FindFile(ctx, message.Key)
+		// Resolve the anonymized submission path to its physical inbox path before locating it.
+		submissionLocation, findFileErr = app.InboxReader.FindFile(ctx, helper.ResolveInboxPath(message.Key, user, app.InboxConfig))
 
 		// Ideally this transaction should span the whole message processing, but for now just spans the RegisterFile
 		tx, err := app.db.BeginTransaction(ctx)
@@ -336,7 +339,7 @@ func (app *Ingest) ingestFile(ctx context.Context, fileID, filePath, user, archi
 		return nil, fmt.Errorf("cannot ingest file with status: %s", status)
 	}
 
-	sourceReader, err := app.InboxReader.NewFileReader(ctx, submissionLocation, helper.UnanonymizeFilepath(filePath, user))
+	sourceReader, err := app.InboxReader.NewFileReader(ctx, submissionLocation, helper.ResolveInboxPath(filePath, user, app.InboxConfig))
 	if err != nil {
 		log.Errorf("failed to read file, due to: %v", err)
 
--- a/sda/cmd/ingest/ingest.go
+++ b/sda/cmd/ingest/ingest.go
@@ -37,14 +37,14 @@ import (
 )
 
 type Ingest struct {
-	ArchiveWriter  storage.Writer
-	BackupWriter   storage.Writer
-	ArchiveReader  storage.Reader
-	ArchiveKeyList []*[32]byte
-	db             database.Database
-	InboxReader    storage.Reader
-	InboxConfig    helper.InboxConfig
-	Broker         brokerv2.Broker
+	ArchiveWriter      storage.Writer
+	BackupWriter       storage.Writer
+	ArchiveReader      storage.Reader
+	ArchiveKeyList     []*[32]byte
+	db                 database.Database
+	InboxReader        storage.Reader
+	InboxProjectConfig helper.InboxProjectConfig
+	Broker             brokerv2.Broker
 }
 
 type decryptResult struct {
@@ -70,7 +70,7 @@ func run() error {
 	if err = configv2.Load(); err != nil {
 		return fmt.Errorf("failed to load config: %v", err)
 	}
-	app.InboxConfig = config.LoadInboxConfig()
+	app.InboxProjectConfig = config.LoadInboxProjectConfig()
 
 	app.Broker, err = rabbitmq.NewRabbitMQBroker(context.Background())
 	if err != nil {
@@ -301,7 +301,7 @@ func (app *Ingest) ingestFile(ctx context.Context, fileID, filePath, user, archi
 		var findFileErr, registerErr error
 		// message.Key is the broker correlation-id, not the submission path; use the trigger's
 		// filePath and resolve it to the physical inbox path before locating it.
-		submissionLocation, findFileErr = app.InboxReader.FindFile(ctx, helper.ResolveInboxPath(filePath, user, app.InboxConfig))
+		submissionLocation, findFileErr = app.InboxReader.FindFile(ctx, helper.ResolveInboxPath(filePath, user, app.InboxProjectConfig))
 
 		// Ideally this transaction should span the whole message processing, but for now just spans the RegisterFile
 		tx, err := app.db.BeginTransaction(ctx)
@@ -343,7 +343,7 @@ func (app *Ingest) ingestFile(ctx context.Context, fileID, filePath, user, archi
 		return nil, fmt.Errorf("cannot ingest file with status: %s", status)
 	}
 
-	sourceReader, err := app.InboxReader.NewFileReader(ctx, submissionLocation, helper.ResolveInboxPath(filePath, user, app.InboxConfig))
+	sourceReader, err := app.InboxReader.NewFileReader(ctx, submissionLocation, helper.ResolveInboxPath(filePath, user, app.InboxProjectConfig))
 	if err != nil {
 		log.Errorf("failed to read file, due to: %v", err)
 
--- a/sda/cmd/ingest/ingest.go
+++ b/sda/cmd/ingest/ingest.go
@@ -341,7 +341,7 @@ func (app *Ingest) ingestFile(ctx context.Context, fileID, filePath, user, archi
 		// upload stuck at "registered", so verify never runs.
 
 	default:
-		log.Warnf("file: %s recieved ingestion trigger with status: %s", fileID, status)
+		log.Warnf("file: %s received ingestion trigger with status: %s", fileID, status)
 
 		return nil, fmt.Errorf("cannot ingest file with status: %s", status)
 	}
--- a/sda/cmd/ingest/ingest.go
+++ b/sda/cmd/ingest/ingest.go
@@ -299,8 +299,9 @@ func (app *Ingest) ingestFile(ctx context.Context, fileID, filePath, user, archi
 		// Catch all for implementations inbox uploading that does not register the file in the DB, e.g. for those not using S3inbox or sftpInbox
 		// Since we dont have the submission location in storage, we need to look through all configured storage locations.
 		var findFileErr, registerErr error
-		// Resolve the anonymized submission path to its physical inbox path before locating it.
-		submissionLocation, findFileErr = app.InboxReader.FindFile(ctx, helper.ResolveInboxPath(message.Key, user, app.InboxConfig))
+		// message.Key is the broker correlation-id, not the submission path; use the trigger's
+		// filePath and resolve it to the physical inbox path before locating it.
+		submissionLocation, findFileErr = app.InboxReader.FindFile(ctx, helper.ResolveInboxPath(filePath, user, app.InboxConfig))
 
 		// Ideally this transaction should span the whole message processing, but for now just spans the RegisterFile
 		tx, err := app.db.BeginTransaction(ctx)
@@ -311,7 +312,8 @@ func (app *Ingest) ingestFile(ctx context.Context, fileID, filePath, user, archi
 		}
 
 		// Register file even if FindFile didnt succeed with submissionLocation == "", as we will add an error file event log to it in that case
-		fileID, registerErr = tx.RegisterFile(ctx, &fileID, submissionLocation, message.Key, user)
+		// Store the anonymized submission path (filePath), not the correlation-id, so the mapper can resolve it back on cleanup.
+		fileID, registerErr = tx.RegisterFile(ctx, &fileID, submissionLocation, filePath, user)
 		if registerErr != nil {
 			log.Errorf("failed to register file, fileID: %s, reason: (%s)", fileID, registerErr.Error())
 			if err := tx.Rollback(); err != nil {
@@ -331,7 +333,9 @@ func (app *Ingest) ingestFile(ctx context.Context, fileID, filePath, user, archi
 			return []func(){app.errorQueue(message), app.setErrorEvent(findFileErr.Error(), message)}, nil
 		}
 
-		return nil, nil
+		// File is now registered; fall through to read + decrypt + archive in a single pass, the same
+		// way a pre-registered "uploaded" file is handled. Returning here would leave a non-s3inbox
+		// upload stuck at "registered", so verify never runs.
 
 	default:
 		log.Warnf("file: %s recieved ingestion trigger with status: %s", fileID, status)
--- a/sda/cmd/ingest/ingest_test.go
+++ b/sda/cmd/ingest/ingest_test.go
@@ -432,7 +432,7 @@ func (ts *TestSuite) TestIngestFile_NotRegistered_FallsThroughToArchive() {
 
 	fileID, err := ts.ingest.db.GetFileIDByUserPathAndStatus(context.Background(), ts.UserName, ts.filePath, "archived")
 	assert.NoError(ts.T(), err, "non-registered file should reach 'archived' in one pass, not park at 'registered'")
-	assert.NotEmpty(ts.T(), fileID, "expected an archived file for the non-registered upload")
+	assert.Equal(ts.T(), correlationID, fileID, "catch-all must register the file under the message correlation-id, finalize looks the file up by it")
 }
 
 func (ts *TestSuite) TestIngestFile_NoSubmissionLocation() {
--- a/sda/cmd/ingest/ingest_test.go
+++ b/sda/cmd/ingest/ingest_test.go
@@ -420,6 +420,21 @@ func (ts *TestSuite) TestIngestFile_BaseCase() {
 	assert.NoError(ts.T(), err, "unexpected error when ingesting file")
 }
 
+func (ts *TestSuite) TestIngestFile_NotRegistered_FallsThroughToArchive() {
+	// Non-s3inbox flow (e.g. FEGA-Norway via TSD): nothing pre-registers the file, so ingest's
+	// catch-all (status "") is the first registration point. It must register AND archive in one
+	// pass; an early return after RegisterFile would park the file at "registered" and verify would
+	// never run. Guards the v3.1.72 catch-all regression.
+	correlationID := uuid.New().String()
+	message := createMessage("ingest", ts.filePath, ts.UserName, correlationID)
+	_, err := ts.ingest.handleMessage(context.Background(), message)
+	assert.NoError(ts.T(), err, "unexpected error ingesting a non-registered file")
+
+	fileID, err := ts.ingest.db.GetFileIDByUserPathAndStatus(context.Background(), ts.UserName, ts.filePath, "archived")
+	assert.NoError(ts.T(), err, "non-registered file should reach 'archived' in one pass, not park at 'registered'")
+	assert.NotEmpty(ts.T(), fileID, "expected an archived file for the non-registered upload")
+}
+
 func (ts *TestSuite) TestIngestFile_NoSubmissionLocation() {
 	fileID, err := ts.ingest.db.RegisterFile(context.Background(), nil, "/inbox", ts.filePath, ts.UserName)
 	assert.NoError(ts.T(), err, "failed to register file in database")
--- a/sda/internal/config/config.go
+++ b/sda/internal/config/config.go
@@ -7,7 +7,6 @@ import (
 	"os"
 	"strings"
 	"time"
-	"unicode"
 
 	"github.com/neicnordic/crypt4gh/keys"
 	"github.com/neicnordic/sensitive-data-archive/internal/broker"
@@ -783,9 +782,9 @@ func (c *Config) configReEncryptServer() (err error) {
 // An absent section yields the zero value, which is stock SDA behavior, so deployments that omit it
 // are unaffected. This is the single source of the storage.inbox.* keys, read both by NewConfig
 // (mapper, api) and directly by the ingest service, which loads through config/v2 but populates
-// the same viper instance. Setting only one of the two keys is an error, as is a delimiter that is
-// not a single separator character: it joins code and username into one directory name, so letters,
-// digits and "/" would corrupt the layout.
+// the same viper instance. Setting only one of the two keys is an error. The delimiter joins code
+// and username into one inbox directory name, so only "-", "_" and "." are accepted: anything else
+// (letters, digits, "/", whitespace) would corrupt or split the layout.
 func LoadInboxProjectConfig() (helper.InboxProjectConfig, error) {
 	cfg := helper.InboxProjectConfig{
 		Code:      viper.GetString("storage.inbox.projectCode"),
@@ -798,9 +797,10 @@ func LoadInboxProjectConfig() (helper.InboxProjectConfig, error) {
 		return helper.InboxProjectConfig{}, errors.New("storage.inbox.projectCode and storage.inbox.projectCodeDelimiter must be set together")
 	}
 
-	r := []rune(cfg.Delimiter)
-	if len(r) != 1 || unicode.IsLetter(r[0]) || unicode.IsDigit(r[0]) || r[0] == '/' {
-		return helper.InboxProjectConfig{}, fmt.Errorf("storage.inbox.projectCodeDelimiter %q is not a delimiter character", cfg.Delimiter)
+	switch cfg.Delimiter {
+	case "-", "_", ".":
+	default:
+		return helper.InboxProjectConfig{}, fmt.Errorf("storage.inbox.projectCodeDelimiter %q is not a delimiter character (use \"-\", \"_\" or \".\")", cfg.Delimiter)
 	}
 
 	return cfg, nil
--- a/sda/internal/config/config.go
+++ b/sda/internal/config/config.go
@@ -41,7 +41,7 @@ type Config struct {
 	ReEncrypt    ReEncConfig
 	Auth         AuthConf
 	RotateKey    RotateKeyConf
-	Inbox        helper.InboxConfig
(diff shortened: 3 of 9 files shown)


sda/cmd/ingest/ingest.go
```diff
@@ -297,7 +299,8 @@ func (app *Ingest) ingestFile(ctx context.Context, fileID, filePath, user, archi
 		// Catch all for implementations inbox uploading that does not register the file in the DB, e.g. for those not using S3inbox or sftpInbox
 		// Since we dont have the submission location in storage, we need to look through all configured storage locations.
 		var findFileErr, registerErr error
-		submissionLocation, findFileErr = app.InboxReader.FindFile(ctx, message.Key)
+		// Resolve the anonymized submission path to its physical inbox path before locating it.
+		submissionLocation, findFileErr = app.InboxReader.FindFile(ctx, helper.ResolveInboxPath(message.Key, user, app.InboxConfig))
```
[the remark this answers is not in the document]
yasinmiran: good catch, this is exactly the regression we're fixing. v3.1.61's catch-all used the body filepath here and fell through to read+archive in one pass; v3.1.72 switched it to message.Key (the correlation-id) and added an early return after register, so for non-registered uploads (our TSD flow, which is what hits this fallback) the file can't be located and never archives. we're restoring the v3.1.61 behaviour (body filepath + fall-through). leaving this as-is until that fix lands.

sda/internal/helper/helper.go
```diff
@@ -464,3 +464,39 @@ func UnanonymizeFilepath(fp string, username string) string {
...
+	if cfg.ProjectCode == "" {
+		return UnanonymizeFilepath(filePath, username)
+	}
+
+	userDir := cfg.ProjectCode + cfg.ProjectCodeDelimiter + username
+	if strings.HasPrefix(filePath, userDir) {
+		return filePath
+	}
+
+	return filepath.Join(userDir, filePath)
```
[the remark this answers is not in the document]
yasinmiran: agreed on both points. the leading-slash case is the old-proxy format (/p11-user/files/x) that older proxies can still send into the v2 stack, which currently doubles the prefix; and the missing path-segment boundary would mis-handle a prefix-overlapping user (p11-user2 treated as already-resolved for userDir p11-user). both are on the list to harden ResolveInboxPath (strip a leading separator + enforce a segment boundary, with tests). leaving as-is until the fix lands.

sda/internal/config/config.go
```diff
@@ -768,6 +772,24 @@ func (c *Config) configReEncryptServer() (err error) {
...
+func LoadInboxConfig() helper.InboxConfig {
+	return helper.InboxConfig{
+		ProjectCode:          viper.GetString("storage.inbox.projectCode"),
+		ProjectCodeDelimiter: viper.GetString("storage.inbox.projectCodeDelimiter"),
+	}
+}
+
+// configInbox populates the inbox layout config for services that resolve anonymized submission
+// paths back to their physical inbox path (mapper, api).
+func (c *Config) configInbox() {
```
jbygdell:
```suggestion
func (c *Config) configInboxProject() {
```

Keep the names clear on what the function does so it isn't easy to miss interpret.
You are not configuring the inbox, just a specific part.

sda/internal/helper/helper.go
```diff
@@ -464,3 +464,39 @@ func UnanonymizeFilepath(fp string, username string) string {
...
+// anonymized submission path (the user prefix stripped) can be resolved back to the physical
+// inbox-relative path. An empty ProjectCode selects stock SDA behavior.
+type InboxConfig struct {
+	// ProjectCode optionally prefixes the per-user inbox directory (e.g. "p11"). Empty keeps the
+	// stock layout where the directory is the normalized username (see UnanonymizeFilepath).
+	ProjectCode string
+	// ProjectCodeDelimiter separates ProjectCode from the username (e.g. "-"). Unused when
+	// ProjectCode is empty.
+	ProjectCodeDelimiter string
+}
```
jbygdell:
```suggestion
type InboxProjectConfig struct {
	// Code optionally prefixes the per-user inbox directory (e.g. "p11"). Empty keeps the
	// stock layout where the directory is the normalized username (see UnanonymizeFilepath).
	Code string
	// ProjectCodeDelimiter separates ProjectCode from the username (e.g. "-"). Unused when
	// ProjectCode is empty.
	Delimiter string
}
```
The name InboxConfig implies that this is the configuration for the inbox when it in reality is a part of the configuration.

jbygdell (commented):
Overall good, the names could have been chosen a bit better som that it is clear what the intent are. As it is is it too easy to misinterpret for a general config.

sda/cmd/ingest/ingest.go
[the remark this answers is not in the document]
yasinmiran: fixed in 8eb424f4: the catch-all now uses the trigger filepath (not message.Key) for FindFile + RegisterFile, and falls through to read + decrypt + archive in one pass. tracked as #2467 with a regression test that drives status "" through to archived.

sda/internal/helper/helper.go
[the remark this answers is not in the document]
yasinmiran: fixed in 581a0590: strip a leading separator before the already-resolved check, and require a path-segment boundary so "p11-user2" isn't treated as already under "p11-user". tests added in helper_test.go.

sda/internal/config/config.go
yasinmiran: done in 8d177fed, renamed to configInboxProject. thanks for the review!

sda/internal/helper/helper.go
yasinmiran: done in 8d177fed: now InboxProjectConfig with Code/Delimiter (and LoadInboxProjectConfig to match). agreed the old name overstated the scope. thanks for the review!

sda/internal/helper/helper.go
```diff
@@ -464,3 +464,45 @@ func UnanonymizeFilepath(fp string, username string) string {
...
+	// stripping it the prefix check below misses and userDir gets prepended a second time.
+	relPath := strings.TrimPrefix(filePath, "/")
+	// Treat as already-resolved only on a path-segment boundary, so "p11-user2/..." is not mistaken
+	// for the "p11-user" directory.
+	if relPath == userDir || strings.HasPrefix(relPath, userDir+"/") {
+		return relPath
+	}
+
+	return filepath.Join(userDir, relPath)
+}
```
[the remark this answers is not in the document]
yasinmiran: @jbygdell wdyt about this one? I mean Copilot has a point.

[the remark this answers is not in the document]
jbygdell: Except that Copilot doesn't have the full picture.

`filePath` is not used as it is, `filePath.Join(storageLocation, filePath)` will generate the same final path regardless if the `filePath` variable has a leading `/` or not.

But to be sure that the code works as intended we can always test it.
* unit test for `UnanonymizeFilepath` when the filePath has a leading `/`
* integration test for the cases where the file path has a project prefix.

sda/internal/helper/helper.go
```diff
@@ -464,3 +464,45 @@ func UnanonymizeFilepath(fp string, username string) string {
...
+		return UnanonymizeFilepath(filePath, username)
+	}
+
+	userDir := cfg.Code + cfg.Delimiter + username
+	// Tolerate a leading separator from older proxy formats (e.g. "/p11-user/files/..."); without
+	// stripping it the prefix check below misses and userDir gets prepended a second time.
+	relPath := strings.TrimPrefix(filePath, "/")
+	// Treat as already-resolved only on a path-segment boundary, so "p11-user2/..." is not mistaken
+	// for the "p11-user" directory.
+	if relPath == userDir || strings.HasPrefix(relPath, userDir+"/") {
```
jbygdell: `relPath == userDir` this should never be the case. since it would mean that the filePath doesn't have a file component.

yasinmiran: good catch, that arm only fires for a bare `userDir` with no file under it, and a submission path always has a file component, so it's unreachable. dropped it, line 503 is now just the segment-boundary check. thanks for the review!

sda/internal/helper/helper.go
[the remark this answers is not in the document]
yasinmiran: agreed, and confirmed it: Go's `filepath.Join` concatenates then `Clean`s, it doesn't treat a leading `/` as an absolute path the way POSIX `join` does, so the userDir prefix is never dropped. e.g. `UnanonymizeFilepath("/files/x", user)` returns `user/files/x`. added a unit test for `UnanonymizeFilepath` with a leading `/` (and one for the stock `ResolveInboxPath` branch); the project-prefix path is already covered end-to-end by the FEGA e2e (projectCode `p11`, `/files/x` resolved through to `verified`). thanks for the review!

sda/cmd/ingest/ingest_test.go
```diff
@@ -420,6 +420,21 @@ func (ts *TestSuite) TestIngestFile_BaseCase() {
...
+	// Non-s3inbox flow (e.g. FEGA-Norway via TSD): nothing pre-registers the file, so ingest's
+	// catch-all (status "") is the first registration point. It must register AND archive in one
+	// pass; an early return after RegisterFile would park the file at "registered" and verify would
+	// never run. Guards the v3.1.72 catch-all regression.
+	correlationID := uuid.New().String()
+	message := createMessage("ingest", ts.filePath, ts.UserName, correlationID)
+	_, err := ts.ingest.handleMessage(context.Background(), message)
+	assert.NoError(ts.T(), err, "unexpected error ingesting a non-registered file")
+
+	fileID, err := ts.ingest.db.GetFileIDByUserPathAndStatus(context.Background(), ts.UserName, ts.filePath, "archived")
```
jbygdell, line 433: You need to use the correlationID generated in this test and validate that the state is the correct one.

sda/cmd/ingest/ingest.go
```diff
@@ -308,7 +312,8 @@ func (app *Ingest) ingestFile(ctx context.Context, fileID, filePath, user, archi
 		}
 
 		// Register file even if FindFile didnt succeed with submissionLocation == "", as we will add an error file event log to it in that case
-		fileID, registerErr = tx.RegisterFile(ctx, &fileID, submissionLocation, message.Key, user)
+		// Store the anonymized submission path (filePath), not the correlation-id, so the mapper can resolve it back on cleanup.
+		fileID, registerErr = tx.RegisterFile(ctx, &fileID, submissionLocation, filePath, user)
```
jbygdell, line 319:
```suggestion
		fileID, registerErr = tx.RegisterFile(ctx, &fileID, submissionLocation, message.Key, user)
```

This DB query needs the `message.Key` i.e the correlation ID in order to work.

sda/internal/config/config.go
```diff
@@ -768,6 +772,24 @@ func (c *Config) configReEncryptServer() (err error) {
...
+// empty (absent) project code yields stock SDA behavior, so deployments that omit the section are
+// unaffected. This is the single source of the storage.inbox.* keys, read both by NewConfig
+// (mapper, api) and directly by the ingest service, which loads through config/v2 but populates
+// the same viper instance.
+func LoadInboxProjectConfig() helper.InboxProjectConfig {
+	return helper.InboxProjectConfig{
+		Code:      viper.GetString("storage.inbox.projectCode"),
+		Delimiter: viper.GetString("storage.inbox.projectCodeDelimiter"),
+	}
+}
```
jbygdell: Technically this isn't needed since the struct could be returned directly in the function below.

sda/internal/config/config.go
```diff
@@ -768,6 +772,24 @@ func (c *Config) configReEncryptServer() (err error) {
...
+
+// configInboxProject populates the inbox layout config for services that resolve anonymized submission
+// paths back to their physical inbox path (mapper, api).
+func (c *Config) configInboxProject() {
+	c.Inbox = LoadInboxProjectConfig()
```
jbygdell:
```suggestion
	c.Inbox = helper.InboxProjectConfig{
		Code:      viper.GetString("storage.inbox.projectCode"),
		Delimiter: viper.GetString("storage.inbox.projectCodeDelimiter"),
	}
```

This is a completely viable solution as well.

jbygdell (dismissed):
An integration test for this use case is needed to ensure that we maintain functionality going forward.

sda/cmd/ingest/ingest.go
yasinmiran: thanks for taking another look!

the requirement is right, but it's already satisfied through a different parameter. `fileID` here *is* `message.Key`: it's bound at the call site ([ingest.go#L185](https://github.com/neicnordic/sensitive-data-archive/blob/f80325b6b8d78d3848ba90f54021ce9e2b0e0b36/sda/cmd/ingest/ingest.go#L185)) and goes into [`registerFileQuery` as `$1`](https://github.com/neicnordic/sensitive-data-archive/blob/f80325b6b8d78d3848ba90f54021ce9e2b0e0b36/sda/internal/database/postgres/method_register_file.go#L11-L20), so the row is registered with `id = <correlation id>`. that's the binding finalize relies on ([finalize.go#L149](https://github.com/neicnordic/sensitive-data-archive/blob/f80325b6b8d78d3848ba90f54021ce9e2b0e0b36/sda/cmd/finalize/finalize.go#L149)).

the 4th argument is `uploadPath` and lands in `submission_file_path` (`$3`). putting `message.Key` there writes the uuid as the file's path, which breaks the consumers of that column:

- [mapper's inbox cleanup](https://github.com/neicnordic/sensitive-data-archive/blob/f80325b6b8d78d3848ba90f54021ce9e2b0e0b36/sda/cmd/mapper/mapper.go#L268-L272) resolves `submission_file_path` to delete the inbox file after mapping; with a uuid there it tries to remove a file that doesn't exist, and the failure is log-only so it's silent
- `GetFileIDByUserPathAndStatus` finds nothing by path (the recovery lookup [ADR 0003](https://github.com/neicnordic/sensitive-data-archive/blob/f80325b6b8d78d3848ba90f54021ce9e2b0e0b36/docs/decisions/0003-shared-state-strategy-for-s3inbox-and-caching.md) standardizes on)
- the `ON CONFLICT` upsert keys on [`unique_ingested`](https://github.com/neicnordic/sensitive-data-archive/blob/f80325b6b8d78d3848ba90f54021ce9e2b0e0b36/postgresql/initdb.d/01_main.sql#L114) which includes `submission_file_path`; every redelivery carries a fresh correlation id, so retried uploads would stop converging onto the existing row and insert a new one each time

it also mirrors what s3inbox writes through the same function (path in that position, anonymized at [L220](https://github.com/neicnordic/sensitive-data-archive/blob/f80325b6b8d78d3848ba90f54021ce9e2b0e0b36/sda/cmd/s3inbox/proxy.go#L220); the correlation id is exported separately as the published message's correlation):

https://github.com/neicnordic/sensitive-data-archive/blob/f80325b6b8d78d3848ba90f54021ce9e2b0e0b36/sda/cmd/s3inbox/proxy.go#L244

the catch-all is the same registration deferred to ingest time for inboxes that can't pre-register, so it should write the same thing.

as far as i can tell `message.Key` came in as a mechanical slip in the broker v2 migration ([b5cf971c](https://github.com/neicnordic/sensitive-data-archive/commit/b5cf971c2ff5540b87178a60dda8d4f7dc86b03d)): `message` changed type from the parsed trigger struct (which had `.FilePath`) to `brokerv2.Message` (which only has `.Key`), and both call sites here got `message.Key` while `message.User` correctly became the new `user` param. [v3.1.61 had `message.FilePath` in both spots](https://github.com/neicnordic/sensitive-data-archive/blob/v3.1.61/sda/cmd/ingest/ingest.go#L323-L330). that's the regression #2467 describes.

the unit test from your other comment now covers both halves: it asserts the registered id equals the message correlation id *and* finds the file by its path; with `message.Key` in the path slot it fails with `sql: no rows in result set`.

sda/cmd/ingest/ingest_test.go
yasinmiran: good catch, the path-based lookup alone wouldn't notice if registration minted a random uuid instead of adopting the correlation id. added `assert.Equal(correlationID, fileID)` in cac77cba, so the test now pins the id binding finalize depends on, and the existing path lookup keeps pinning `submission_file_path`. together they also guard the other thread: with `message.Key` passed as the upload path the test fails with `sql: no rows in result set`.

sda/internal/config/config.go
yasinmiran: i'd rather keep it: ingest loads through config/v2 and never builds a `Config`, so it reads these keys via `config.LoadInboxProjectConfig()` directly ([ingest.go#L73](https://github.com/neicnordic/sensitive-data-archive/blob/f80325b6b8d78d3848ba90f54021ce9e2b0e0b36/sda/cmd/ingest/ingest.go#L73)). inlining the struct into `configInboxProject` would mean the same two viper reads exist twice (here and in ingest) and have to be kept in sync by hand.

that said, if the extra function really bothers you i'm happy to refactor it differently, open to suggestions.

sda/internal/config/config.go
yasinmiran: covered in the comment on `LoadInboxProjectConfig` above: ingest (config/v2) calls it directly, so it needs to stay callable outside `Config`.

sda/internal/helper/helper.go
```diff
@@ -464,3 +464,46 @@ func UnanonymizeFilepath(fp string, username string) string {
...
+// deployment needs a project code together with normalization; add an explicit toggle if one does.
+func ResolveInboxPath(filePath, username string, cfg InboxProjectConfig) string {
+	if cfg.Code == "" {
+		return UnanonymizeFilepath(filePath, username)
+	}
+
+	userDir := cfg.Code + cfg.Delimiter + username
+	// Tolerate a leading separator from older proxy formats (e.g. "/p11-user/files/..."); without
+	// stripping it the prefix check below misses and userDir gets prepended a second time.
+	relPath := strings.TrimPrefix(filePath, "/")
```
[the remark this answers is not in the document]
yasinmiran: right, `TrimPrefix` only strips one. switched to `strings.TrimLeft(filePath, "/")` and extended the leading-separator test with the `//`-prefixed already-resolved case in ef11cadf.

sda/internal/helper/helper.go
```diff
@@ -464,3 +464,46 @@ func UnanonymizeFilepath(fp string, username string) string {
...
+	relPath := strings.TrimLeft(filePath, "/")
+	// Treat as already-resolved only on a path-segment boundary, so "p11-user2/..." is not mistaken
+	// for the "p11-user" directory. A submission path always has a file component, so relPath is
+	// never the bare userDir.
+	if strings.HasPrefix(relPath, userDir+"/") {
+		return relPath
+	}
+
+	return filepath.Join(userDir, relPath)
+}
```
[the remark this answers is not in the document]
yasinmiran: same ground as https://github.com/neicnordic/sensitive-data-archive/pull/2466#discussion_r3381729360: the `cfg.Code == ""` branch intentionally defers to stock `UnanonymizeFilepath` unchanged, so deployments without a project code keep byte-identical stock behavior; normalizing before that branch would change stock semantics, which is out of scope for this PR. the leading-slash behavior of the stock branch is pinned by `TestUnanonymizeFilepath_leadingSeparator` and `TestResolveInboxPath_stockDefault_leadingSeparator`, and no producer emits an already-resolved path with a leading separator in the stock layout.

sda/internal/config/config.go
jbygdell: That's what you get for being out of sync with the latest changes (me that is).

Having taken a closer look at this I think it will be better to remove the `configInboxProject()` below. Since it can be used in all places where this is needed.
And instead expand this function to to do basic validation i.e. that the `delimiter` is an actual delimiter character and not an alphabetic character or a number.
The function should also return an error if only one of the values are set.

yasinmiran: thanks for taking a closer look, agreed this is a better shape 🙏 done in 1a8d3229 + 55dec521 + ba5fadef: dropped `configInboxProject()`, both NewConfig call sites + ingest call `LoadInboxProjectConfig()` directly. the loader now errors when only one of the two keys is set, the delimiter is validated against an allowlist (`-`, `_`, `.`) rather than just not-letter/not-digit (a denylist still admitted whitespace, control chars and `/`, which would corrupt the inbox directory layout), and per a follow-up copilot comment the project code itself is also rejected if it contains path separators, whitespace or control chars. easy to widen the delimiter set if a node ever needs another separator. tests cover the half-set + bad-delimiter + bad-code cases and that NewConfig fails startup on a misconfig.

sda/internal/helper/helper.go
```diff
@@ -464,3 +464,46 @@ func UnanonymizeFilepath(fp string, username string) string {
...
+	// stock layout where the directory is the normalized username (see UnanonymizeFilepath).
+	Code string
+	// Delimiter separates Code from the username (e.g. "-"). Unused when
+	// Code is empty.
+	Delimiter string
+}
+
+// ResolveInboxPath reconstructs the physical inbox-relative path for an anonymized submission
+// filePath. An already-resolved path (e.g. on reprocessing) is returned as-is, with any leading
+// separator normalized away.
```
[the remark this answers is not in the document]
yasinmiran: fair, the comment claimed it for both branches. reworded in 19e2069e: the already-resolved + leading-separator guarantees are now stated under the projectCode branch only, and the stock paragraph says explicitly that it defers to `UnanonymizeFilepath` unchanged (that asymmetry is deliberate, stock behavior stays byte-identical).

sda/cmd/ingest/ingest.go
```diff
@@ -328,15 +336,17 @@ func (app *Ingest) ingestFile(ctx context.Context, fileID, filePath, user, archi
 			return []func(){app.errorQueue(message), app.setErrorEvent(findFileErr.Error(), message)}, nil
 		}
 
-		return nil, nil
+		// File is now registered; fall through to read + decrypt + archive in a single pass, the same
+		// way a pre-registered "uploaded" file is handled. Returning here would leave a non-s3inbox
+		// upload stuck at "registered", so verify never runs.
 
 	default:
 		log.Warnf("file: %s recieved ingestion trigger with status: %s", fileID, status)
```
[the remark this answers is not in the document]
yasinmiran: fixed in 4979356f. fyi the same typo exists at ingest.go:142 ("recieved signal"), also from b5cf971c, but that line is outside this PR's diff so i left it alone.

sda/internal/config/config.go
```diff
@@ -768,6 +778,34 @@ func (c *Config) configReEncryptServer() (err error) {
...
+		return helper.InboxProjectConfig{}, errors.New("storage.inbox.projectCode and storage.inbox.projectCodeDelimiter must be set together")
+	}
+
+	switch cfg.Delimiter {
+	case "-", "_", ".":
+	default:
+		return helper.InboxProjectConfig{}, fmt.Errorf("storage.inbox.projectCodeDelimiter %q is not a delimiter character (use \"-\", \"_\" or \".\")", cfg.Delimiter)
+	}
+
+	return cfg, nil
```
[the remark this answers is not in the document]
yasinmiran: good catch, symmetric gap with the delimiter check. done in ba5fadef: the project code is now rejected at startup if it contains `/`, `\`, whitespace or control chars, since it prefixes a single per-user directory name and any of those silently splits or corrupts the path. kept it a denylist (vs the delimiter's allowlist) because codes are deployment-specific free text. tests cover both sides.

sda/internal/helper/helper.go
```diff
@@ -464,3 +464,48 @@ func UnanonymizeFilepath(fp string, username string) string {
 
 	return filepath.Join(strings.Replace(username, "@", "_", 1), fp)
 }
+
+// InboxProjectConfig describes how a deployment names its per-user inbox directories, so an
+// anonymized submission path (the user prefix stripped) can be resolved back to the physical
+// inbox-relative path. An empty Code selects stock SDA behavior.
+type InboxProjectConfig struct {
```
KarlG-nbis, line 471: I would put this in its own package and utilse config/v2 package for the configuration loading, as then it will be loaded "automatically" when `configv2.Load()` is called, and the applications that need it would register those flags by importing that package.

So something as

sda/internal/v2/inbox_project_config/inbox_project_config.go

```go
package inbox_project_config

...

var (
  delimiter, code string
)


func init() {
	config.RegisterFlags(
		&config.Flag{
			Name: "storage.inbox.project_code",
				flagSet.String(flagName, "", "Some good description")
			},
			Required: false,
				code = viper.GetString(flagName)
			},
		},
		&config.Flag{
			Name: "storage.inbox.project_code_delimiter",
				flagSet.String(flagName, "", "Some good description")
			},
			Required: false,
				delimiter = viper.GetString(flagName)
			},
		},
	)
}


func ResolveInboxPath(filePath, username string) string {
    if code == "" {
		return helper.UnanonymizeFilepath(filePath, username)
	}

	userDir := code + delimiter + username

	relPath := strings.TrimLeft(filePath, "/")

	if strings.HasPrefix(relPath, userDir+"/") {
		return relPath
	}

	return filepath.Join(userDir, relPath)
}
```

As then when applications use `inbox_project_config.ResolveInboxPath` that package is imported and its init function included which registers those flags and loads them in

kjellp: Thanks @KarlG-nbis ! As agreed in the meeting, we'll create a second issue to update this config loading later.

KarlG-nbis (approved):

sda/internal/helper/helper.go
yasinmiran: tracked here https://github.com/neicnordic/sensitive-data-archive/issues/2477

kjellp (approved):
Also tested within TSD in our staging stack.

Issue #2459, opened 2026-06-01
Implement the broker v2 package 


viklund:
# Description

Included in [2381](https://github.com/neicnordic/sensitive-data-archive/pull/2381) there is a new broker package that enables each service to: requeue messages, run callbacks after ack/nack, ignore the ack/nack loop in the application package using the broker service. In the same change there is a example of how this can be used in the ingest service. The same should be implemented for the remaining services:

- [x] api
- [x] finalize
- [x] intercept
- [x] mapper
- [ ] orchestrate **(Excluded from this issue, tracked by #2602)**
- [x] rotatekey
- [x] s3inbox
- [x] sync
- [ ] syncapi **(Excluded from this issue, tracked by #2601)**
- [x] verify
- [x] All chart templates are up to date

# Nice to have:

Since we also have the database interface change merged in [2389](https://github.com/neicnordic/sensitive-data-archive/pull/2389) while going through each service, see if we can also get rid of the [legacy config package](https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/internal/config/config.go)

- [ ] Drop `global.broker.serverName` from the sda-svc chart once nothing reads it, with a CHANGELOG entry. broker/v2 has no `server_name` option and the v1 broker is the last consumer (`internal/config/config.go:592`). Both spellings are in the chart today, `serverName` for the services still on the v1 broker and a leftover `server_name` in finalize.

# A/C

- All services make use of the v2 broker package
- All chart templates carry the snake case semantics for broker related config options

jhagberg:
While reviewing the four open broker v2 PRs (#2551, #2563, #2571, #2579) I ran into three things that are not really about any one of them but about the feature branch as a whole. Rather than attach them to a single PR I am raising them here, and leaving it to you (@KarlG-nbis, @jbygdell) whether they deserve issues of their own, a fix on the branch, or nothing.

[...]

PR #2482, opened 2026-06-19
refactor(api): brokerv2, replace gin, unit-tests

zeidlitz:
partial commit for #2459 

included in this change are the following:

  - adding the brokerv2 package for improved message handling and interface support

  - adding a standalone config to replace legacy config package

  - replace use of gin to instead use net/http to reduce dependencies outside of the standard library

  - improve unit-tests to make use of database and broker interfaces

  - created a shared `/mocks` directory that implements interfaces so it can be reused and shared among other unit tests


sda/cmd/api/config/config.go
```diff
@@ -0,0 +1,253 @@
...
+			},
+		},
+		&config.Flag{
+			Name: "schema_type",
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.String(flagName, "isolated", "Path to JSON schemas to validate rabbitmq messages against")
+			},
+			Required: false,
+			AssignFunc: func(flagName string) {
+				schemaType := viper.GetString("schema_type")
```
nanjiangshu:
```suggestion
				schemaType := viper.GetString(flagName)
```

zeidlitz: fixed! :+1:

sda/cmd/api/config/config.go
```diff
@@ -0,0 +1,253 @@
...
+	clientCert    string
+	grpcHost      string
+	grpcPort      int
+	grpcTimeout   int
+)
+
+func init() {
+	config.RegisterFlags(
+		&config.Flag{
+			Name: "api_host",
```
nanjiangshu, line 36: currently, all fields are set to `required=False`, probably better to configure some of them as required.
e.g. 

`api_host`, `rbac_file`. 

Another problem is that all fields are using snake_case now, may it encounter parsing problem?

sda/cmd/api/api.go
```diff
@@ -86,1279 +70,114 @@ func run() error {
...
-		fileID, err = db.GetFileIDByUserPathAndStatus(c, accession.User, accession.FilePath, "verified")
-		if err != nil {
-			if fileID == "" {
-				c.AbortWithStatusJSON(http.StatusBadRequest, err.Error())
-			} else {
-				c.AbortWithStatusJSON(http.StatusInternalServerError, err.Error())
+	go func() {
+		slog.Info("starting", "addr", addr)
+		if apiconfig.SSL() {
+			if err := app.server.ListenAndServeTLS(apiconfig.ClientCert(), apiconfig.ClientKey()); err != nil && !errors.Is(err, http.ErrServerClosed) {
```
nanjiangshu, line 156: I might be misunderstanding the intent here, but this refactor seems to use `client_cert` and `client_key` for both the outbound gRPC client TLS config and the inbound HTTP server TLS config (`ListenAndServeTLS`). In the previous implementation, those are separate: the API server used `ServerCert/ServerKey`, while gRPC used `Conf.API.Grpc.ClientCreds`.

The refactoring may have changed the meaning of the config in a way that could  mix up client and server certificates. Could we keep separate config for API server TLS and gRPC client TLS, or otherwise clarify that the same cert/key pair is expected to be used for both?

sda/cmd/api/handlers.go
```diff
@@ -0,0 +1,1110 @@
...
+}
+
+func (api *API) readinessResponse(w http.ResponseWriter, r *http.Request) {
+	if !api.mq.Alive() {
+		w.WriteHeader(http.StatusInternalServerError)
+		w.Write([]byte("unable to reach rabbitmq"))
+	}
+
+	if err := api.db.Ping(context.TODO()); err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
```
nanjiangshu: I think `503 Service Unavailable` is more appropriate than `500`

sda/cmd/api/config/config.go
zeidlitz: The use of snake case is intentional since It's considered standard practice for environment variables. Parsing problems should not be a issue.

Any flag that needs a value should have a reasonable one set as default. For example `api_host = 0.0.0.0` and `rbac_file=/rbac.json`

sda/cmd/api/api.go
zeidlitz: The intent is that from the `api` services perspective, It only have one identity. It's the same identity it uses for serving http as it is using for communicating with the `reencrypt` service using grpc. I could also only find a single set of client certificate files, namely `tls.key`, `tls.crt` and `ca.crt` on the service. So as of right now we don't seem to provision any other certificates that could be used. 

If we are not actually separating these I find it much clearer to simply reference them as is. 

Otherwise we have 4 set of variables that all point to the same files,

i.e:

`grpc.client_cert` -> `tls.crt`

`grpc.client_key` -> `tls.key`

`grpc.ca_cert` -> `ca.crt`

`server.client_cert` -> `tls.crt`

`server.client_key` -> `tls.key`

`server.ca_cert` -> `ca.crt`


which is a bit confusing in my mind and causes a lot of extra overhead.

sda/cmd/api/handlers.go
zeidlitz: Agreed, will adjust

sda/cmd/api/config/config.go
nanjiangshu: sounds good.

sda/cmd/api/api.go
nanjiangshu: I see what you mean and I agree it might look messy to have four sets of variables all point to the same files. 
Maybe we can add a comment to the code to clarify that the same certs are used and that's why  `ClientCert()` and `ClientKey` are used for servers here.

Alternatively, we could implement a fallback logic so that different certs can be set for grpc if needed.
For example, but this is probably not a requirement. 
```golang
func ClientCert() string {
    if grpcClientCert != "" {
        return grpcClientCert
    }
    return clientCert // Fallback
}
```

sda/CHANGELOG.md
```diff
@@ -7,6 +7,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [3.1.74] - 2026-06-22
+
+### Added
+
+- Standalone config package: `/cmd/api/config/config.go` to handle configuration for the `api` service 
+
+- Shared mock packages `/mocks` that holds mock implementations of `database.go` and `broker.go` to be used for unit testing
```
nanjiangshu:
```suggestion
- Shared mock packages `/mocks` that hold mock implementations of `database.go` and `broker.go` to be used for unit testing
```

sda/CHANGELOG.md
```diff
@@ -7,6 +7,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+### Changed
+
+- `api.go` to use `v2/broker` package
+
+- `api.go` to use `net/http` instead of `gin-gonic/gin` for routing
+
+- `api_test.go` to make use of interfaces for `broker.go` and `database.go` to be able to run tests isolated without docker instances of those services
+
+- `api_test.go` to run table driven test cases for each handler / endpoint
```
nanjiangshu:
```suggestion
- `api_test.go` to run table-driven test cases for each handler / endpoint
```

sda/CHANGELOG.md
```diff
@@ -7,6 +7,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+- Shared mock packages `/mocks` that holds mock implementations of `database.go` and `broker.go` to be used for unit testing
+
+### Changed
+
+- `api.go` to use `v2/broker` package
+
+- `api.go` to use `net/http` instead of `gin-gonic/gin` for routing
+
+- `api_test.go` to make use of interfaces for `broker.go` and `database.go` to be able to run tests isolated without docker instances of those services
```
nanjiangshu:
```suggestion
- `api_test.go` to make use of interfaces for `broker.go` and `database.go` to be able to run tests in isolation without docker instances of those services
```

nanjiangshu (commented):
This refactoring successfully modernized the configuration layer, and the implementation is very clean. Well done! I have left a few comments with suggestions for potential improvement, but overall, it looks excellent.

sda/cmd/api/api.go
zeidlitz: Sure, will add some clarification on it

nanjiangshu (approved):
Looks good to me!

sda/cmd/api/config/config.go
```diff
@@ -0,0 +1,253 @@
...
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.Bool(flagName, false, "Toggles running with/without TLS encryption")
+			},
+			Required: false,
+			AssignFunc: func(flagName string) {
+				ssl = viper.GetBool(flagName)
+			},
+		},
+		&config.Flag{
+			Name: "grpc_host",
```
KarlG-nbis: While we are updating the config, I would prefer if we renamed this to a more descriptive name, eg "reencrypt_grpc_host" (and similar for the other related configs) and having a more descriptive description

sda/mocks/database_mock.go
```diff
@@ -0,0 +1,250 @@
+package mocks
```
KarlG-nbis, line 1: If these mocks are to be generic and used by all unit test better to prepare them for that, ie add the called, and returns from mock.Mock to all impls
etc:
```go
func (m *MockDatabase) UpdateUserInfo(_ context.Context, userID, name, email string, groups []string) error {
  args := m.Called(userID, name, email string, groups)

  return args.Error(0)
}

```

sda/internal/userauth/userauth.go
KarlG-nbis: Do we also want to replace this with 

```go
if err := u.ReadJwtPubKeyBytes(keyData); err != nil {
  return err
}
```

to reduce amount of duplicate code? and possibly update the error returns from `ReadJwtPubKeyBytes` to be more descriptive in which step failed, as it is here

sda/cmd/api/config/config.go
```diff
@@ -0,0 +1,253 @@
...
+			},
+			Required: false,
+			AssignFunc: func(flagName string) {
+				grpcTimeout = viper.GetInt(flagName)
+			},
+		},
+		&config.Flag{
+			Name: "schema_type",
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.String(flagName, "isolated", "Path to JSON schemas to validate rabbitmq messages against")
```
KarlG-nbis: makes sense to include the supported values in the description

sda/cmd/api/config/config.go
```diff
@@ -0,0 +1,253 @@
...
+			Required: false,
+			AssignFunc: func(flagName string) {
+				jwtPubKeyPath = viper.GetString(flagName)
+			},
+		},
+		&config.Flag{
+			Name: "ssl",
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.Bool(flagName, false, "Toggles running with/without TLS encryption")
+			},
```
KarlG-nbis: imo flag is a bit overloaded as it controls both if the http server the sda-api application hosts should be served over TLS, and if the sda-api application should connect to the reencrypt grpc service over tls, would make sense to have separate configuration options for these imo

sda/cmd/api/config/config.go
```diff
@@ -0,0 +1,253 @@
...
+
+func APIAddr() string {
+	return fmt.Sprintf("%s:%d", apiHost, apiPort)
+}
+
+func SSL() bool {
+	return ssl
+}
+
+func CaCert() string {
```
KarlG-nbis: Unused func

sda/cmd/api/config/config.go
jbygdell: Or simply `reencrypt_host`

sda/cmd/api/config/config.go
```diff
@@ -0,0 +1,253 @@
...
+			Name: "schema_type",
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.String(flagName, "isolated", "Path to JSON schemas to validate rabbitmq messages against")
+			},
+			Required: false,
+			AssignFunc: func(flagName string) {
+				schemaType := viper.GetString(flagName)
+				switch schemaType {
+				case "federated":
+					schemaPath = "/schemas/federated/"
```
jbygdell, line 175:
```suggestion
				switch schemaType {
				case "bigpicture":
					schemaPath = "/schemas/bigpicture/"
				case "federated":
					schemaPath = "/schemas/federated/"
```

Bigpicture has its own schema specification.

sda/cmd/api/config/config.go
```diff
@@ -0,0 +1,253 @@
...
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.String(flagName, "", "GRPC Host")
+			},
+			Required: false,
+			AssignFunc: func(flagName string) {
+				grpcHost = viper.GetString(flagName)
+			},
+		},
+		&config.Flag{
+			Name: "grpc_port",
```
jbygdell:
```suggestion
			Name: "reencrypt_port",
```

sda/cmd/api/config/config.go
```diff
@@ -0,0 +1,253 @@
...
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.Int(flagName, 8080, "Port for the api server")
+			},
+			Required: false,
+			AssignFunc: func(flagName string) {
+				apiPort = viper.GetInt(flagName)
+			},
+		},
+		&config.Flag{
+			Name: "ca_cert",
```
jbygdell, line 56:
```suggestion
			Name: "client_ca_cert",
```

sda/cmd/api/config/config.go
```diff
@@ -0,0 +1,253 @@
...
+			Name: "grpc_timeout",
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.Int(flagName, 30, "GRPC Timeout")
+			},
+			Required: false,
+			AssignFunc: func(flagName string) {
+				grpcTimeout = viper.GetInt(flagName)
+			},
+		},
+		&config.Flag{
```
jbygdell, line 165:
```suggestion
		},
		&config.Flag{
			Name: "server_ca_cert", 
			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
				flagSet.String(flagName, "", "HTTP Server TLS CA Certificate file")
			},
			Required: false,
			AssignFunc: func(flagName string) {
				caCert = viper.GetString(flagName)
			},
		},
		&config.Flag{
			Name: "server_key",
			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
				flagSet.String(flagName, "", "HTTP Server TLS Key file")
			},
			Required: false,
			AssignFunc: func(flagName string) {
				clientKey = viper.GetString(flagName)
			},
		},
		&config.Flag{
			Name: "server__cert",
			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
				flagSet.String(flagName, "", "HTTP Server TLS Certificate file")
			},
			Required: false,
			AssignFunc: func(flagName string) {
				clientCert = viper.GetString(flagName)
			},
		},
		&config.Flag{
```

There is no connection between client and server certs since it is possible to run with one or the other in any combinaton.

charts/sda-svc/templates/api-secrets.yaml
```diff
@@ -8,33 +8,42 @@ metadata:
...
-      serverCert: {{ template "tlsPath" . }}/tls.crt
-      serverKey: {{ template "tlsPath" . }}/tls.key
+    rbac_file: {{ template "secretsPath" . }}/rbac.json
+    {{- if .Values.global.tls.enabled }}
+    ssl: true
+    ca_cert: {{ template "tlsPath" . }}/ca.crt
+    client_cert: {{ template "tlsPath" . }}/tls.crt
+    client_key: {{ template "tlsPath" . }}/tls.key
+    {{- end }}
+    grpc_host: {{ required "A valid grpc host is required" .Values.global.reencrypt.host }}
```
jbygdell:
```suggestion
    renecrypt_host: {{ required "A valid grpc host is required" .Values.global.reencrypt.host }}
```

charts/sda-svc/templates/api-secrets.yaml
```diff
@@ -8,33 +8,42 @@ metadata:
...
+    {{- if .Values.global.reencrypt.port}}
+    grpc_port: {{ .Values.global.reencrypt.port }}
```
jbygdell:
```suggestion
    reencrypt_port: {{ .Values.global.reencrypt.port }}
```

charts/sda-svc/templates/api-secrets.yaml
```diff
@@ -8,33 +8,42 @@ metadata:
...
 stringData:
   config.yaml: |-
-    api:
-      rbacFile: {{ template "secretsPath" . }}/rbac.json
-      {{- if .Values.global.tls.enabled }}
-      serverCert: {{ template "tlsPath" . }}/tls.crt
-      serverKey: {{ template "tlsPath" . }}/tls.key
+    rbac_file: {{ template "secretsPath" . }}/rbac.json
+    {{- if .Values.global.tls.enabled }}
+    ssl: true
```
jbygdell:
```suggestion
```

Shouldn't be needed if the client certs are set

charts/sda-svc/templates/api-secrets.yaml
```diff
@@ -8,33 +8,42 @@ metadata:
...
+    ca_cert: {{ template "tlsPath" . }}/ca.crt
```
jbygdell, line 14:
```suggestion
    client_ca_cert: {{ template "tlsPath" . }}/ca.crt
```

charts/sda-svc/templates/api-secrets.yaml
```diff
@@ -8,33 +8,42 @@ metadata:
...
+    client_cert: {{ template "tlsPath" . }}/tls.crt
+    client_key: {{ template "tlsPath" . }}/tls.key
+    {{- end }}
```
jbygdell, line 17:
```suggestion
    client_key: {{ template "tlsPath" . }}/tls.key
    server_cert: {{ template "tlsPath" . }}/tls.crt
    server_key: {{ template "tlsPath" . }}/tls.key
    {{- end }}
```

A bit hackish maybe but better to have proper app config since the chart is partly broken.

sda/cmd/api/config/config.go
zeidlitz: Sure, will adjust

sda/cmd/api/config/config.go
zeidlitz: Will add

sda/cmd/api/config/config.go
zeidlitz: Sure, will split it up into two

sda/cmd/api/config/config.go
zeidlitz: Removed

sda/cmd/api/config/config.go
zeidlitz: Will add

sda/cmd/api/config/config.go
zeidlitz: I find it very unintuitive to separate it since they will point to the same files. Why have 4 sets of variables pointing to the same 2 files?

Won't we just have something like:

`client_cert` -> `tls.crt`
`server_cert` -> `tls.crt` 

`client_key` -> `tls.key`
`server_key` -> `tls.key`

Aren't we just adding extra overhead?

charts/sda-svc/templates/api-secrets.yaml
zeidlitz: adjusted

charts/sda-svc/templates/api-secrets.yaml
zeidlitz: I changed so that the behavior is no longer controlled by whether or not the cert variables are set. So this needs to be set if the point is to run with TLS. But the variable name is inaccurate and I have changed it.

charts/sda-svc/templates/api-secrets.yaml
zeidlitz: done

sda/internal/userauth/userauth.go
zeidlitz: Agreed. Will add

sda/mocks/database_mock.go
zeidlitz: Sure, will adjust

sda/cmd/api/handlers.go
```diff
@@ -0,0 +1,1195 @@
...
+	}
+	if li > maxPageLimit {
+		return 0, fmt.Errorf("invalid limit parameter: must not exceed %d", maxPageLimit)
+	}
+
+	return li, nil
+}
+
+func (api *API) getFiles(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
```
KarlG-nbis: redundant setting of header, as it would be set twice now once here and then by `writeJSON` func, or on [L204](https://github.com/neicnordic/sensitive-data-archive/pull/2482/changes#diff-cab70b780d20855e2fc83cf3558d46228e2c47e76adf7a4aee98ef0a5e342009R204) if no error

sda/mocks/database_mock.go
```diff
@@ -0,0 +1,343 @@
...
+	return args.Get(0).(string), args.Get(1).(string), args.Error(2)
+}
+
+func (m *MockDatabase) GetFileIDByUserPathAndStatus(_ context.Context, submissionUser, filePath, status string) (string, error) {
+	args := m.Called(submissionUser, filePath, status)
+
+	return args.Get(0).(string), args.Error(1)
+}
+
+func (m *MockDatabase) CheckAccessionIDOwnedByUser(_ context.Context, accessionID, user string) (bool, error) {
```
KarlG-nbis: Can be removed as Database interface no longer has this func

KarlG-nbis (approved):
Nice work!

sda/cmd/api/handlers.go
zeidlitz: Good catch! Will adjust

sda/mocks/database_mock.go
zeidlitz: Will adjust

PR #2500, opened 2026-07-15
Feature/api user delete inbox file

### LINKED ISSUE #2428: [s3inbox] User endpoint for deleting a file from their inbox
# Please describe the feature

As an uploader,
I want to be able to delete files from my inbox 
so that I can correct my misstakes.

 * Should send a message to EGA (in case of FEGA).
 * Should update file event table.
 * Should only go through in case a file is either uploaded or disabled.
 * Should only go through in case a file is uploaded by the same user.
 * Needs a new file event "removed".
 * Should use the file id for deletion, file id can be found by using the listing API functionality.
 * Update the admin api for deletion to set the file status to removed instead of disabled.

# Acceptance criteria

- [ ] An API endpoint that can be called with a user to delete a file.
- [ ] Admin API updated to set filestatus to "removed" insetad of "disabled" when deleting a file.
- [ ] Tests verifying the changes are added.

# Additional context

# Estimation of size

small/medium/big

# Estimation of priority

low/medium/high

pahatz:
## Related issue(s) and PR(s)
This PR closes #2428.

### DIFF
--- a/.github/integration/sda/rbac.json
+++ b/.github/integration/sda/rbac.json
@@ -42,7 +42,12 @@
        },
        {
           "role": "submission",
-          "path": "/file/*",
+          "path": "/users/:username/file/:fileid",
+          "action": "DELETE"
+       },
+       {
+          "role": "admin",
+          "path": "/file/:username/:fileid",
           "action": "DELETE"
        },
        {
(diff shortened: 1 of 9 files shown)


pahatz:
As suggested, this will be implemented in s3-inbox

PR #2523, opened 2026-08-14
Feature/sda/otel tracing and metrics

KarlG-nbis:
## Related issue(s) and PR(s)
This PR closes #2520 .

## Description
- Add observability package which adds support for tracing and metrics collecting and exporting with OpenTelemetry
  - Add OpenTelemetry instrumentation for the PostgreSQL database connection
  - Add support for injecting/extracting trace headers on the broker rabbitmq to support distributed tracing
  - Add tracing in the storage reader/writer libraries for s3 and posix
- Initialize observability package and otel middleware for http/grpc client and servers, and add span creating in the following
  - api, auth, download, finalize, ingest, mapper, reencrypt, rotatekey, s3inbox, sync, and verify

## ADR
Relates to [ADR-0006-metrics-and-tracing](https://github.com/neicnordic/sensitive-data-archive/blob/main/docs/decisions/0006-metrics-and-tracing.md)

## How to test

- The integration-sda-s3 will run the stack with observability enabled, and will launch Tempo, Prometheus, and Grafana instances, navigating to localhost:3000 (and logging in with default Grafana credentials) will allow browsing the Prometheus metrics, and traces.


sda/cmd/api/handlers.go
```diff
@@ -388,54 +424,54 @@ func (api *API) ingestFile(w http.ResponseWriter, r *http.Request) {
...
-		if count == 5 {
-			writeJSON(w, http.StatusInternalServerError, "failed to remove file from inbox")
+	if err := api.inboxWriter.RemoveFile(ctx, location, filePath); err != nil {
+		span.Error("failed to remove file from inbox", err, slog.String("location", location), slog.String("file_path", filePath))
+		writeJSON(w, http.StatusInternalServerError, http.StatusText(http.StatusInternalServerError))
 
-			return
-		}
-		time.Sleep(time.Duration(math.Pow(2, float64(count))) * time.Second)
+		return
```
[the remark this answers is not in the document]
KarlG-nbis: nah, if it fails caller should retry.

sda/cmd/auth/main.go
```diff
@@ -417,6 +435,7 @@ func main() {
 	}
 
 	app.Use(sess.Handler())
+	app.Use(iris.FromStd(otelhttp.NewMiddleware("auth")))
```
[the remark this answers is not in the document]
KarlG-nbis: Nah from iris

// FromStd converts native http.Handler, http.HandlerFunc & func(w, r, next) to context.Handler.
//
// Supported form types:
// 		 .FromStd(h http.Handler)
// 		 .FromStd(func(w http.ResponseWriter, r *http.Request))
// 		 .FromStd(func(w http.ResponseWriter, r *http.Request, next http.HandlerFunc))
//
// A shortcut for the `handlerconv#FromStd`.

sda/cmd/rotatekey/rotatekey.go
```diff
@@ -45,8 +48,20 @@ func main() {
...
 	}
 
+	shutdown, err := observability.SetupOTelSDK(ctx, "sda-rotatekey")
+	if err != nil {
+		panic(fmt.Errorf("failed to setup OTel SDK: %v", err))
+	}
+	defer func() {
+		if err := shutdown(ctx); err != nil {
+			slog.Error("failed to shutdown OTel SDK", "err", err)
+		}
```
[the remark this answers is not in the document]
KarlG-nbis: yeah that is to be future improvement

nanjiangshu (approved):
The implementation looks great and the metrics dashboard at grafana works well when I tested. 

The color scheme of heatmap is a bit surprising to me at the beginning since I though white means the least frequent, but it's fine. 

Great work!

sda/cmd/reencrypt/reencrypt.go
```diff
@@ -162,6 +166,25 @@ func main() {
...
+			slog.Error("failed to shutdown OTel SDK", "err", err)
+		}
+	}()
+
+	ctx, startupSpan := observability.StartSpan(ctx, "start up")
+	defer startupSpan.End()
+
 	sigc := make(chan os.Signal, 5)
 	signal.Notify(sigc, os.Interrupt, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
 	defer func() {
```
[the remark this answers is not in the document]
KarlG-nbis: Yeah as mentioned, future improvement

sda/internal/storage/v2/s3/writer/write_file.go
```diff
@@ -42,6 +49,10 @@ func (writer *Writer) WriteFile(ctx context.Context, filePath string, fileConten
 			break
 		}
 	}
+
+	span.SetAttributes(attribute.String("activeBucket", activeBucket))
+	span.SetAttributes(attribute.String("activeEndpoint", writer.activeEndpoint.Endpoint))
+
 	writer.Unlock()
```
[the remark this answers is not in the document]
KarlG-nbis: out of scope to fix as part of this PR

sda/cmd/api/handlers.go
```diff
@@ -584,56 +638,56 @@ func (api *API) setAccession(w http.ResponseWriter, r *http.Request) {
...
 		accession.DecryptedChecksums = []schema.Checksums{{Type: "sha256", Value: fileDecrChecksum}}
 
 	default:
-		slog.Error("requiered parameters not found")
-		writeJSON(w, http.StatusBadRequest, "requiered parameters not found, need either query parameters (fileid, accessionid) or httpbody (filepath & accession_id)")
+		span.Warn("required parameters not found")
+		writeJSON(w, http.StatusBadRequest, "required parameters not found, need either query parameters (fileid, accessionid) or httpbody (filepath & accession_id)")
 
 		return
 	}
```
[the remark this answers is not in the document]
KarlG-nbis: out of scope to fix as part of this PR

sda/internal/broker/v2/rabbitmq/rabbitmq.go
```diff
@@ -78,15 +107,33 @@ func (b *rmqBroker) Publish(ctx context.Context, destinationQueue string, messag
...
+
+	for k, v := range message.Headers {
+		headers[k] = v
+	}
+
+	carrier := propagation.MapCarrier{}
+	otel.GetTextMapPropagator().Inject(ctx, carrier)
+
+	for k, v := range carrier {
+		headers[k] = v
```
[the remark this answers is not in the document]
KarlG-nbis: n/a

sda/cmd/mapper/mapper.go
```diff
@@ -95,6 +110,7 @@ func run() error {
 
 	sigc := make(chan os.Signal, 1)
 	signal.Notify(sigc, os.Interrupt, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
+	startupSpan.End()
```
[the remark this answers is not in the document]
KarlG-nbis: nah, not for now, as we dont want long lived traces, but still want to capture setup actions, eg new broker, new db, etc, and for these to react to the main ctx canellation

sda/cmd/finalize/finalize.go
```diff
@@ -107,6 +124,7 @@ func run() error {
 
 	sigc := make(chan os.Signal, 1)
 	signal.Notify(sigc, os.Interrupt, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
+	startupSpan.End()
```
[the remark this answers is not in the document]
KarlG-nbis: nah, not for now, as we dont want long lived traces, but still want to capture setup actions, eg new broker, new db, etc, and for these to react to the main ctx canellation

sda/cmd/verify/verify.go
```diff
@@ -100,6 +119,7 @@ func run() error {
 
 	sigc := make(chan os.Signal, 1)
 	signal.Notify(sigc, os.Interrupt, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
+	startupSpan.End()
```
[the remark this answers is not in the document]
KarlG-nbis: nah, not for now, as we dont want long lived traces, but still want to capture setup actions, eg new broker, new db, etc, and for these to react to the main ctx canellation

sda/cmd/sync/sync.go
```diff
@@ -107,6 +126,7 @@ func run() error {
 	}
 
 	log.Info("Starting sync service")
+	startupSpan.End()
```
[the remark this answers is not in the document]
KarlG-nbis: nah, not for now, as we dont want long lived traces, but still want to capture setup actions, eg new broker, new db, etc, and for these to react to the main ctx canellation

sda/cmd/rotatekey/rotatekey.go
```diff
@@ -45,8 +48,20 @@ func main() {
...
+	}()
```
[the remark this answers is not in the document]
KarlG-nbis: yeah ok, for now to be fixed in #2536

KarlG-nbis:
@claude review PR

charts/sda-svc/values.yaml
```diff
@@ -14,6 +14,18 @@ global:
...
   jwtPath: "jwt"
 
+  observability:
+    # If application observability is to be enabled, if enabled the sda applications will host a
+    # prometheus endpoint at 9090
+    # and export traces to the configured otelExporterOtlpEndpoint
+    # If global.networkPolicy.create is enabled a network policy to allow egress to the OTLP endpoint, and allow ingress to services at 9090 needs to be added
+    enabled: false
+    # If client transport security(TLS) should be disabled/enabled on the exporter connection
+    otelExporterOtlpInsecure: false
```
jhagberg: I noticed that the deploy templates always render `OTEL_EXPORTER_OTLP_INSECURE` when observability is enabled, with `"false"` by default. In otlptracehttp v1.45.0 (`internal/otlpconfig/envconfig.go`) the endpoint option is applied first, so an `http://` scheme selects insecure, and then the INSECURE option is applied, where an explicit `false` maps to `WithSecure()`.

Would not this mean that an operator who sets `otelExporterOtlpEndpoint: http://tempo:4318` and keeps the default gets TLS attempts against a plaintext receiver, and no traces? The integration compose works because it sets the variable to `true`.

Would it be safer to only render the env var when the value is `true`, so the endpoint scheme decides otherwise?

charts/sda-svc/README.md
```diff
@@ -133,6 +133,9 @@ The following table lists the configurable parameters of the `sda-svc` chart and
...
+| `global.observability.enabled`                   | If application observability is to be enabled, if enabled the sda applications will host a prometheus endpoint at 9090, And export traces to the configured otelExporterOtlpEndpoint.                                                                                                          | `false`                                         |
+| `global.observability.otelExporterOtlpInsecure`  | If client transport security(TLS) should be disabled/enabled on the exporter connection.                                                                                                                                                                                                       | `false`                                         |
+| `global.observability.otelExporterOtlpEndpoint`  | Endpoint to which the applications export the otlp trace information, needs to be set if observability is to be enabled.                                                                                                                                                                       | `""`                                            |
```
jhagberg: Since `otel.go` uses `otlptracehttp`, could we say here that the endpoint must be an OTLP/HTTP receiver, commonly port 4318, with an example such as `http://tempo:4318`? I suspect that is the first thing an operator gets wrong. Until the insecure default is changed, the `http://` example would also need `otelExporterOtlpInsecure: true` next to it. A link to the OTel SDK environment variable spec would also help, the flag help text in `config.go` already has one.

sda/CHANGELOG.md
```diff
@@ -10,6 +10,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - s3inbox: Allow forwarding of the [HeadObject action](https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html)
 - api: add new API /dataset/{dataset_id} for getting state of a dataset
+- Add observability package which adds support for tracing and metrics collecting and exporting with OpenTelemetry
```
jhagberg: I noticed that none of the service docs under `sda/cmd/*/*.md` mention `OBSERVABILITY_ENABLED` or the `OTEL_EXPORTER_OTLP_*` variables. Would it be worth adding a short "Observability settings" subsection to them, or one shared page linked from each service, covering the toggle, the fixed port 9090 with `/metrics`, and that traces go out as OTLP over HTTP?

sda/CHANGELOG.md
```diff
@@ -10,6 +10,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - Add OpenTelemetry instrumentation for the PostgreSQL database connection
+  - Add support for injecting/extracting trace headers on the broker rabbitmq to support distributed tracing
+  - Add tracing in the storage reader/writer libraries for s3 and posix
+- Initialize observability package and otel middleware for http/grpc client and servers, and add span creating in the following
+  - api, auth, download, finalize, ingest, mapper, reencrypt, rotatekey, s3inbox, sync, and verify
```
jhagberg: I noticed that intercept, syncapi and notify are not in this list and do not import the observability package. That seems like a reasonable scope for this PR, but should we open a follow-up issue for them so the ADR's "time spent in each service" outcome has a path to completion?

.github/integration/sda/prometheus.yml
```diff
@@ -0,0 +1,47 @@
+global:
+  scrape_interval: 10s
+scrape_configs:
```
jhagberg: Since the s3 stack now starts Tempo and Prometheus, would it be worth adding a small telemetry check to the integration tests so CI catches a broken exporter? For example at the end of `20_ingest-verify_test.sh`, after the ingest of `NA12878.bam` completes, we could query Prometheus (`prometheus:9090/api/v1/query?query=up{job="ingest"}`) to confirm the scrape works, and poll Tempo search (`tempo:3200/api/search?tags=service.name%3Dsda-ingest`) with a bounded retry for a trace from this ingest. It would need gating, for example on an env var set only in `sda-s3-integration.yml`, since those scripts are shared with the posix and sync stacks. As it is, CI stays green if nothing reaches either of them.

.github/integration/sda-s3-integration.yml
```diff
@@ -16,6 +16,46 @@ services:
       - ./scripts:/scripts
       - shared:/shared
 
+  tempo:
+    image: grafana/tempo:latest
```
jhagberg: I noticed that tempo, prometheus and grafana are unpinned here. Should we pin them to explicit versions so the integration run does not change under us? On a smaller note, the datasource provisioning we use does not need anything from Enterprise, so `grafana/grafana` would do as well, but that is optional.

charts/sda-svc/CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add observability configuration options
```
jhagberg: Maybe this entry could name the three `global.observability.*` keys and the new Services and ports for the internal components (finalize, ingest, mapper, verify, sync, rotatekey), noting that they only render when observability is enabled? Operators reading the changelog before an upgrade would then see what changes for them.

jhagberg (commented):
I had a look at the non-code parts against ADR-0006. The technology choices match the ADR, and `helm template` with observability, network policies, sync and rotatekey enabled renders valid manifests with matching Service selectors and pod labels. That says nothing about reachability under the network policies, which copilot already raised. My inline comments are about the chart default for the insecure flag, which I think blocks export to `http://` endpoints, and about operator docs. The insecure default is the only one I would like to see fixed before merge, the rest can be follow-ups.

After merge, should we update the "More Information" section of ADR-0006 so it points at `sda/internal/observability` and the integration setup instead of the old POC branch, keeping the decision itself as is?

charts/sda-svc/values.yaml
KarlG-nbis: hmm might be worth to just remove OTEL_EXPORTER_OTLP_INSECURE then if the otelExporterOtlpEndpoint controlls the scheme anyway

charts/sda-svc/README.md
KarlG-nbis: yeah makes sense to update the description to mention OTLP/HTTP reciver and example, but for the link to OTel SDK environment variable spec I dont see it helps in this context, as currently not exposing a way for a helm chart user to set those

charts/sda-svc/values.yaml
jhagberg: Agreed, removing it is simpler and matches how the SDK behaves. One small thing then: the scheme becomes mandatory. A value like `tempo:4318` without `http://` parses to an empty host in Go's `url.Parse`, so the exporter ends up secure with nowhere to connect. Maybe the README could say the endpoint must include `http://` or `https://`?

charts/sda-svc/README.md
jhagberg: Fair point on the link, I withdraw that part since the chart does not expose those variables anyway. The OTLP/HTTP wording and example is the important bit. :)

sda/CHANGELOG.md
KarlG-nbis: Yeah good point, will add!

jbygdell:
The only thing I see as a bit strange is that the passive Prometheus exporter is tied to the active OTEL exporter, one is a server the other is a client.
I would expect that the Prometheus endpoint is always active and the OTEL can be switched on when needed.

.github/integration/sda-s3-integration.yml
KarlG-nbis: good point, will update! But will keep the enterprise as that is the recommended version by grafana https://grafana.com/grafana/download

sda/CHANGELOG.md
KarlG-nbis: Yeah can create follow up for those, skipped those since they arent used in BigPicture, but yeah might be worth updating anyway

jhagberg:
I had the same reaction reading `SetupOTelSDK`. With a single flag, an operator who runs Prometheus but no trace backend gets the OTLP client pointed at its default localhost:4318 and export errors every batch, or has to skip metrics. Would a separate toggle for tracing work, or alternatively the contrib autoexport package so `OTEL_TRACES_EXPORTER=none` disables it the standard way? It seems cheaper to settle before the config keys are released.

KarlG-nbis:
> The only thing I see as a bit strange is that the passive Prometheus exporter is tied to the active OTEL exporter, one is a server the other is a client. I would expect that the Prometheus endpoint is always active and the OTEL can be switched on when needed.

Oh yeah thats true, not sure why I went with that approach in hindsight, will update to allow the enabling / disabling of these separatly

> I had the same reaction reading SetupOTelSDK. With a single flag, an operator who runs Prometheus but no trace backend gets the OTLP client pointed at its default localhost:4318 and export errors every batch, or has to skip metrics. Would a separate toggle for tracing work, or alternatively the contrib autoexport package so OTEL_TRACES_EXPORTER=none disables it the standard way? It seems cheaper to settle before the config keys are released.

Yeah will look into using some of the sdk supported configuration to achieve this

charts/sda-svc/values.yaml
```diff
@@ -14,6 +14,21 @@ global:
   tlsPath: "tls"
   jwtPath: "jwt"
 
+  # Observability related environment variable configuration
+  # See https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/ for available environment variables
+  observability:
```
jhagberg: I noticed that `observability:` has only comments under it now, so it ends up as null in the values. The services and deploys then do `eq "prometheus" .Values.global.observability.OTEL_METRICS_EXPORTER`, and Helm fails with `nil pointer evaluating interface {}.OTEL_METRICS_EXPORTER` for anyone who does not set the key.

I tried it with the integration chart values: it fails as is and renders fine with `observability: {}`, which is also what the README lists as the default. Should we set that, or guard the lookups with `dig`?

charts/sda-svc/values.yaml
```diff
@@ -14,6 +14,21 @@ global:
...
+    # If trace exporting is desired, suggested method is OTEL_TRACES_EXPORTER: "otlp"
+    # Then OTEL_EXPORTER_OTLP_ENDPOINT should also most likely be configured which the applications export the otlp trace information, must be an OTLP/HTTP receiver, commonly port hosted on 4318, for example http://tempo:4318.
+    # See https://opentelemetry.io/docs/specs/otel/protocol/exporter/ for additional OTLP Exporter configurations.
+    # OTEL_TRACES_EXPORTER: "none"
+    # OTEL_EXPORTER_OTLP_ENDPOINT: http://tempo:4318
+    #
+    # If metrics exporting is desired, suggested method is OTEL_METRICS_EXPORTER: "prometheus"
+    # OTEL_METRICS_EXPORTER: "prometheus"
+    # With the "prometheus" exporter the applications which support observability will host a Prometheus endpoint (/metrics) on
+    # host: "localhost" (configurable by OTEL_EXPORTER_PROMETHEUS_HOST), and port: 9464 (configurable by OTEL_EXPORTER_PROMETHEUS_PORT)
```
jhagberg: Small one: this says the endpoint is on host "localhost", but `SetupOTelSDK` now defaults `OTEL_EXPORTER_PROMETHEUS_HOST` to 0.0.0.0 and the package README says so too. Would it be worth aligning the comment, since a reader could otherwise think the Service cannot reach it?

charts/sda-svc/values.yaml
```diff
@@ -14,6 +14,21 @@ global:
...
+    # Set additional opentelemetry environment variables under this scope as desired, these will be propagated to the applications.
```
jhagberg: The note about `global.networkPolicy.create` needing egress to the OTLP endpoint and ingress on the metrics port seems to have gone with the rework. Should it come back here, now with 9464 instead of 9090?

jhagberg (commented):
Thanks Karl, the autoexport approach reads much cleaner than the single flag, and it closes the coupling Jocke raised. Two things before I approve: the chart does not render with the default values (see the values.yaml comment), and the branch conflicts with main so only shellcheck ran on 25d5762a. Once it is rebased I will run through it again. The question about a small Prometheus query in the integration test is still open, but that can also be a follow-up if you prefer.

charts/sda-svc/values.yaml
```diff
@@ -14,6 +14,21 @@ global:
...
+  # See https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/ for available environment variables
+  observability:
+    # If trace exporting is desired, suggested method is OTEL_TRACES_EXPORTER: "otlp"
+    # Then OTEL_EXPORTER_OTLP_ENDPOINT should also most likely be configured which the applications export the otlp trace information, must be an OTLP/HTTP receiver, commonly port hosted on 4318, for example http://tempo:4318.
+    # See https://opentelemetry.io/docs/specs/otel/protocol/exporter/ for additional OTLP Exporter configurations.
+    # OTEL_TRACES_EXPORTER: "none"
+    # OTEL_EXPORTER_OTLP_ENDPOINT: http://tempo:4318
+    #
+    # If metrics exporting is desired, suggested method is OTEL_METRICS_EXPORTER: "prometheus"
+    # OTEL_METRICS_EXPORTER: "prometheus"
```
jbygdell:
```suggestion
    # otelMetricsExporter: "prometheus"
```

In the values file these should be CamelCase any switch to capitalized snake case should be done in the deployment templates.

Also recommend having the value "none" as an option here for when metrics collecting should be disabled on the chart level.
Second when the value is "prometheus" service monitors should be created. But since they need extra config something like this needs to be added here:

```
serviceMonitor:
  ## If true, a ServiceMonitor CR is created for a prometheus operator
  ## https://github.com/coreos/prometheus-operator
  ##
  enabled: false
  path: /metrics
  #  namespace: monitoring  (defaults to use the namespace this chart is deployed to)
  labels: {}
  # Set these to override the Prometheus global scrape interval/timeout.
  # interval: 30s
  # scrapeTimeout: 30s
  scheme: http
  tlsConfig: {}
  relabelings: []
  metricRelabelings: []
  basicAuth: {}
  targetLabels: []
```

charts/sda-svc/templates/verify-service.yaml
```diff
@@ -0,0 +1,18 @@
...
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ template "sda.fullname" . }}-verify
+  labels:
+    app: {{ template "sda.fullname" . }}-verify
+spec:
+  ports:
+  - name: prometheus
+    port: {{ default 9464 .Values.global.observability.OTEL_EXPORTER_PROMETHEUS_PORT }}
```
jbygdell:
```suggestion
    port: {{ default 9464 .Values.global.observability.prometheus.port }}
```

These should follow helm semantics

charts/sda-svc/templates/verify-deploy.yaml
```diff
@@ -59,6 +59,16 @@ spec:
           value: {{ include "confFile" . }}
         - name: CONFIG_FILE
           value: {{ include "confFile" . }}
+        {{- range $name, $value := .Values.global.observability }}
+        - name: {{ $name }}
+          value: {{ $value | quote }}
```
jbygdell: Helms values file should use camel case.
so for this to end up as correctly formatted ENVs we need to template them.

{{ snakcase $name | upper }}

charts/sda-svc/values.yaml
jbygdell: I would probably have split this into `traces` and `metrics`, or set as subsections. Since prometheus needs extra config for the service monitors that are outside of what the otel package supplies.

KarlG-nbis:
Moving to draft again as need to rebase, and possibly do some changes after the #2600 PR

.github/integration/sda/prometheus.yml
```diff
@@ -0,0 +1,47 @@
...
+  - job_name: prometheus
+    static_configs:
+      - targets:
+          - prometheus:9090
+  - job_name: s3inbox
+    static_configs:
```
MalinAhlberg: utehsu

charts/sda-svc/values.yaml
KarlG-nbis: Well the problem is that in the helm chart we do not controll the deployment of the promethues server so we do not know where to allow ingress from

and similar for egress of the OTLP traces, we do not know where they will be exported to

charts/sda-svc/values.yaml
KarlG-nbis: The idea was that is a setting similar as [here](https://github.com/grafana-community/helm-charts/blob/main/charts/grafana/values.yaml) where the user can configure the env variables directly which will be propagated to the containers, as I do not think its worth maintaining a custom values to [open telemetry sdk env variables](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/) mapping

charts/sda-svc/templates/verify-service.yaml
KarlG-nbis: See https://github.com/neicnordic/sensitive-data-archive/pull/2523#discussion_r4093947427 for related discussion

charts/sda-svc/templates/verify-deploy.yaml
KarlG-nbis: See https://github.com/neicnordic/sensitive-data-archive/pull/2523#discussion_r4093947427 for related discussion

charts/sda-svc/values.yaml
KarlG-nbis: Also on the service monitors I didnt add any to the helm chart, as the helm chart does not control the grafana, those will be added in the deployment configuration where the chart is used.

PR #2524, opened 2026-08-14
[sda-admin] Optional filtering for listing files

kostas-kou:
## Description
With this PR two optional parameters have been added for the `file list` command.
- The admin can filter the files of the user by using path prefix. With this feature the admin can check files which are belonging in the same dataset.
- The admin can filter the user files by their status.

The path prefix filtering is happening on the api-side and the status filtering is happening on the client side. 

## How to test
This can be tested by running the `sda-admin` and add `4f7df025d3c8c6c645a8be3e6d42b5789ffc953e@lifescience-ri.eu` as user in the `BP-staging`. There are files with different statuses and with different inbox path.

You will need the download token which can be fetched by logging in to the BP-staging.

### DIFF
--- a/sda-admin/README.md
+++ b/sda-admin/README.md
@@ -33,6 +33,25 @@ available, `sda-admin` prints the current page and prompts for `Enter` or
 `Space` before fetching the next page. Users do not need to pass a cursor
 manually.
 
+### Filtering by status
+
+Use `-status` to only show files with a given `fileStatus` (e.g. `uploaded`):
+```sh
+sda-admin file list -user test-user@example.org -status uploaded
+```
+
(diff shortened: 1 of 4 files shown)


nanjiangshu (commented):
Nice features! Except for the comments given by Claude code, which I think are relevant, the PR looks good to me.

pahatz:
Looks good overall, I agree on the claude input.

jhagberg (commented):
Also had a look at this one and agree with @nanjiangshu that the PR looks good overall. The earlier comments from copilot and the Claude review still seem relevant against the current head, so it would be nice to get those threads resolved before merging.

PR #2526, opened 2026-08-18
Migrate Finalize to use the Broker V2 package

### LINKED ISSUE #2533: [finalize] Implement the broker v2 package

jbygdell:
## Description
This PR closes #2533 .

This PR migrates Finalize to use the new V2 Broker package including a proper test suite using the Mock libraries.

### DIFF
--- a/.github/integration/sda-s3-integration.yml
+++ b/.github/integration/sda-s3-integration.yml
@@ -177,7 +177,7 @@ services:
       - BROKER_PASSWORD=finalize
       - BROKER_USER=finalize
       - BROKER_QUEUE=accession
-      - BROKER_ROUTINGKEY=completed
+      - BROKER_ROUTING_KEY=completed
       - DATABASE_PASSWORD=finalize
       - DATABASE_USER=finalize
     restart: always
--- a/charts/sda-svc/templates/finalize-secrets.yaml
+++ b/charts/sda-svc/templates/finalize-secrets.yaml
@@ -35,8 +35,6 @@ stringData:
       port: {{ default (ternary 5671 5672 .Values.global.tls.enabled) .Values.global.broker.port }}
       prefetch_count: {{ default 1 .Values.global.broker.prefetchCount }}
       password: {{ required "MQ password is required" (include "mqPassFinalize" .) }}
-      queue: {{ default "accession" .Values.global.broker.finalizeQueue }}
-      routing_key: {{ default "completed" .Values.global.broker.finalizeRoutingKey }}
     {{- if ne "" ( default "" .Values.global.broker.serverName ) }}
       serverName: {{.Values.global.broker.serverName }}
     {{- end }}
@@ -66,5 +64,7 @@ stringData:
       format: {{ .Values.global.log.format }}
       level: {{ .Values.global.log.level }}
     schema_type: {{ default "federated" .Values.global.schemaType }}
+    source_queue: {{ default "accession" .Values.global.broker.finalizeQueue }}
+    routing_key: {{ default "completed" .Values.global.broker.finalizeRoutingKey }}
 {{- end }}
 {{- end }}
--- a/charts/sda-svc/templates/finalize-secrets.yaml
+++ b/charts/sda-svc/templates/finalize-secrets.yaml
@@ -33,10 +33,10 @@ stringData:
       exchange: {{ default "sda" .Values.global.broker.exchange }}
       host: {{ required "A valid MQ host is required" .Values.global.broker.host }}
       port: {{ default (ternary 5671 5672 .Values.global.tls.enabled) .Values.global.broker.port }}
-      prefetchCount: {{ default 1 .Values.global.broker.prefetchCount }}
+      prefetch_count: {{ default 1 .Values.global.broker.prefetchCount }}
       password: {{ required "MQ password is required" (include "mqPassFinalize" .) }}
       queue: {{ default "accession" .Values.global.broker.finalizeQueue }}
-      routingKey: {{ default "completed" .Values.global.broker.finalizeRoutingKey }}
+      routing_key: {{ default "completed" .Values.global.broker.finalizeRoutingKey }}
     {{- if ne "" ( default "" .Values.global.broker.serverName ) }}
       serverName: {{.Values.global.broker.serverName }}
     {{- end }}
@@ -65,7 +65,6 @@ stringData:
     log:
       format: {{ .Values.global.log.format }}
       level: {{ .Values.global.log.level }}
-    schema:
-      type: {{ default "federated" .Values.global.schemaType }}
+    schema_type: {{ default "federated" .Values.global.schemaType }}
 {{- end }}
 {{- end }}
--- a/charts/sda-svc/templates/finalize-secrets.yaml
+++ b/charts/sda-svc/templates/finalize-secrets.yaml
@@ -24,10 +24,10 @@ stringData:
     {{- include "sda.locationBrokerCacheTTL" (dict "ctx" . "priority" (list "backupArchive" "archive")) | nindent 4 }}
     broker:
     {{- if .Values.global.tls.enabled }}
-      caCert: {{ template "tlsPath" . }}/ca.crt
+      ca_cert: {{ template "tlsPath" . }}/ca.crt
       {{- if .Values.global.broker.verifyPeer }}
-      clientCert: {{ template "tlsPath" . }}/tls.crt
-      clientKey: {{ template "tlsPath" . }}/tls.key
+      client_crt: {{ template "tlsPath" . }}/tls.crt
+      client_key: {{ template "tlsPath" . }}/tls.key
       {{- end }}
     {{- end }}
       exchange: {{ default "sda" .Values.global.broker.exchange }}
@@ -36,12 +36,12 @@ stringData:
       prefetch_count: {{ default 1 .Values.global.broker.prefetchCount }}
       password: {{ required "MQ password is required" (include "mqPassFinalize" .) }}
     {{- if ne "" ( default "" .Values.global.broker.serverName ) }}
-      serverName: {{.Values.global.broker.serverName }}
+      server_name: {{.Values.global.broker.serverName }}
     {{- end }}
       ssl: {{ .Values.global.tls.enabled }}
       user: {{ required "MQ user is required" (include "mqUserFinalize" .) }}
     {{- if .Values.global.tls.enabled }}
-      verifyPeer: {{ .Values.global.broker.verifyPeer }}
+      verify_peer: {{ .Values.global.broker.verifyPeer }}
     {{- end }}
       vhost: {{ include "brokerVhost" . }}
     database:
--- a/sda/cmd/finalize/config/config.go
+++ b/sda/cmd/finalize/config/config.go
@@ -17,7 +17,7 @@ var (
 func init() {
 	config.RegisterFlags(
 		&config.Flag{
-			Name: "broker.queue",
+			Name: "source_queue",
 			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
 				flagSet.String(flagName, "accession", "The queue where the service consumes messages from")
 			},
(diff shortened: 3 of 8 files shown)


sda/cmd/finalize/finalize.go
```diff
@@ -22,278 +23,151 @@ import (
...
 
-	// Mark file as "ready"
-	if err := db.UpdateFileEventLog(ctx, fileID, "ready", "finalize", "{}", string(delivered.Body)); err != nil {
-		log.Errorf("set status ready failed, file-id: %s, reason: %v", fileID, err)
-		if err := delivered.Nack(false, true); err != nil {
-			log.Errorf("failed to Nack message, reason: %v", err)
-		}
-
-		return
+		return nil, fmt.Errorf("file with file-id: %s is not verified yet, aborting work", message.Key)
```
KarlG-nbis: Should we requeue the message in this case?

sda/cmd/finalize/finalize.go
```diff
@@ -323,24 +197,138 @@ func backupFile(ctx context.Context, delivered amqp.Delivery) error {
...
+	switch accessionIDExists {
+	case "duplicate":
+		log.Errorf("accession ID already exists in the system, file-id: %s, accession-id: %s\n", message.Key, ingestionAccession.AccessionID)
+		// Send the message to an error queue so it can be analyzed.
+		return []func(){app.errorQueue(message, "Duplicate accession ID")}, nil
+	case "same":
+		log.Infof("file already has an accession ID, marking it as ready, file-id: %s", message.Key)
+	default:
+		if app.archiveReader != nil && app.backupWriter != nil {
+			if err = app.backupFile(ctx, message); err != nil {
```
KarlG-nbis: imo we should pass the tx to backupFile, so we dont have to begin another one inside it, and so that if tx.SetAccessionID, etc fails, the db execs from within `backupFile` is also rolled back

sda/cmd/finalize/finalize.go
```diff
@@ -22,278 +23,151 @@ import (
...
 
-	archiveData, err := db.GetArchived(ctx, fileID)
+	archiveData, err := app.db.GetArchived(ctx, message.Key)
 	if err != nil {
 		return fmt.Errorf("failed to get file archive information, reason: %v", err)
 	}
 
 	if archiveData == nil {
-		return fmt.Errorf("file archive data not found in database, file-id: %s", fileID)
+		return fmt.Errorf("file archive data not found in database, file-id: %s", message.Key)
```
KarlG-nbis: Should we requeue the message in this case, as a retry wouldnt be expected to succed if the data is missing in the db

sda/cmd/finalize/finalize.go
```diff
@@ -302,7 +176,7 @@ func backupFile(ctx context.Context, delivered amqp.Delivery) error {
 		return fmt.Errorf("archive file size does not match registered file size, (disk size: %d, db size: %d)", diskFileSize, archiveData.FileSize)
 	}
```
KarlG-nbis: Should we requeue the message in this case, as a retry wouldnt be expected to succeed if the data size is different from db and s3

KarlG-nbis:
Missing some of the configuration changes needed to configure broker v2 in the sda-svc chart, or is the plan to do those as part of another PR?

sda/cmd/finalize/finalize.go
jbygdell: As I understand the V2 library any errors that are propagated up through `handleMessages` will result in the message getting nacked and requeued.

jbygdell:
> Missing some of the configuration changes needed to configure broker v2 in the sda-svc chart, or is the plan to do those as part of another PR?

Intentionally left out at this point.

sda/cmd/finalize/finalize.go
KarlG-nbis: Yeah thats how it works. What I mean is do we want to requeue in this case when the latest file_event is something finalize isnt expecting, since would we expect the that state to have changed on a retry?

KarlG-nbis: @jbygdell Do we want to keep requeue here?

sda/cmd/finalize/finalize.go
```diff
@@ -22,289 +24,166 @@ import (
...
 	// If the file has been canceled by the uploader, don't spend time working on it.
-	status, err := db.GetFileStatus(ctx, fileID)
+	status, err := app.db.GetFileStatus(ctx, message.Key)
 	if err != nil {
-		log.Errorf("failed to get file status, file-id: %s, reason: %v", fileID, err)
-		if err := delivered.Nack(false, true); err != nil {
-			log.Errorf("failed to Nack message, reason: %v", err)
+		log.Errorf("failed to get file status, file-id: %s, reason: %v", message.Key, err)
+
+		if err == sql.ErrNoRows {
```
KarlG-nbis: would suggest to use `errors.Is(err, sql.ErrNoRows)` for more robust error comparision

sda/cmd/finalize/finalize.go
```diff
@@ -323,24 +202,124 @@ func backupFile(ctx context.Context, delivered amqp.Delivery) error {
...
+	if app.archiveReader != nil && app.backupWriter != nil {
+		if callbacks, err := app.backupFile(ctx, tx, message); err != nil {
+			log.Errorf("failed to backup file, file-id: %s, reason: %v", message.Key, err)
+
+			if callbacks != nil {
+				// Send the message to an error queue  but don't requeue it
+				return callbacks, nil
+			}
+
+			return nil, err
```
KarlG-nbis: minor: in theory if `app.backupFile` returns callbacks, and nil error, the callbacks wouldnt be called, but doesnt affect current logic as it doesnt return such atm

sda/cmd/finalize/config/config.go
```diff
@@ -0,0 +1,74 @@
...
+var (
+	sourceQueue string
+	routingKey  string
+	schemaPath  string
+)
+
+func init() {
+	config.RegisterFlags(
+		&config.Flag{
+			Name: "broker.queue",
```
KarlG-nbis: Imo, this is a little overstepping domain boundaries for the finalize service to register flags under broker.

The idea for the source and destination not being on the broker level, is that in theory an application could consume from multiple queues, and publish to multiple (currently they kinda do, but the "error" queue is hardcoded), and having these kinda of configs outside broker simplifies how to handle such scenario

jbygdell: It is.
But it also makes the configuration consistent with what we have.
Modifying these will result in a braking change configuration wise and as such none of these changes should go into main until we have the same behaviour for all apps.

sda/cmd/finalize/finalize.go
jbygdell: That's the question.
Since this is a race condition error we can just decide how we want to deal with it.

This is starting to become a full refactoring rather that just a simple library swap.

sda/cmd/finalize/config/config.go
KarlG-nbis: Created branch https://github.com/neicnordic/sensitive-data-archive/tree/feature/broker_v2_migration for this

sda/cmd/finalize/finalize_test.go
```diff
@@ -1,20 +1,251 @@
...
+	accession := "file-asdfg-1234"
+
+	ts.mockDB.On("GetFileStatus", fileID).Return("uploaded", nil)
+
+	message := createMessage(filePath, userName, accession, fileID)
+	_, err := ts.app.handleMessage(context.Background(), message)
+	assert.Equal(ts.T(), fmt.Sprintf("file with file-id: %s is not verified yet, aborting work", fileID), err.Error())
+}
+
+func (ts *TestSuite) TestHandleMessage_missing() {
```
pahatz: minor: Maybe capitalize these to keep the same style

pahatz (approved):
Don't have any comments, lgtm.

sda/cmd/finalize/config/config.go
```diff
@@ -0,0 +1,74 @@
...
+			},
+			Required: false,
+			AssignFunc: func(flagName string) {
+				sourceQueue = viper.GetString(flagName)
+			},
+		},
+		&config.Flag{
+			Name: "broker.routing_key",
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.String(flagName, "completed", "The routing key where the service publishes messages to")
```
nanjiangshu:
```suggestion
				flagSet.String(flagName, "completed", "The routing key used by the service to publish messages")
```

Maybe better wording like this. Since we **don't publish** to a routing key.

nanjiangshu (approved):
Looks good to me. Just left a very minor comment.

sda/cmd/finalize/config/config.go
nanjiangshu: Yeah, I think better to use `source_queue` or `broker.source_queue` for consistency. Will it be addressed in that branch https://github.com/neicnordic/sensitive-data-archive/tree/feature/broker_v2_migration?

charts/sda-svc/templates/finalize-secrets.yaml
```diff
@@ -33,10 +33,8 @@ stringData:
       exchange: {{ default "sda" .Values.global.broker.exchange }}
       host: {{ required "A valid MQ host is required" .Values.global.broker.host }}
       port: {{ default (ternary 5671 5672 .Values.global.tls.enabled) .Values.global.broker.port }}
-      prefetchCount: {{ default 1 .Values.global.broker.prefetchCount }}
+      prefetch_count: {{ default 1 .Values.global.broker.prefetchCount }}
```
jhagberg: I noticed that the `broker:` block here (lines 27-31) still uses `caCert`, `clientCert` and `clientKey`, while the v2 broker registers the flags as `broker.ca_cert`, `broker.client_cert` and `broker.client_key` (`sda/internal/broker/v2/rabbitmq/config.go:108-134`). Would not this mean the certificates are never picked up when TLS is enabled?

The ingest chart already uses the snake_case names, so maybe we should follow that here as well? The `serverName` and `verifyPeer` keys a few lines down seem to have no counterpart in v2 at all, but that looks like a gap in the library rather than in this PR.

.github/integration/sda-s3-integration.yml
```diff
@@ -179,7 +179,7 @@ services:
       - BROKER_PASSWORD=finalize
       - BROKER_USER=finalize
       - BROKER_QUEUE=accession
```
jhagberg: I noticed that `config.go` now registers `source_queue` and `routing_key` without the `broker.` prefix (`sda/cmd/finalize/config/config.go:20,30`), so I think the env vars would be `SOURCE_QUEUE` and `ROUTING_KEY`. Would not this mean `BROKER_QUEUE` and `BROKER_ROUTING_KEY` here are no longer read, and the test only passes because the values match the defaults?

`sda-finalize --help` lists `SOURCE_QUEUE` and `ROUTING_KEY` as the env variables, and `finalize.md:60-61` documents the same two old names, so maybe both should be updated together?

charts/sda-svc/templates/finalize-secrets.yaml
jhagberg: Thanks for the fix. I noticed that line 29 became `client_crt` while the flag is [`broker.client_cert`](https://github.com/neicnordic/sensitive-data-archive/blob/154d9fda3197ccf4a2cdfcbc2ff28848c57bac21/sda/internal/broker/v2/rabbitmq/config.go#L118), and the loader [only builds the keypair](https://github.com/neicnordic/sensitive-data-archive/blob/154d9fda3197ccf4a2cdfcbc2ff28848c57bac21/sda/internal/broker/v2/rabbitmq/config.go#L224) when both `clientCert` and `clientKey` are non-empty, so would not this mean the whole client keypair is still skipped? The `database:` block on line 50 and the verify/mapper charts on the base branch spell it `client_cert`.

Also, `server_name` (39) and `verify_peer` (44) have no flags in v2 either, and e3647dd3 removed them from the verify chart rather than renaming them. Maybe we should do the same here?

.github/integration/sda-s3-integration.yml
jhagberg: The readme update looks good. I noticed that `.github/integration/sda-s3-integration.yml:179-180` still sets `BROKER_QUEUE` and `BROKER_ROUTING_KEY` for finalize though, and the same rebase dropped those two lines from the verify service in the same file. Maybe finalize should follow?

.github/integration/sda-sync-integration.yml
```diff
@@ -177,8 +177,8 @@ services:
     environment:
       - BROKER_PASSWORD=finalize
       - BROKER_USER=finalize
-      - BROKER_QUEUE=accession
-      - BROKER_ROUTINGKEY=completed
+      - source_QUEUE=accession
```
jhagberg: Should this not be SOURCE_QUEUE ?

jbygdell: yes

PR #2540, opened 2026-08-24
[s3inbox]: log reasons for failed readiness check 

### LINKED ISSUE #1540: [s3inbox] log reason for readiness failure

As a systems administrator, I want the s3inbox to log the reason why the readiness probe failed so that it is possible to do postmortem and figure out the root cause of the failure.
When s3inbox can't connect to the s3 backend it should give a reason and we should log that. 

## Acceptance criteria

- [ ] Healtcheck errors are logged
- [ ] (Tests verifying the changes are added)

## Additional context

Currently we can see things like this:
```
Readiness probe failed: Get "https://10.42.219.31:8000/health": net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```
But no further information on what the root cause was.

## Estimation of size

small/medium/big

## Estimation of priority

low/medium/high

nanjiangshu:
## Related issue(s) and PR(s)
This PR closes #1540 

## Description
This PR enhances the s3inbox readiness probe logging by enriching error details when the S3 backend health check fails or times out.

## How to test
Unit tests pass

### DIFF
--- a/sda/cmd/s3inbox/healthchecks.go
+++ b/sda/cmd/s3inbox/healthchecks.go
@@ -61,7 +61,7 @@ func (p *Proxy) CheckHealth(w http.ResponseWriter, r *http.Request) {
 
 		return
 	}
-	err = p.httpsGetCheck(s3url)
+	err = p.httpsGetCheck(r.Context(), s3url)
 	if err != nil {
 		log.Error(err)
 		w.WriteHeader(http.StatusServiceUnavailable)
@@ -72,9 +72,9 @@ func (p *Proxy) CheckHealth(w http.ResponseWriter, r *http.Request) {
 }
 
 // httpsGetCheck sends a request to the S3 backend and makes sure it is healthy
-func (p *Proxy) httpsGetCheck(uri string) error {
+func (p *Proxy) httpsGetCheck(ctx context.Context, uri string) error {
 	// Use a dedicated context timeout for readiness checks so s3inbox logs the timeout first
-	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+	ctx, cancel := context.WithTimeout(ctx, 15*time.Second)
 	defer cancel()
 
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, uri, nil)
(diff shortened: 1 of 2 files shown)


sda/cmd/s3inbox/healthchecks_test.go
```diff
@@ -85,9 +87,38 @@ func (ts *HealthcheckTestSuite) TearDownTest() {
...
 
+	// Success path (200 OK)
 	url, _ := p.getS3ReadyPath()
 	assert.NoError(ts.T(), p.httpsGetCheck(url))
-	assert.Error(ts.T(), p.httpsGetCheck("http://127.0.0.1:8888/nonexistent"), "404 should fail")
+
+	// HTTP Non-200 Status Response (Server reachable, but path not found)
+	err := p.httpsGetCheck("http://127.0.0.1:8888/nonexistent")
+	assert.Error(ts.T(), err)
+	assert.Contains(ts.T(), err.Error(), "S3 check to http://127.0.0.1:8888/nonexistent returned status 404")
```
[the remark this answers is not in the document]
nanjiangshu: fixed.

sda/cmd/s3inbox/healthchecks.go
```diff
@@ -69,13 +73,25 @@ func (p *Proxy) CheckHealth(w http.ResponseWriter, r *http.Request) {
 
 // httpsGetCheck sends a request to the S3 backend and makes sure it is healthy
 func (p *Proxy) httpsGetCheck(uri string) error {
-	resp, e := p.client.Get(uri) // #nosec G704 uri originates from configuration
+	// Use a dedicated context timeout for readiness checks so s3inbox logs the timeout first
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
jbygdell:
```suggestion
	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
```

This might be to low in real life, I suggest we set the timeout to 50% of the default client timeout.

KarlG-nbis: Would be better to pass the request context, to `httpsGetCheck` and use that for this, such that in the scenario where the callers cancels the health check this call is also cancelled instead of waiting for the timeout

KarlG-nbis (approved):

kostas-kou (approved):
Looks good to me

PR #2541, opened 2026-08-25
docs(download): document session cookie authentication in v2 API spec

jhagberg:
## Related issue(s) and PR(s)
This PR closes #2369.

## Description
The download v2 service issues a session cookie on the first authenticated
request, and neither `swagger_v2.yml` nor `download.md` said so.

`download.md` now covers the two caches (token-keyed and session-keyed) and why
the cookie is not what makes caching work, the cookie attributes, how the TTL
is bounded, when no cookie is issued, the per-replica and revocation
consequences, and what a browser client can and cannot do with a
`SameSite=Lax` cookie.

`swagger_v2.yml` gains a `SetSessionCookie` response header referenced from
every response the middleware can emit it on, plus session-cache notes on
`info.description` and `bearerAuth`. The cookie is deliberately not declared as
an alternative security scheme: the server accepts it alone, but the cache is
local to one replica, so no client should present it as its only credential.

Documentation only, no code changes.

## ADR

- [ ] This PR includes an architecturally significant decision (see `docs/decisions/README.md`)
  - [ ] A decision record has been added or updated in `docs/decisions/`
  - [ ] The decision record is listed in the `docs/decisions/README.md` index table

## How to test
Docs only. `swagger_v2.yml` parses and validates against the OpenAPI 3.0.3
schema, and `markdownlint-cli2 sda/cmd/download/download.md` reports no new
errors compared to `main`.

PR #2543, opened 2026-08-25
test(download): probe and test the session cookie with an unseen token

### LINKED ISSUE #2542: test(download): Test17_SessionCookieReuse always skips because the cookie probe reuses an already-cached token
## Describe the bug

`.github/integration/tests/download/main_test.go` gates `Test17_SessionCookieReuse`
(line 459) on `ts.hasSessionCache`, which is set by a probe in `probeCapabilities()`
(line 62). The probe sends `GET /datasets` with `ts.token` and looks for an
`sda_session` cookie in the response (lines 81-91).

`probeCapabilities()` starts by calling `getFirstFileID()` (line 944), which already
made an authenticated `GET /datasets` with the same token. That first request is the
one that authenticates from scratch and receives `Set-Cookie`. By the time the probe
runs, the token is in the token-keyed cache and the response carries no cookie, so
`hasSessionCache` is always false and Test17 skips on every run. The skip message
("visa processing may prevent caching in combined mode", line 480) points at the
wrong cause.

Test17 has the same problem internally: its own request uses `ts.token`, so even with
a fixed probe `Require().NotNil(sessionCookie)` at line 483 would fail.

No unit test asserts `Set-Cookie` either (`middleware/visa_middleware_test.go:175`
only covers reading the legacy cookie name), so the session cookie currently has no
executable coverage at all.

## Steps to reproduce

1. Open a recent run of `pr_build_and_test_images.yml`, e.g. run 32856396371
   (2026-08-25), download v2 integration job.
2. Search the log for `sessionCache=` and `Test17_SessionCookieReuse`.
3. Observe `Environment capabilities: ... sessionCache=false` and
   `--- SKIP: TestDownloadTestSuite/Test17_SessionCookieReuse`.

Running the same configuration locally (`dev-tools/download-v2-dev`,
`permission.model: combined`) with a fresh token, the first request does return
`Set-Cookie: sda_session=...` and the second request with the same token does not.

## Expected behavior
- [ ] The probe and Test17 each use a token the service has not seen before
      (e.g. `ts.generateToken("probe@example.org")`), or the probe runs before any
      other authenticated request in `probeCapabilities()`.
- [ ] Test17 runs in CI (not skipped) and asserts: `Set-Cookie: sda_session` on the
      first request, no `Set-Cookie` on a second request with the same token, and
      200 on a request carrying only the cookie.

## Additional context

Found while documenting the session cookie in #2541 (for #2369). Behaviour verified
manually against the PR2541 image.

## Estimation of size
small

## Estimation of priority
low

jhagberg:
## Related issue(s) and PR(s)
This PR closes #2542.

## Description
`Test17_SessionCookieReuse` never ran. `probeCapabilities()` sent `GET /datasets`
with `ts.token` and looked for an `sda_session` cookie, but `getFirstFileID()` had
already authenticated with that token a few lines earlier. The service only sets
the cookie on a request it authenticates from scratch, so the probe hit the token
cache, saw no cookie, and the test skipped on every CI run with a message blaming
combined-mode visa processing. Test17 used `ts.token` for its own first request
too, so it would have failed even with a working probe.

The probe and Test17 now mint a token the service has not seen. `generateToken`
adds a nanosecond `jti`, since `iat`/`exp` have second precision and two tokens
for the same subject minted in the same second were otherwise identical.

Test17 also asserts what the docs in #2541 describe: `Set-Cookie` on the first
request, no new `Set-Cookie` on a second request with the same token, and 200 on
a request carrying only the cookie.

## ADR

## How to test
Run the download v2 integration job and check the log: it should show
`sessionCache=true` and `--- PASS: TestDownloadTestSuite/Test17_SessionCookieReuse`
instead of the `--- SKIP` every earlier run has.

Verified locally against `dev-tools/download-v2-dev` on the PR2541 image: the
file on `main` logs `sessionCache=false` and skips, this one logs
`sessionCache=true` and passes.

### DIFF
--- a/.github/integration/tests/download/main_test.go
+++ b/.github/integration/tests/download/main_test.go
@@ -78,19 +78,17 @@ func (ts *TestSuite) probeCapabilities() {
 	resp, _, reqErr := ts.doRequest("GET", "/files/"+fileID+"/content", nil, ts.authHeaders())
 	ts.hasStorageFile = reqErr == nil && resp.StatusCode == http.StatusOK
 
-	// Probe session cookie: check if first auth request sets sda_session
-	req, _ := http.NewRequestWithContext(context.Background(), "GET", ts.downloadURL+"/datasets", nil)
-	req.Header.Set("Authorization", "Bearer "+ts.token)
-	cookieResp, cookieErr := http.DefaultClient.Do(req)
-	if cookieErr == nil {
-		for _, c := range cookieResp.Cookies() {
-			if c.Name == "sda_session" {
-				ts.hasSessionCache = true
-
-				break
-			}
+	// Probe session cookie. The service only sets sda_session on a request it
+	// authenticates from scratch, and ts.token is already in its token cache
+	// from getFirstFileID above, so probe with a token it has not seen.
+	if probeToken, tokenErr := ts.generateToken("integration_test@example.org"); tokenErr == nil {
+		req, _ := http.NewRequestWithContext(context.Background(), "GET", ts.downloadURL+"/datasets", nil)
+		req.Header.Set("Authorization", "Bearer "+probeToken)
+		cookieResp, cookieErr := http.DefaultClient.Do(req)
+		if cookieErr == nil {
+			ts.hasSessionCache = sessionCookie(cookieResp) != nil
+			cookieResp.Body.Close()
 		}
-		cookieResp.Body.Close()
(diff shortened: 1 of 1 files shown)


nanjiangshu (approved):
Looks good to me!

KarlG-nbis (approved):

PR #2545, opened 2026-08-27
fix(storage-v2): honour writer_disabled when selecting writers

jhagberg:
## Related issue(s) and PR(s)

No issue. Found while working on the sda-svc chart 4.0 multi-endpoint refactor
in #2427, which exposes `writerDisabled` to operators and documents it as the
way to run a mixed backend.

## Description

`sda/internal/storage/v2/README.md` documents `writer_disabled` as "if this is
just the config for a reader", but neither writer consulted it when verifying
endpoints. `loadConfig` uses the flag only to skip per-entry required-field
validation and then returns the full slice, and `NewWriter` appended every
returned entry to its endpoint list.

A backend with a reader-only entry on one side and an active writer on the
other therefore produced a non-nil writer from both implementations, so
`storage.NewWriter` returned `ErrorMultipleWritersNotSupported` and the service
failed to start:

```yaml
storage:
  archive:
    s3:
    - endpoint: https://s3.example.com
      bucket_prefix: archive
      writer_disabled: true
    posix:
    - path: /archive
```

Both `NewWriter` loops now skip writer-disabled entries. The list passed to
`RegisterSizeAndCountFinderFunc` is built before the loop and is unchanged, so
reads and size/count lookups still cover reader-only locations.

One thing I would like your view on: `RemoveFile` resolves its client through
`getS3ClientForEndpoint(ctx, writer.configuredEndpoints, endpoint)`, so a
writer-disabled location is no longer in that list and deleting an object there
would not resolve a client. As far as I can tell that path was already
unreachable in a mixed setup, since `storage.NewWriter` returns a single writer
and the posix writer could never have served an s3 location anyway. But if the
intent is that reader-only locations stay deletable, this needs a different
shape and I would rather you decide it than me.

## ADR

## How to test

```
cd sda && go test ./internal/storage/v2/...
```

The four tests in `internal/storage/v2/writer_test.go` cover a reader-only
entry on either side, two active writers still conflicting, and every entry
being reader-only. Reverting either `continue` guard fails the first three.

### DIFF
--- a/sda/internal/storage/v2/posix/writer/config.go
+++ b/sda/internal/storage/v2/posix/writer/config.go
@@ -33,7 +33,12 @@ func loadConfig(backendName string) ([]*endpointConfig, error) {
 		return nil, err
 	}
 
+	var writerEndpoints []*endpointConfig
+
 	for _, e := range endpointConf {
+		// A writer_disabled entry is the config for a reader. Writing to it is
+		// not supported, and deleting counts as a write, so it is dropped here
+		// and the writer never sees it.
 		if e.WriterDisabled {
 			continue
 		}
@@ -47,9 +52,11 @@ func loadConfig(backendName string) ([]*endpointConfig, error) {
 			}
 			e.maxSizeBytes = byteSize.Bytes()
 		}
+
+		writerEndpoints = append(writerEndpoints, e)
 	}
 
-	return endpointConf, nil
+	return writerEndpoints, nil
 }
 
 func (endpointConf *endpointConfig) isUsable(ctx context.Context, backendName string, locationBroker locationbroker.LocationBroker) (bool, error) {
--- a/sda/internal/storage/v2/s3/writer/config.go
+++ b/sda/internal/storage/v2/s3/writer/config.go
@@ -57,7 +57,12 @@ func loadConfig(backendName string) ([]*endpointConfig, error) {
 		return nil, err
 	}
 
+	var writerEndpoints []*endpointConfig
+
 	for _, e := range endpointConf {
+		// A writer_disabled entry is the config for a reader. Writing to it is
+		// not supported, and deleting counts as a write, so it is dropped here
+		// and the writer never sees it.
 		if e.WriterDisabled {
 			continue
 		}
@@ -110,9 +115,11 @@ func loadConfig(backendName string) ([]*endpointConfig, error) {
 				e.Region = "us-east-1"
 			}
 		}
+
+		writerEndpoints = append(writerEndpoints, e)
 	}
 
-	return endpointConf, nil
+	return writerEndpoints, nil
 }
 
 func (endpointConf *endpointConfig) getS3Client(ctx context.Context) (*s3.Client, error) {
--- a/sda/internal/storage/v2/writer_test.go
+++ b/sda/internal/storage/v2/writer_test.go
@@ -1,0 +1,202 @@
+package storage
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/neicnordic/sensitive-data-archive/internal/storage/v2/storageerrors"
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/assert"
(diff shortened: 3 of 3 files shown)


jhagberg:
@claude review

sda/internal/storage/v2/s3/writer/writer.go
```diff
@@ -47,6 +47,13 @@ func NewWriter(ctx context.Context, backendName string, locationBroker locationb
 
 	// Verify endpointConfig connections
 	for _, e := range endPointConf {
+		// A writer_disabled entry is the config for a reader, so it must not
+		// make this backend count as having an s3 writer. It stays in the list
+		// handed to RegisterSizeAndCountFinderFunc above, so reads and
+		// size/count lookups still cover it.
+		if e.WriterDisabled {
```
KarlG-nbis: could make sense to edit the `loadConfig` func instead so the disabled endpoints are not included in the returned slice, since then no worry about the `RegisterSizeAndCountFinderFunc`

eg something like:


```go
var endpointConf, configuredEndpoints []*endpointConfig

if err := viper.UnmarshalKey(
	"storage."+backendName+".posix",
	&endpointConf,
	func(config *mapstructure.DecoderConfig) {
		config.WeaklyTypedInput = true
		config.ZeroFields = true
	},
); err != nil {
	return nil, err
}
for _, e := range endpointConf {
	if e.WriterDisabled {
		continue
	}   
    ...
    configuredEndpoints = append(configuredEndpoints, e)   
}

return configuredEndpoints, nil
```

jhagberg: Thanks for that, your version is tidier than mine!

I think it runs into the same thing the bot caught though. Both `RemoveFile` implementations look up the target in `configuredEndpoints` (`s3/writer/remove_file.go:18` and `posix/writer/remove_file.go:20`), and that slice is built from what `loadConfig` returns, so filtering there would mean an object at the reader-only endpoint can no longer be deleted. I tried it against main with two s3 entries where one is `writer_disabled`, and `RemoveFile` goes from nil to `no endpoint configured for location`. That was my own bug first!

So now I keep them in `configuredEndpoints` and only keep them out of writer selection. Posix was easy since `activeEndpoints` is already separate.

Or would you rather we say that deleting from a reader-only location is simply not supported?

KarlG-nbis: Yeah I would say deleting from a reader-only location is not to be supported, as deleting is imo a type of "write" action

jhagberg: Makes sense, that is a much better way to think about it. Moved the filtering into `loadConfig` as you suggested, so both `writer.go` files are back to what they look like on main.

One side effect worth knowing: the disabled entries also disappear from the list handed to `RegisterSizeAndCountFinderFunc`, so size and count lookups for those locations stop working too. Seems right to me since that is only used for capacity checks when picking somewhere to write, but shout if you disagree.

KarlG-nbis (approved):

jbygdell (approved):

sda/internal/storage/v2/s3/writer/writer.go
```diff
@@ -47,6 +47,13 @@ func NewWriter(ctx context.Context, backendName string, locationBroker locationb
...
+			continue
+		}
 		_, err := e.findActiveBucket(ctx, backendName, writer.locationBroker)
 		if err != nil {
 			if errors.Is(err, storageerrors.ErrorNoFreeBucket) {
```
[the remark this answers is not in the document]
jhagberg: Real one, thanks. It went the other way in the end though: Karl's take is that deleting counts as a write, so a reader-only location should not be a valid target at all. The filtering moved into `loadConfig` instead, and there is a test asserting `RemoveFile` at such a location returns `ErrorNoEndpointConfiguredForLocation`.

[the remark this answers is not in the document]
jhagberg: Agreed, and pre-existing as you say, so it stays out of this PR. Karl spotted a second one right next to it in `WriteFile`, where `activeBucket` can still be empty after the rollover loop and the upload then runs with an empty bucket name. Both are tracked in #2546 and fixed in #2547.

PR #2551, opened 2026-08-28
Refactor/s3inbox broker v2

### LINKED ISSUE #2537: [s3inbox] Implement the broker v2 package

KarlG-nbis:
## Related issue(s) and PR(s)
This PR closes #2537 .

### s3inbox:
- Update to use broker/v2 instead of broker (v1)
- Fix reuploads to publish the "remove" message before the "upload" messages, instead of after 
- Update unit test to use mocks instead of docker containers
- Add additional checks in detectS3RequestType, do reject additional s3 action no previously rejected but not supported
- Remove dependency on config (v1) by moving required configuration registration to s3inbox/config

### sda-svc
- Populate broker_v2 config for s3inbox

## How to test
- integration test pass
- unit test pass


KarlG-nbis:
@claude review pr

KarlG-nbis:
@claude review pr

sda/cmd/s3inbox/config/config.go
```diff
@@ -0,0 +1,185 @@
...
+	serverJwtPubKeyURL  string
+	serverCert          string
+)
+
+func init() {
+	config.RegisterFlags(
+		&config.Flag{
+			Name: "destination_queue",
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.String(flagName, "inbox", "The queue where the s3inbox service publishes messages to")
```
jbygdell:
```suggestion
			Name: "routing_key",
			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
				flagSet.String(flagName, "inbox", "The routing key s3inbox service publishes messages with")
```

We should try to keep the terminology consistent with what RabbitMQ uses.

KarlG-nbis: I do not agree with that, in my opinion this terminology should be generic, as otherwise this whole broker_v2 work which made the broker interface generic from the implementation was quite pointless.

But if the team agrees with keeping the terminology consistent with RabbitMQ I will update.

sda/cmd/s3inbox/proxy.go
```diff
@@ -290,14 +301,29 @@ func (p *Proxy) handleUpload(s3RequestType S3RequestType, w http.ResponseWriter,
...
+		if isReupload {
+			log.Infof("user: %s, reuploaded file: %s, with id: %s, checksum: %s", username, filePath, fileID, checksum)
+			if err := p.sendMessageOnOverwrite(context.WithoutCancel(r.Context()), username, fileID, s3FilePath); err != nil {
+				p.internalServerError(w, token.Subject(), r.Method, r.URL.Path, r.URL.RawQuery, err.Error())
+
+				return
+			}
+		} else {
+			log.Infof("user: %s, uploaded file: %s, with id: %s, checksum: %s", username, filePath, fileID, checksum)
+		}
```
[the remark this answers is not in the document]
KarlG-nbis: thats ok for now

sda/cmd/s3inbox/proxy_test.go
```diff
@@ -1,752 +1,811 @@
...
-	// Create a configuration for the fake MQ
-	s.MQConf = broker.MQConf{
-		Host:     "127.0.0.1",
-		Port:     mqPort,
-		User:     "guest",
-		Password: "guest",
-		Vhost:    "/",
-		Exchange: "",
-	}
+				ms.On("ServeHTTP", http.MethodGet, "/test_inbox_bucket", "location=").Return(200, nil, nil).Once()
```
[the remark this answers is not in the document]
KarlG-nbis: n/a

sda/cmd/s3inbox/proxy_test.go
```diff
@@ -1,752 +1,811 @@
...
-	assert.True(s.T(), proxy.messenger.Connection.IsClosed())
-	r, _ := http.NewRequest("PUT", "/dummy/mqunavailable-file", nil)
-	w := httptest.NewRecorder()
-	s.fakeServer.resp = "<ListBucketResult xmlns=\"http://s3.amazonaws.com/doc/2006-03-01/\"><Name>test</Name><Prefix>/elixirid/db-test-file.txt</Prefix><KeyCount>1</KeyCount><MaxKeys>2</MaxKeys><Delimiter></Delimiter><IsTruncated>false</IsTruncated><Contents><Key>/elixirid/file.txt</Key><LastModified>2020-03-10T13:20:15.000Z</LastModified><ETag>&#34;0a44282bd39178db9680f24813c41aec-1&#34;</ETag><Size>5</Size><Owner><ID></ID><DisplayName></DisplayName></Owner><StorageClass>STANDARD</StorageClass></Contents></ListBucketResult>"
-	s.fakeServer.headHeaders = map[string]string{"ETag": "\"0a44282bd39178db9680f24813c41aec-1\"", "Content-Length": "5"}
-	proxy.ServeHTTP(w, r)
-	assert.Equal(s.T(), 500, w.Result().StatusCode) // nolint:bodyclose
-}
+func (m *mockAuthenticator) Authenticate(r *http.Request) (jwt.Token, error) {
+	args := m.Called(r.Header.Get("Authorization"))
```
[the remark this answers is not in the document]
KarlG-nbis: n/a

sda/cmd/s3inbox/proxy_test.go
```diff
@@ -1,752 +1,811 @@
...
+					// cant verify the port as the mock s3 server gets different port each run by httptest, so just verify it sets the bucket and format correctly
+					u, err := url.Parse(loc)
+
+					return err == nil &&
+						u.Scheme == "http" &&
+						u.Hostname() == "127.0.0.1" &&
+						u.Path == "/test_inbox_bucket"
+				}), "test", "unit_test_user").Return("file_id_123", nil).Once()
+				mdb.On("Commit").Return(nil).Once()
+				ms.On("ServeHTTP", http.MethodPost, "/test_inbox_bucket/unit_test_user/test", "uploads=").Return(200, nil, nil).Once()
```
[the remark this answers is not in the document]
KarlG-nbis: n/a

sda/cmd/s3inbox/config/config.go
nanjiangshu: I think it would be better to change this to `routing_key`, since the value is actually used as a routing key when publishing the message.

The issue is rooted in the Publish function in [rabbitmq.go](https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/internal/broker/v2/rabbitmq/rabbitmq.go#L84). The second argument, currently named `destinationQueue`, is passed directly as the routing key to `PublishWithContext`. Therefore, I think the parameter should be renamed from `destinationQueue` to `routingKey`, and the corresponding configuration name should also be changed to `routing_key`.

nanjiangshu (commented):
Looks good to me, except for the `routing_key` naming issue. In my opinion, this is not just a matter of consistency with RabbitMQ terminology; the current name is technically misleading because the value is actually used as the routing key when publishing the message.

sda/cmd/s3inbox/config/config.go
KarlG-nbis: I dont argue that the currently named `destinationQueue` is used as the `routingKey` in the rabbitmq implementation, what I argue for is readability, and clarity from the application / developer perspective. 

If we feel that `routing_key` is a more descriptive configuration then sure, but then the [broker v2 interface](https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/internal/broker/v2/broker.go#L12) should also name it RoutingKey instead of DestinationQueue for consistency.

But what I think is that what's important when configuring and developing an application is what queue the messages end up in, and sure for rabbitmq that controlled by the `routing_key` but we do not have scenarios where a routing_key name differs from the queue the message ends up in. Sure we have some fanouts where routing_key: `mappings` goes to two queues, `mappings`, and `sync`, etc, but that setup is due to how rabbitmq works and not supporting having multiple consumer on the same queue getting all the messages with separate acks.

But will update configuration to routing_key for now, and we can revisit in future if needed.

nanjiangshu: > If we feel that routing_key is a more descriptive configuration then sure, but then the [broker v2 interface](https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/internal/broker/v2/broker.go#L12) should also name it RoutingKey instead of DestinationQueue for consistency.

I actually agree with this, unlike 
```golang
Subscribe(ctx context.Context, sourceQueue string, handleFunc func(ctx context.Context, msg *Message) ([]func(), error)) error
```
in which `sourceQueue` is referring to a queue

```golang
Publish(ctx context.Context, destinationQueue string, message Message) error
```

`destinationQueue` refers to a routing_key 

But we can leave this to another issue.

nanjiangshu (approved):
Looks good to me now. Nice work!

sda/cmd/s3inbox/healthchecks.go
```diff
@@ -10,40 +10,25 @@ import (
...
-		}
-	}
-
-	if p.messenger.Channel.IsClosed() {
-		log.Warning("channel is closed, recreating...")
-		err := p.messenger.CreateNewChannel()
-		if err != nil {
-			log.Warning(err)
-			w.WriteHeader(http.StatusServiceUnavailable)
+	if !p.broker.Alive() {
```
jhagberg: `CheckHealth` (healthchecks.go:25) now answers 503 when `!p.broker.Alive()`, where the previous version recreated the connection and channel in the same place. In broker/v2 the reconnect lives in `ensureConnected`, which only `Publish` and `Subscribe` call (the `NotifyClose` goroutine at rabbitmq.go:292 just logs).

I tried this with the real `CheckHealth` against a RabbitMQ container, with the DB and S3 checks mocked as healthy. Before the restart it answered 200. After `docker restart` it answered 503 once a second for 20 seconds while RabbitMQ was back up the whole time, and the first `Publish` reconnected, after which the next check answered 200. So for s3inbox, which only publishes, a broker restart in Kubernetes would go: readiness fails, the pod leaves the Service endpoints, no upload arrives, nothing calls `Publish`, and with `s3Inbox.livenessProbe` commented out in values.yaml the pod stays NotReady until someone restarts it.

The api service on main has the same shape, so this is probably a broker/v2 question rather than something for this PR. Would it be worth a reconnect method on the `Broker` interface that the health checks can call, or should we enable the liveness probe in the chart instead?

sda/CHANGELOG.md
```diff
@@ -29,6 +29,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - verify:
   - Update to use broker/v2 instead of broker (v1)
   - Add unit tests
+- s3inbox:
```
jhagberg: `s3inbox.md` still lists `SERVER_JWTPUBKEYPATH`, `SERVER_JWTPUBKEYURL`, `BROKER_ROUTINGKEY`, `BROKER_QUEUE` and `BROKER_PREFETCHCOUNT` (lines 46-59). After this PR the names are `SERVER_JWT_PUB_KEY_PATH`, `SERVER_JWT_PUB_KEY_URL`, the top-level `ROUTING_KEY` and `BROKER_PREFETCH_COUNT`, and the old ones bind to nothing.

Because config/v2 ignores unknown keys, an operator following the doc would hit the "no JWT public key url or JWT public key path specified" abort at s3inbox.go:102-104, or get the default routing key without any error. `BROKER_ROUTINGKEY=inbox` in `sda-s3-integration.yml:102`, `sda-sync-integration.yml:101`, `sda-validator-orchestrator-integration.yml:244` and `dev-tools/key-rotation-test/compose.yml:163` is in the same situation and only passes CI because the value equals the new default, and the six `dev-tools/key-rotation-test/config*.yaml` files still use `jwtpubkeyurl`.

Could we update the doc and the compose files in this PR, as the verify and mapper PRs did? It would also help to mention the renamed keys in the CHANGELOG entry, since they break deployments that do not render config from the chart.

sda/cmd/s3inbox/proxy_test.go
```diff
@@ -1,752 +1,811 @@
...
-	_ = ts.Listener.Close()
-	ts.Listener = l
-	ts.Start()
 
-	f.ts = ts
+			w := httptest.NewRecorder()
+			p.ServeHTTP(w, tc.req)
 
-	return &f
+			assert.Equal(t, 200, w.Code)
```
jhagberg: A question on coverage. The mock-based suite asserts the 200 (proxy_test.go:306) and 403 (proxy_test.go:773) responses, but the 400, 401 and 500 paths the old docker suite exercised are gone, and so are `TestServeHTTP_concurrent_put_noRace` (the #2295 regression guard) and the `formatUploadFilePath` unit test.

Would it be worth adding table cases for those status codes and for the mixed-slash and disallowed-character branches of `formatUploadFilePath` (proxy.go:584-597)? A small concurrent PUT case under `-race` would keep the #2295 guard as well.

The remove-before-upload order on reupload that copilot asked about earlier is also still not asserted, as far as I can see.

jhagberg (commented):
Thanks Karl, the migration itself looks good to me. I have left three inline comments: a question about the readiness check after a broker restart (which I tried against a RabbitMQ container), one about the docs, compose files and CHANGELOG that still carry the old key names, and one about coverage in the rewritten proxy tests.

One small thing I could not attach inline since the lines are outside the diff: `s3-inbox-secrets.yaml:58-59` still renders a `schema:` / `type:` block, but the new `cmd/s3inbox/config` registers no schema key and broker/v2 does no schema validation, so I think that key is dead now and could go the same way as `serverName` and `verifyPeer`.

sda/cmd/s3inbox/healthchecks.go
KarlG-nbis: hmm good spot, I would say this is a broker/v2 issue, and that the broker/v2 should know / check this somehow, for now to fix this I would suggest updating the broker rabbitmq `Alive` implementation to 

```go
func (b *rmqBroker) Alive() bool {
	if err := b.ensureConnected(b.ctx); err != nil {
		return false
	}

	return true
}
func (b *rmqBroker) alive() bool {
	b.mu.Lock()
	defer b.mu.Unlock()

	if b.connection == nil || b.connection.IsClosed() {
		return false
	}

	if b.publishChannel == nil || b.publishChannel.IsClosed() {
		return false
	}

	if b.consumeChannel == nil || b.consumeChannel.IsClosed() {
		return false
	}

	return true
}

func (b *rmqBroker) ensureConnected(ctx context.Context) error {
	if b.alive() {
		return nil
	}

	if err := ctx.Err(); err != nil {
		return err
	}

	if err := b.connect(); err != nil {
		log.Errorf("failed to reconnect, reason: %v", err)

		return err
	}

	log.Info("successfully reconnected to broker")

	return nil
}
```

Do you see any issues with this?

sda/cmd/s3inbox/proxy_test.go
KarlG-nbis: Added back TestServeHTTP_concurrent_put_noRace, added test cases for malformed upload paths, and add NotBefore checks to ensure remove message is published before upload if file already exists

sda/cmd/s3inbox/healthchecks.go
jhagberg: Thanks, and I see it is in as 3fa0fe13 now. It fixes the case I hit: the first probe after RabbitMQ is back reconnects and readiness recovers on its own, so the liveness probe is not needed for this. Four things I would still look at, none of them a blocker for this PR, but worth deciding where they go.

1. `ensureConnected` drops `b.mu` between `alive()` and `connect()`. A probe and a Publish that both see a dead connection both dial, and the second `connect()` closes the first one's fresh connection and channels, so a Publish that is between reading `ch` (rabbitmq.go:72-75) and `PublishWithContext` fails on a closed channel. Two concurrent Publishes can do that already, but a probe every 10 seconds adds a third participant. If the close lands after RabbitMQ accepted the message, the caller gets "publish confirm channel closed" for a message that was in fact published. Holding the mutex across the check and the reconnect closes the window; wrapping the existing methods in `b.mu` as they are would deadlock on the nested locks, so it needs a locked helper.

2. `connect()` holds `b.mu` through `amqp.DialConfig`. amqp091-go v1.14.0 uses a 30 second TCP dial timeout (connection.go:30) and then a second 30 second deadline for the TLS and AMQP handshake (connection.go:212), so if RabbitMQ is unreachable rather than refusing, one probe can hold the mutex for up to a minute while kubelet gives up after `timeoutSeconds: 9`, and every Publish waits on the same mutex in `alive()`. `Alive()` also runs on `b.ctx` rather than the request context, so a probe kubelet has abandoned keeps dialing. `amqpConf.Dial = amqp.DefaultDial(...)` with something below the probe timeout would at least bound the TCP part.

3. `Alive()` dials after `Close()` too, since nothing marks the broker as closed. In s3inbox the deferred `server.Shutdown` runs before `mqBroker.Close()`, so no probe reaches the handler afterwards; api has the same order but builds its broker on `context.Background()`, so `ctx.Err()` would never stop a reconnect there. #2587 adds a `closed` flag in the same file for the drain case, checked under the lock inside `connect()`, so the two changes belong together. Since broker/v2 lives on main, would you rather put 3fa0fe13 on main and I rebase #2587 on it, or shall I carry the `Alive()` change in #2587?

4. The CHANGELOG entry covers the s3inbox env changes but not this broker/v2 behaviour change. A line under broker/v2 would help the other services notice that a readiness probe now reconnects.

sda/cmd/s3inbox/proxy_test.go
jhagberg: Thanks, that covers what I asked for: 401 and 400 have cases, the concurrent PUT is back under `-race`, and the remove-before-upload order is pinned with `NotBefore`. Only the lint job is unhappy with the new test (proxy_test.go:1004, 1008 and 1036, two empty lines and `fileId`).

sda/cmd/s3inbox/proxy.go
```diff
@@ -290,14 +301,29 @@ func (p *Proxy) handleUpload(s3RequestType S3RequestType, w http.ResponseWriter,
...
+
 		jsonMessage, err := json.Marshal(message)
 		if err != nil {
 			p.internalServerError(w, token.Subject(), r.Method, r.URL.Path, r.URL.RawQuery, fmt.Sprintf("failed to marshal rabbitmq message to json: %v", err))
 
 			return
 		}
 
-		if err = p.checkAndSendMessage(fileID, jsonMessage); err != nil {
+		if err := p.broker.Publish(context.WithoutCancel(r.Context()), p.routingKey, broker.Message{
```
jhagberg: `context.WithoutCancel(r.Context())` here and at line 307 keeps the publish alive when the client goes away, which is what we want, but it also drops every deadline, and `Publish` waits for the confirmation with `case <-ctx.Done()` as the only way out (rabbitmq.go:102-112). If a confirmation never arrives and the channel stays open, the upload handler waits forever. Would `context.WithTimeout(context.WithoutCancel(r.Context()), <a few seconds>)` be a reasonable middle ground?

jhagberg:
A heads-up in case the `cancel()` before return pattern from #2563 and #2579 comes here as well: the server shutdown context at s3inbox.go:148 is `context.WithTimeout(ctx, 10*time.Second)` with the run context as parent. With `ctx` already cancelled, `server.Shutdown` still closes the listeners but returns `context.Canceled` without waiting for the in-flight requests, and the broker and database are closed right after, so uploads in progress would be cut off instead of getting the 10 seconds. `context.WithoutCancel(ctx)` as the parent keeps the drain. api.go:167 has the same shape.

sda/cmd/s3inbox/proxy.go
KarlG-nbis: yeah makes sense!

sda/cmd/s3inbox/healthchecks.go
KarlG-nbis: Yeah just carry the Alive() change in https://github.com/neicnordic/sensitive-data-archive/pull/2587, I think thats the easiest atm

KarlG-nbis: And yeah we probably should look into a more robust solution that does not have the potential locking and tcp dial timeout issues

jhagberg: Sounds good, I will carry the `Alive()` change in #2587 and take the dial-under-mutex and the `alive()`/`connect()` check-then-act along with it there, so this PR does not need to wait for them. Then the only thing left on my side here is the shutdown context question in the issue-level comment: `server.Shutdown` at `s3inbox.go:148` runs on the already cancelled `ctx`, so it returns `context.Canceled` without waiting for active requests. Would it be fine to give it its own `context.WithTimeout(context.Background(), 10*time.Second)`?

jhagberg: With the rebase and the CI fixes in, I think this thread is the only thing left on my side before approving. `server.Shutdown` at s3inbox.go:148 still runs on the already cancelled `ctx`, so it returns `context.Canceled` straight away and in-flight uploads get cut off. Would you take `context.WithTimeout(context.Background(), 10*time.Second)` in this PR, or do you prefer to leave it as is and I note it on #2459 instead? Either way is fine with me, I mainly want to know which one so I can wrap up the review.

jhagberg:
@KarlG-nbis one small thing before I approve, which I think got lost since I replied in a thread that was already resolved ([here](https://github.com/neicnordic/sensitive-data-archive/pull/2551#discussion_r4024409370)). `server.Shutdown` at s3inbox.go:148 runs on the already cancelled `ctx`, so `context.WithTimeout(ctx, 10*time.Second)` is cancelled from the start and in-flight uploads get cut off instead of getting the 10 seconds. Would you take `context.WithTimeout(context.Background(), 10*time.Second)` in this PR, or do you prefer to leave it as is and I note it on #2459 instead? Either way is fine with me, I mainly want to know which so I can wrap up the review.

KarlG-nbis:
> @KarlG-nbis one small thing before I approve, which I think got lost since I replied in a thread that was already resolved ([here](https://github.com/neicnordic/sensitive-data-archive/pull/2551#discussion_r4024409370)). `server.Shutdown` at s3inbox.go:148 runs on the already cancelled `ctx`, so `context.WithTimeout(ctx, 10*time.Second)` is cancelled from the start and in-flight uploads get cut off instead of getting the 10 seconds. Would you take `context.WithTimeout(context.Background(), 10*time.Second)` in this PR, or do you prefer to leave it as is and I note it on #2459 instead? Either way is fine with me, I mainly want to know which so I can wrap up the review.

Ah yeah lost that one sorry, Ill address it here as its a quick fix

jhagberg (approved):
Thanks Karl, `context.WithoutCancel(ctx)` is a neat way to do it, the shutdown gets its 10 seconds and the context values are kept. With that in, everything from my review is addressed and CI is green, so approving. Nice work getting the whole series through.

.github/integration/sda-s3-integration.yml
```diff
@@ -99,7 +99,7 @@ services:
     environment:
       - BROKER_PASSWORD=inbox
       - BROKER_USER=inbox
-      - BROKER_ROUTINGKEY=inbox
+      - ROUTINGKEY=inbox
```
nanjiangshu:
```suggestion
      - ROUTING_KEY=inbox
```

Shouldn't it be `ROUTING_KEY` as `routing_key` is set in the config

Same for the other integration test compose yaml files

nanjiangshu (approved):
The PR looks good to me now. 

However, I left a minor comment regarding the ENV for `routing_key`. No blocking.

.github/integration/sda-s3-integration.yml
KarlG-nbis: Ah yeah that was just a typo, will fix!

PR #2553, opened 2026-08-31
docs(rfcs): spell out the review loop, sync ADR statuses, lint RFCs

jhagberg:
Reviewing #2453 surfaced a contradiction in `docs/rfcs/README.md`: step 5 says an RFC merges in its current state and keeps exploring in the repo, while the "How this relates to ADRs" table says an RFC PR lives indefinitely. The table row was a leftover from the first draft of #2401, not a deliberate model. In practice the table's model is the one being followed: all four RFC PRs (#2263, #2320, #2449, #2453) are open with a single commit each while review feedback accumulates in the threads.

This PR keeps step 5's model, which is what ADR-0005 set out to do ("a git artifact in its own right, instead of living only in a GitHub PR thread"), and writes down the pieces that were missing:

* Review feedback is folded into the file before merge: new options under Considered Options, objections under Pros and Cons, the rest under Open Questions. This is the common rule across Rust, Python PEPs, Swift Evolution and Kubernetes KEPs; PEP 1 even makes "Rejected Ideas" a required section.
* A lightweight merge gate: one approval plus the thread's substance in the file. Merging publishes an editable exploration, not team agreement; an objection still live at merge time becomes a named Open Question for the next meet-up.
* The body is editable while `exploring` or `ready-for-decision` and frozen once `promoted` or `withdrawn`. The README used to freeze everything but `exploring`, which contradicted ADR-0005.
* The index row is part of writing the RFC (steps 1–5 never mentioned it; #2449 and #2453 lack one).
* A `discussion` front-matter field links the file to its PR, as Rust (`RFC PR`), PEP (`Discussions-To`), Swift (`Review`) and Oxide RFDs (`discussion`) do.
* The table row now says "Short" for both columns, with a new "Time to a conclusion" row carrying the indefinite part.

Second commit: ADR-0005 is flipped to `accepted`. It merged in #2401 on 2026-04-24 and both Confirmation steps are delivered (#2263/#2320 converted to RFC-0001/0002). Per `docs/decisions/README.md`, `proposed` means "under discussion in a PR". The same commit syncs the index row for ADR-0003, whose file has said `accepted` since 2026-03-13 while the index still said `proposed`.

Third commit: the markdown lint workflow only covered `docs/decisions/**`, so RFC files were never linted. It now runs on `docs/rfcs/**` as well, with the same `.markdownlint.yml`.

Follow-up for the open RFC PRs, not part of this PR: fold the thread into the file and merge per the new step 6. For #2453 that means the objection in the changes-requested review (keep `api`, drop the `sda-admin` CLI) and the third option raised in review (split user and admin APIs) belong in the file before it merges.

### DIFF
--- a/.github/workflows/adr_lint_markdown.yml
+++ b/.github/workflows/adr_lint_markdown.yml
@@ -5,19 +5,21 @@ on:
   pull_request:
     paths:
       - "docs/decisions/**"
+      - "docs/rfcs/**"
       - ".github/workflows/adr_lint_markdown.yml"
       - ".markdownlint.yml"
 
 jobs:
   markdownlint:
-    name: Decision records
+    name: Decision records and RFCs
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@v7
 
-      - name: Lint decision records
+      - name: Lint decision records and RFCs
         uses: DavidAnson/markdownlint-cli2-action@v24
         with:
           globs: |
             docs/decisions/**/*.md
+            docs/rfcs/**/*.md
--- a/docs/decisions/0005-introduce-rfcs-as-upstream-exploration-phase.md
+++ b/docs/decisions/0005-introduce-rfcs-as-upstream-exploration-phase.md
@@ -1,6 +1,6 @@
 ---
-status: proposed
-date: "2026-04-20"
+status: accepted
+date: "2026-04-24"
 decision-makers:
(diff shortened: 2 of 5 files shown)


jbygdell (approved):

nanjiangshu (approved):
Well written!

MalinAhlberg (approved):
Well written, and sounds reasonable to me!

PR #2554, opened 2026-09-01
Feature/user delete inbox file

### LINKED ISSUE #2428: [s3inbox] User endpoint for deleting a file from their inbox

pahatz:
## Related issue(s) and PR(s)
This PR closes #2428.

### DIFF
--- a/sda/cmd/s3inbox/proxy.go
+++ b/sda/cmd/s3inbox/proxy.go
@@ -77,6 +77,7 @@ const (
 	AbortMultiPartUpload
 	GetBucketLocation
 	HeadObject
+	DeleteObject
 )
 
 // NewProxy creates a new S3Proxy. This implements the ServerHTTP interface.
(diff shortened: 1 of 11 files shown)


sda/cmd/s3inbox/proxy.go
```diff
@@ -568,7 +577,77 @@ func (p *Proxy) checkFileExists(ctx context.Context, s3FilePath string) (bool, e
...
+	}()
+
+	if s3Response.StatusCode >= 200 && s3Response.StatusCode < 300 {
+		if err := p.sendInboxRemoveMessage(username, fileID, s3FilePath); err != nil {
+			p.internalServerError(w, token.Subject(), r.Method, r.URL.Path, r.URL.RawQuery, err.Error())
+
+			return
+		}
+
+		if err := p.database.UpdateFileEventLog(r.Context(), fileID, "disabled", username, "{}", "{}"); err != nil {
```
KarlG-nbis: we should probably use a different event for this, as `disabled` event is treated as if event is in the `uploaded` state in the ingest service (or use a different event for the cancel action)

jbygdell (changes requested):
The status checks for removed must be included in ingest and maybe finalize as well.
Including the DB version check in the init portion.

KarlG-nbis (approved):

jbygdell (approved):

viklund:
Add a changelogentry

pahatz:
Thank you @KarlG-nbis

PR #2555, opened 2026-09-01
docs(rfc): RFC-0005 explore where to handle CORS for browser clients

jhagberg:
## Related issue(s) and PR(s)

There is no issue for this. It was drafted as ADR-0007 after sprint planning
leaned towards handling CORS at the ingress without settling it, and converted
to RFC-0005 per
[ADR-0005](https://github.com/neicnordic/sensitive-data-archive/blob/main/docs/decisions/0005-introduce-rfcs-as-upstream-exploration-phase.md)
before opening, because application layer vs edge is still open. Builds on the
session cookie documentation merged in #2541.

## Description

An archive operator wants to run web UIs against the SDA API services from
other origins and wants an answer that covers all services. The RFC lays out
the options and what the current code does to each of them.

Two things I found while writing it that hold whichever option wins:

- `sda-auth` is the only service with CORS today, and its chart wiring is
  broken: `global.auth.corsOrigins`, `corsMethods` and `corsCreds` render keys
  the service never reads, and nothing tests the plumbing.
- Download v2 has no `OPTIONS` routes, so a preflight falls through to gin's
  `404 page not found` before any group middleware runs.

The RFC compares five options (application middleware, ingress or reverse
proxy, do nothing and let browser apps proxy, same-origin routing at the edge,
chart-bundled gateway or sidecar). It also has a header contract for download
v2 derived from what the handlers read and set, a service inventory, and
acceptance criteria written against option 1 that say which of them carry over
to the edge.

The driver is other browser applications, not sda-download-ui. That UI proxies
through its own origin on purpose, and its authors consider the CORS
prerequisite in their older design doc outdated. The RFC leaves the proxy
alone; open question 8 asks what a direct browser path to v2 would need, with
the File System Access resume work in NBISweden/sda-download-ui#173 as the
browser-side half.

**Direction currently favoured:** option 1, application-layer middleware with
a configured origin allowlist, with option 4 open to any site that prefers it.
The edge-error gap (ingress 502/503/504 reach browsers without CORS headers)
and the release dependency are the counter-arguments I take most seriously.
Status in the front matter is `exploring`.

Eight open questions. The ones I most want input on: which layer (1), whether
v1 and `doa` are out of scope (2), and how to migrate direct `sda-auth`
deployments off `*` and patterns without a login outage (6).

## How to test

- Review the RFC at `docs/rfcs/0005-cors-application-layer-or-edge.md`
- Code references are permalinks pinned to main at 60389bd1; check the claims
  against them
- Verify the Open Questions reflect what is actually unresolved on the team
- `markdownlint-cli2 docs/rfcs/0005-cors-application-layer-or-edge.md` passes

### DIFF
--- a/docs/rfcs/0005-cors-application-layer-or-edge.md
+++ b/docs/rfcs/0005-cors-application-layer-or-edge.md
@@ -1,0 +1,461 @@
+---
+status: exploring
+date: "2026-08-28"
+discussion: ""
+authors:
+  - "@jhagberg"
+consulted: []
+informed: []
+---
+
+# Where to handle CORS for browser clients: application layer or edge
+
+## Context and Problem Statement
+
+An archive operator wants to run web UIs against the SDA API services from
+other origins, and wants an answer that covers all services. Sprint planning
+leaned towards the ingress but did not settle it. This RFC lays out the
+options and what the current code does to each of them.
+
+Only `sda-auth` handles CORS today. It registers
+`iris-contrib/middleware/cors` when `cors.origins` is set
+([main.go#L409-L417][auth-cors]), [auth.md][auth-md-cors] documents a
+credentialed browser login via `/oidc/cors_login`, and the chart has
+`global.auth.corsOrigins`, `corsMethods` and `corsCreds`
+([values.yaml#L192-L197][values-auth-cors]). Those values never reach the
+service, though: what the template renders
+([auth-secrets.yaml#L18-L23][auth-secrets]) does not match the keys the
+service reads ([config.go#L427-L440][config-cors]). The service also accepts
+`*` and `null` origins. No other service sends `Access-Control-*` headers.
+
+The one browser client we have, [sda-download-ui][ui-repo], does not need
+CORS. Both of its download modes, the tar stream ([tar-download.md][ui-tar])
+and the [File System Access batch download][ui-fsa], go through its own
+origin, and that proxy ([`/api/files/:fileId`][ui-proxy]) does things CORS
+cannot: it keeps the bearer and the recipient key on the server, gives the
+browser only a session cookie, and stitches the per-request Crypt4GH header
+onto the stable `/content` stream so the browser sees one resumable file.
+The CORS prerequisite in its older [decision document][ui-design] is
+outdated, according to its authors. This RFC leaves that proxy alone. What a
+direct browser path to v2 would need is open question 8.
+
+So the driver is general. Other browser applications should be able to call
+the API directly, at any site, without each of them building a proxy. That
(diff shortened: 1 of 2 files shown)


jbygdell:
Given that no applications really needs CORS since all APIs require the JWT token in each call I say we can let this slide. And if we feel like we need this later on implement it at the gateway/ingress/reverse-proxy level.

The CORS stuff in `sda-auth` is not used and maybe not even fully functional as it is today. The idea there where to enable CLI based login. Something that since then hasn't been worked on.

jhagberg:
Thanks Jocke, good to have your view in the thread.

I agree that no application calls the API cross-origin from a browser today, and the RFC says the same: the one browser client we have proxies through its own origin. So "do nothing now" is Option 3 in the file, and I can record it as your preference, with the edge as the fallback if the need comes back.

I read the token argument a bit differently. A bearer on every call does not remove the need for CORS for a browser application, it only changes what CORS has to do. The `Authorization` header makes every call a preflighted request, so a page on another origin cannot read any response without the headers. The bearer does mean there is no ambient credential, so `Allow-Credentials` is not needed and there is no security reason to hurry.

Would it be fair to say that what is left to settle is whether the ask for direct browser access is deferred? @viklund, since this came out of sprint planning, do you know if that ask is still live or if it can wait? If it can wait, I would move the RFC to ready-for-decision with Option 3 now and Option 2 later as the outcome, and leave the header contract in the file for whoever configures the edge.

On `sda-auth`: thanks for the background on the CLI login idea, I did not know that. If the CORS code there is unused and was never finished, would it be simpler to remove it (the middleware, the `cors.*` keys, the `global.auth.cors*` values and the auth.md section) rather than fix the chart wiring? I could open a small issue for that separately from this RFC if you agree.

I have folded your points into the file in the latest commit: the origin of the `sda-auth` CORS code in the context, your preference under Options 2 and 3, removal as an alternative in open question 6, a new open question 9 on whether the ask is still live, and a paragraph in the suggested direction that records the other reading next to mine.

MalinAhlberg (commented):
Really thorough and good job, @jhagberg, interesting to read!

I assume that @viklund needs to give input on the actual need for this, but anyhow the RFC seems complete to me.

nanjiangshu (commented):
Thanks for drafting this detailed RFC!

I agree with Jocke regarding `sda-auth` - since the CORS code there is unused and incomplete, I'd advocate for removing it entirely along with its associated config keys, chart values, and documentation.

Also, one quick detail on the `Header Contract`: we should probably add `Content-Type` to `Access-Control-Allow-Headers`. Standard application/json payloads aren't CORS-safelisted, so any JSON requests would fail the preflight check without it.

PR #2544, opened 2026-08-26
Refactor/mapper broker v2

KarlG-nbis:
## Related issue(s) and PR(s)
This PR closes #2535 .

## Description
- Update mapper to use broker v2

- database: add ErrUniqueViolation, ErrNotNullViolation, ErrForeignKeyViolation errors,
   - Add pq error code parsing for these in the Postgres implementation

## How to test
- Unit test passes
- Integration test pass

PR #2557, opened 2026-09-02
refactor(config): extract inbox project code into its own config/v2 package

### LINKED ISSUE #2477: Extract inbox project-code config into a dedicated config/v2 package
Follow-up to #2466 (non-blocking, #2466 can merge as-is).

## Context
#2466 adds config-driven projectCode inbox paths. In review, @KarlG-nbis suggested (https://github.com/neicnordic/sensitive-data-archive/pull/2466#discussion_r3412525703) loading that config through the `config/v2` framework in its own package, so the flags register automatically on `configv2.Load()`. Review separately flagged that `internal/config/config.go` importing `internal/helper` reads as unnatural.

Both point at one root cause: the `InboxProjectConfig` type lives in `internal/helper`, which forces `config.go` to import `helper`, and the config is loaded via a hand-called `LoadInboxProjectConfig()` rather than the `config/v2` flag mechanism.

## Proposed change
Split the two concerns:

- New package `internal/v2/inboxproject` (or similar name) owns the config: the `Config` type, `storage.inbox.projectCode` / `projectCodeDelimiter` flag registration via `config.RegisterFlags` in `init()`, and validation. It does **not** import `helper`.
- `helper.ResolveInboxPath` stays in `helper` (it's a path util) but takes primitives `(filePath, username, projectCode, delimiter string)` instead of a config struct, so `helper` carries no config dependency.
- `Load() (Config, error)` keeps returning an error on invalid config (both-keys-set-together, delimiter allowlist, projectCode charset).
- Drop `Inbox` from the legacy `config.Config`, drop `LoadInboxProjectConfig`, drop the `helper` import from `config.go`.

Call sites (ingest, mapper, api) load via `inboxproject.Load()` and pass the two strings to `ResolveInboxPath`.

## Acceptance criteria
- inbox flags register through `config/v2` (show in `--help`, auto-loaded by `configv2.Load()`).
- validation logic stays the same.
- defaults unchanged: eg. empty projectCode == stock inbox layout.

yasinmiran:
Closes #2477.

## The refactor

`InboxProjectConfig` was squatting in `internal/helper`, which forced `internal/config` to import `helper` and mixed up where the config lives with where the path function lives. `internal/v2/inboxproject` now owns the type, registers `storage.inbox.projectCode` and `storage.inbox.projectCodeDelimiter` with `config/v2` in `init()`, and validates them in `Load()`. `helper.ResolveInboxPath` stays put and takes the code and delimiter as two strings, so `helper` has no config dependency left. ingest, mapper and api call `inboxproject.Load()` and pass the strings down.

Against the acceptance criteria: both flags now appear in `--help` on ingest, mapper and api, and not on services that do not import the package. Validation moved across unchanged. A deployment that sets neither key still gets the stock inbox layout, since an empty project code makes `ResolveInboxPath` delegate to `UnanonymizeFilepath` exactly as before.

Two deviations from @KarlG-nbis's sketch on #2466, both deliberate. The keys keep their shipped camelCase spelling rather than becoming `project_code` / `project_code_delimiter`, because renaming them breaks every config written against v3.1.73. And the package is `inboxproject` rather than `inbox_project_config`, underscores not being idiomatic for Go package names. Worth saying out loud: the `internal/v2/` grouping comes from the issue, but every other v2 package here is `internal/<domain>/v2`, so if you would rather this lived somewhere else, say the word and I will move it.

## The `Load()` shape, and why it reads viper when called

`AssignFunc` records that `config/v2` ran; the values themselves are read inside `Load()`. That is not the obvious shape, and it exists because mapper populates viper in two phases.

mapper calls `configv2.Load()` and then `config.NewConfig("mapper")`. The two loaders resolve the config file from different env vars, `CONFIG_FILE` for the former and `CONFIGFILE` for the latter, and `configv2.Load()` swallows a missing file. So on a deployment that supplies its config through `CONFIGFILE` alone, a value captured at assign time is empty, `Load()` returns the zero `Config` with a nil error, and the zero `Config` is a valid stock layout. Mapper then resolves stock paths and its post-mapping inbox cleanup misses every file, logging at `Errorf` and continuing. The helm chart sets both env vars so chart deployments were never exposed, and CI never sets a project code, which is why the eager version passed everything locally.

Reading at call time restores the old semantics for mapper and changes nothing for ingest and api, which are `config/v2`-only. There is a unit guard for it, and `Load()` still refuses to run before `configv2.Load()`, though that guard proves only that the flags are bound.

## The api fix

Second commit, and it stands on its own. `deleteFile` and `downloadFile` were wired to `ResolveInboxPath` in v3.1.73 and #2482 put `UnanonymizeFilepath` back while replacing gin with net/http. Nothing in that PR discusses the inbox layout, so this looks like collateral rather than a decision, but it is in v3.1.75 and v3.1.76 today.

It only bites a deployment that runs `api` with a project code configured, which as far as I can tell is nobody right now. When it does bite, `downloadFile` returns a 500. `deleteFile` is worse on s3: `DeleteObject` on a missing key succeeds, so the handler leaves its retry loop, marks the file `disabled` and answers 200 while the encrypted file stays in the inbox. posix at least fails loudly. The `downloadFile` error log also named the unresolved path rather than the one the reader was given, which is now fixed.

Heads up that #2523 rewrites the same `downloadFile` line and keeps `UnanonymizeFilepath`, so whichever order these land in, that one needs a look. #2544 touches `config.LoadInboxProjectConfig` and `helper.InboxProjectConfig`, both of which this PR removes.

## Tests

New tests cover `Load()` end to end from a nested `storage.inbox` block, the values-arrive-late case above, the registration seam, and the project-code path through ingest's catch-all branch, which is the shape a non-s3inbox deployment actually takes. Both api handlers get stock and project-code rows in one table. Each of these was checked by breaking the code and confirming the test fails, and each uses a username containing `@` so the two layouts differ by the username itself and not only by a prefix. The five `cmd/ingest` expectations that computed their own expected value by calling `ResolveInboxPath` are now literals, since the old form passed even when the function returned nonsense.

Two gaps I have not closed. The `Load()` to struct-field wiring can be deleted in any of the three services without a unit test noticing, which wants integration coverage that does not exist for this feature yet. And mapper's resolve call has no unit test at all, because `handleMessage` drives package globals and a concrete `*broker.AMQPBroker`; that seems better left to #2544 than fought here.

`go build ./...`, `go vet ./...` and `golangci-lint` are clean, and the affected packages pass under `-race`.

### DIFF
--- a/sda/CHANGELOG.md
+++ b/sda/CHANGELOG.md
@@ -21,6 +21,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - Remove redundant unit tests which are covered by other tests, and remove unit tests which should be covered by integration tests.
   - Use log/slog for logging instead of logrus.
 - Update remaining mocks to implement github.com/stretchr/testify/mock.Mock.
+- inbox project code:
+  - Moved `storage.inbox.projectCode` and `storage.inbox.projectCodeDelimiter` out of `internal/config` into `internal/v2/inboxproject`, registered through `config/v2` so both appear in `--help`. Key names, validation and the stock default are unchanged.
+  - `helper.ResolveInboxPath` takes the project code and delimiter as strings rather than a config struct, so `internal/helper` holds no config type and `internal/config` no longer imports it.
+  - Only ingest, mapper and api validate the section now. finalize, verify and intercept never read it and no longer fail at startup on a malformed one.
 
 ### Fixed
 
--- a/sda/CHANGELOG.md
+++ b/sda/CHANGELOG.md
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed downloading files by the `file_dataset.download_path` in the sda-download(v1) 
 - s3 writer: don't panic or upload to an empty bucket name when every endpoint is full
+- api: resolve the inbox path through the configured project-code layout again when deleting and downloading inbox files. The `gin` -> `net/http` rewrite in 3.1.75 reverted both call sites to the stock `UnanonymizeFilepath`, so on a deployment with `storage.inbox.projectCode` set they addressed a directory that does not exist: the download returned 500, and on an s3 inbox the delete reported success while leaving the file in place.
 
 ## [3.1.76] - 2026-07-15
 
--- a/sda/cmd/api/api.go
+++ b/sda/cmd/api/api.go
@@ -26,6 +26,7 @@ import (
 	"github.com/neicnordic/sensitive-data-archive/internal/storage/v2"
 	"github.com/neicnordic/sensitive-data-archive/internal/storage/v2/locationbroker"
 	"github.com/neicnordic/sensitive-data-archive/internal/userauth"
+	"github.com/neicnordic/sensitive-data-archive/internal/v2/inboxproject"
 	"google.golang.org/grpc"
 )
 
@@ -42,6 +43,7 @@ type API struct {
 	server      *http.Server
 	inboxReader storage.Reader
 	inboxWriter storage.Writer
+	inbox       inboxproject.Config
 	db          database.Database
 	mq          broker.Broker
 }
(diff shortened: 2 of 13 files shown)


yasinmiran:
Superseded by #2561, same two commits rebased onto current main. Reopened from a branch on this repo rather than my fork so the image builds can push to GHCR: on the fork they failed with `denied: installation not allowed to Write organization package`, which left every integration test job skipped.

PR #2561, opened 2026-09-02
refactor(config): extract inbox project code into its own config/v2 package

### LINKED ISSUE #2477: Extract inbox project-code config into a dedicated config/v2 package

yasinmiran:
Closes #2477.

## The refactor

`InboxProjectConfig` was squatting in `internal/helper`, which forced `internal/config` to import `helper` and mixed up where the config lives with where the path function lives.

`internal/inboxpath` now owns both. It registers `inboxpath.project_code` and `inboxpath.project_delimiter` with `config/v2` on import, and carries `ResolveInboxPath(filePath, username)`. Services call `inboxpath.Load()` once at startup and then resolve paths without holding any inbox configuration of their own:

```go
inboxPath := inboxpath.ResolveInboxPath(submissionFilePath, submissionUser)
```

`UnanonymizeFilepath` moved into the package too, nothing else was using it, and is unexported now. `AnonymizeFilepath` stays in `helper`, since `cmd/s3inbox/proxy.go` still calls it. `internal/config` loses the `Inbox` field, `LoadInboxProjectConfig`, and the `helper` import.

Against the acceptance criteria: both flags appear in `--help` on ingest, mapper and api, and not on services that do not import the package. Validation is unchanged in substance. A deployment that sets neither key still gets the stock inbox layout, since an empty project code takes the same branch `UnanonymizeFilepath` used to serve.

`Load() error` is the one piece of the earlier shape I kept. It is where validation lives, so a half-configured section or a delimiter outside `-` `_` `.` stops the service at startup rather than silently building a wrong directory name — the ask from round 3 of #2466 was an error rather than a panic in `AssignFunc`. It doubles as the seam the unit tests configure through, so they arrive the same way a service does.

## Breaking: the keys are renamed

| before | after |
| --- | --- |
| `storage.inbox.projectCode` | `inboxpath.project_code` |
| `storage.inbox.projectCodeDelimiter` | `inboxpath.project_delimiter` |

The old names shipped in v3.1.73. There is no aliasing, so a deployment still setting them reads as unset, which is a valid stock layout and therefore silent. The CHANGELOG says to update the config before upgrading. FEGA-Norway is the only deployment I know of that sets these, and it is tracked in ELIXIR-NO/FEGA-Norway#879.

## The api fix

Second commit, and it stands on its own. `deleteFile` and `downloadFile` were wired to `ResolveInboxPath` in v3.1.73, and #2482 put `UnanonymizeFilepath` back while replacing gin with net/http. Nothing in that PR discusses the inbox layout, so it looks like collateral rather than a decision, but it is in v3.1.75 and v3.1.76 today.

It only bites a deployment that runs `api` with a project code configured, which as far as I can tell is nobody right now. When it does bite, `downloadFile` returns a 500. `deleteFile` is worse on s3: `DeleteObject` on a missing key succeeds, so the handler leaves its retry loop, marks the file `disabled` and answers 200 while the encrypted file stays in the inbox. posix at least fails loudly. The `downloadFile` error log also named the unresolved path rather than the one the reader was given, which is now fixed.

Heads up that #2523 rewrites the same `downloadFile` line and keeps `UnanonymizeFilepath`, so whichever order these land in, that one needs a look. #2544 touches `config.LoadInboxProjectConfig` and `helper.InboxProjectConfig`, both of which this PR removes.

## Tests

New tests cover `Load()` end to end from a nested config block, the registration seam, and the project-code path through ingest's catch-all branch, which is the shape a non-s3inbox deployment actually takes. Both api handlers get stock and project-code rows in one table. Each was checked by breaking the code and confirming the test fails, and each uses a username containing `@` so the two layouts differ by the username itself and not only by a prefix. The five `cmd/ingest` expectations that computed their own expected value by calling `ResolveInboxPath` are now literals, since the old form passed even when the function returned nonsense.

Two gaps I have not closed. The `Load()` call can be deleted from any of the three services without a unit test noticing, which wants integration coverage that does not exist for this feature yet. And mapper's resolve call has no unit test at all, because `handleMessage` drives package globals and a concrete `*broker.AMQPBroker`; that seems better left to #2544 than fought here.

`go build ./...`, `go vet ./...` and `golangci-lint` are clean, and the affected packages pass under `-race`.


jhagberg:
@claude review

KarlG-nbis (changes requested):
Looks good! But I would change a few things, mainly the 

> The Load() shape, and why it reads viper when called
> mapper calls configv2.Load() and then config.NewConfig("mapper"). The two loaders resolve the config file from different env vars, CONFIG_FILE for the former and CONFIGFILE for the latter, and configv2.Load() swallows a missing file.

should not be a concern, as the config (v1) package is to be deprecated once all configuration extraction needed is placed elsewhere, and then services are expected to populate `CONFIG_FILE` instead of `CONFIGFILE`.

Another is related to the `internal/v2/inboxproject` and `internal/helper` packages, I would argue that its cleaner and more readable to instead create a package for example `internal/inboxpath` which is responsible for the configuration registration and dependent on configv2.Load(), and which has the `ResolveInboxPath` func, so something similar as:

```go
package inboxpath

...

var projectCode, projectDelimiter string

func init() {
  config.RegisterFlags(
		&config.Flag{
			Name: "inboxpath.project_code",
			...
		},
		&config.Flag{
			Name: "inboxpath.project_delimiter",
			...
		},
	)
}

func ResolveInboxPath(filePath, username string) string {
  ...

}
```
(and moving any other helper funcs just used for this purpose to this package)

This allows the application, after having called configv2.Load() to just do:

```go
inboxPath := inboxpath.ResolveInboxPath(submissionFilePath, submissionUser)
```
without having to handle the specific InboxProjectConfig struct

yasinmiran:
good call on both, pushed in 6ac08406.

`internal/inboxpath` now owns the config and `ResolveInboxPath` together, so callers just do `inboxpath.ResolveInboxPath(submissionFilePath, submissionUser)` and nobody carries a config struct around. `UnanonymizeFilepath` moved in with it since nothing else used it, and it's unexported now; `AnonymizeFilepath` stayed in helper because s3inbox still needs it. keys renamed to `inboxpath.project_code` / `inboxpath.project_delimiter` as you wrote them.

one thing i kept from the old shape: `Load() error`, called once at startup. that's where the validation lives, so a half-configured section or a delimiter outside `-` `_` `.` still refuses to boot instead of quietly building a wrong directory name. jbygdell asked for that on #2466 (error, not a panic in AssignFunc) and it seemed worth keeping now that the failure is a silent one rather than a crash. it doubles as the seam the unit tests configure through, so they go in the same way a service does.

on CONFIGFILE, fair enough, i've dropped that from the design and from the PR description. we're the deployment that sets only `CONFIGFILE` on mapper, so that's ours to fix, tracked in ELIXIR-NO/FEGA-Norway#879 along with the key rename.

worth flagging since both are yours: #2523 rewrites the same `downloadFile` line this PR fixes and keeps `UnanonymizeFilepath`, so whichever of us merges second should double check that one, otherwise the inbox path resolution goes away again. #2544 uses `config.LoadInboxProjectConfig` and `helper.InboxProjectConfig`, both gone here.

sda/internal/inboxpath/inboxpath.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	for _, r := range projectCode {
+		if r == '/' || r == '\\' || unicode.IsSpace(r) || unicode.IsControl(r) {
+			return fmt.Errorf("%s %q must not contain path separators or whitespace", codeKey, projectCode)
+		}
+	}
+
+	return nil
+}
```
KarlG-nbis: Hmm thinking of a way to include such flag validation in the configv2 package, one idea is to update the `AssignFunc func(flagName string)` -> `AssignFunc func(flagName string) error` and then we could do 

```go
[]*config.Flag{
	{
		Name: codeKey,
			flagSet.String(flagName, "", `Project code prefixing every per-user inbox directory, e.g. "p11". Empty selects the stock layout, where the directory is the normalized username`)
		},
		Required: false,
		AssignFunc: func(flagName string) error {
			projectCode = viper.GetString(flagName)

        	if projectCode != "" && viper.GetString(delimiterKey) == "" {
		       return fmt.Errorf("%s and %s must be set together", codeKey, delimiterKey)
        	}

            for _, r := range projectCode {
        		if r == '/' || r == '\\' || unicode.IsSpace(r) || unicode.IsControl(r) {
		        	return fmt.Errorf("%s %q must not contain path separators or whitespace", codeKey, projectCode)
	        	}
        	}
		},
	},
	{
		Name: delimiterKey,
			flagSet.String(flagName, "", `Character joining the project code to the username in an inbox directory name, one of "-", "_" or "."`)
		},
		Required: false,
			projectDelimiter = viper.GetString(flagName)

           	if projectDelimiter != "" && viper.GetString(codeKey) == "" {
		       return fmt.Errorf("%s and %s must be set together", codeKey, delimiterKey)
        	}

            switch projectDelimiter {
           	case "-", "_", ".":
	        default:
       	    	return fmt.Errorf("%s %q is not a delimiter character (use \"-\", \"_\" or \".\")", delimiterKey,  projectDelimiter)
        	}
            
		},
	},
}
```

with the following change in config/v2/config.go

```go
func Load() error {
...
  var assignErrors error
  for _, flag := range registeredFlags {
    if err := flag.AssignFunc(flag.Name); err != nil {
      assignErrors = errors.Join(assignErrors, fmt.Errorf("failed to assign %s: %s", flag.name, err.Error())
    }
  }

  if assignErrors != nil {
    return err
  }
}
```

What do you think? Could be done as a separate PR as it will require changes to all flag registrations, and we have some similar already [here](https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/cmd/ingest/config/config.go#L53) but instead it panics if validation fails which isnt ideal

As imo the Load() func here I think could lead to some easy mistakes if a new service needs this, and developer forgets to add the Load() for example, and having a more clear way to add flag value validation makes sense to me

KarlG-nbis (approved):
Looks good! Please take a look a my [comment](https://github.com/neicnordic/sensitive-data-archive/pull/2561#discussion_r3931569611) thou and say what you think!

Parisa68 (approved):

PR #2562, opened 2026-09-03
docs(adr): ADR-0007 use Conventional Commits for commit messages

jhagberg:
## Related issue(s) and PR(s)

No issue. The team already agreed on this, so this PR writes it down and adds the checks.

## Description

ADR-0007 records that commit messages follow Conventional Commits. It is marked `accepted` since the decision is already taken.

The PR also adds the enforcement:

- `.githooks/commit-msg` checks the subject line locally. Enable it once per clone with `git config core.hooksPath .githooks`.
- A new workflow runs that same script on every non-merge commit of a PR against `main` and fails on `fixup!` and `squash!` commits. We rebase-merge, so the PR title never reaches `main` and each commit has to be right on its own.
- `CONTRIBUTING.md` describes the format and the hook and drops the old "50 characters, capitalised" rules. The last commit fixes a few markdownlint nits in that file and repoints two links that still went to the old neic-sda repo.

Of the last 200 commits on `main`, 5 would fail the check. It stays advisory until an admin adds "Check commit messages" as a required status check in the `main` ruleset. Should we do that right away, or wait until the hook has been in use for a while?

## ADR

- [x] This PR includes an architecturally significant decision (see `docs/decisions/README.md`)
  - [x] A decision record has been added or updated in `docs/decisions/`
  - [x] The decision record is listed in the `docs/decisions/README.md` index table

## How to test

The "Conventional commits" workflow runs on this PR. Locally:

```sh
git config core.hooksPath .githooks
git commit --allow-empty -m "fix review comments"   # rejected
git commit --allow-empty -m "docs: try the hook"    # accepted
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,19 +54,58 @@ Each commit should contain changes that are functionally connected and/or relate
 > NOTE:
 > The commands `git add -p` or `git commit -p` can prove useful for selecting chunks of changed files in order to group unrelated things into multiple separate commits.
 
+#### Commit message format
+
+Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/), see [ADR-0007](docs/decisions/0007-use-conventional-commits.md):
+
+```text
+<type>[optional scope][!]: <description>
+
+<optional body>
+
+<optional footer(s)>
+```
+
+- `type` is one of `build`, `chore`, `ci`, `docs`, `feat`, `fix`, `perf`, `refactor`, `revert`, `style`, `test`.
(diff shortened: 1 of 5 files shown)


jbygdell (approved):

KarlG-nbis (approved):

MalinAhlberg (approved):
:+1:

CONTRIBUTING.md
MalinAhlberg:
```suggestion
- `feat/`
```

jhagberg: Good call, thanks. I took it one step further so the whole list matches the commit types: `bugfix/` is now `fix/` as well, and the intro links to the commit message section instead of keeping a separate list. Applied in fa17aa9f with you as co-author.

PR #2565, opened 2026-09-03
Add RabbitMQ configuration to disable queue creation at start up

### LINKED ISSUE #2564: Disable auto-creation of rabbit mq queue on sda-doa startup
Currently, the default behavior of the application is to init a rabbit mq queue on start up. This is somewhat undesired given that the queues already exists in RMQ.
However, since the current setup works with this assumption, the disabling of this start up process should be configurable via environment variables so that the behavior can be disabled when desired and defaulting to its current behavior otherwise.

joshbaskaran:
## Related issue(s) and PR(s)
This PR closes [ #2564].

## Description
Make the RabbitMQ configurable via environment vars including the option to disable queue creation at startup

## ADR

### DIFF
--- a/sda-doa/README.md
+++ b/sda-doa/README.md
@@ -22,7 +22,7 @@ Environment variables used:
 | OUTBOX_ENABLED                         | true                                                                 | Enables/disables the outbox functionality          |
 | OUTBOX_TYPE                            | POSIX                                                                | Outbox type: `POSIX` or `S3`                       |
 | OUTBOX_QUEUE                           | exportRequests                                                       | MQ queue name for files/datasets export requests   |
-| OUTBOX_QUEUE_AUTO_CREATE               | true                                                                 | Automatically create the queue; when false, the queue must be provisioned externally |
+| OUTBOX_QUEUE_AUTO_CREATE               | true                                                                 | Automatically create an exclusive and non-durable queue; when false, the queue must be provisioned externally |
 | OUTBOX_LOCATION                        | /ega/outbox/p11-%s/files/                                            | Outbox location with placeholder for the username  |
 | BROKER_HOST                            | private-mq                                                           | Local RabbitMQ broker hostname                     |
 | BROKER_PORT                            | 5671                                                                 | Local RabbitMQ broker port                         |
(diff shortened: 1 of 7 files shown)


sda-doa/src/main/java/no/uio/ifi/localega/doa/mq/RabbitMQConfiguration.java
```diff
@@ -0,0 +1,18 @@
...
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ConditionalOnProperty(name = "outbox.enabled", havingValue = "true")
+public class RabbitMQConfiguration {
+
+    @Bean
+    @ConditionalOnProperty(name = "outbox.queue-auto-create", havingValue = "true", matchIfMissing = true)
+    public Queue exportQueue(@Value("${outbox.queue}") String queueName) {
+        return new Queue(queueName, false, true, true);
```
kjellp: Having the 3 last parameters meaning/incluence on the queue creation visible in the code would be good, since this line replaces the deleted lines in the ExportRequestListener.java file where it's self-explanatory. I noticed this when reviewing to see if the "exclusive" and autodelete features were kept from the previous implementation :)

sda-doa/README.md
```diff
@@ -22,6 +22,7 @@ Environment variables used:
 | OUTBOX_ENABLED                         | true                                                                 | Enables/disables the outbox functionality          |
 | OUTBOX_TYPE                            | POSIX                                                                | Outbox type: `POSIX` or `S3`                       |
 | OUTBOX_QUEUE                           | exportRequests                                                       | MQ queue name for files/datasets export requests   |
+| OUTBOX_QUEUE_AUTO_CREATE               | true                                                                 | Automatically create the queue; when false, the queue must be provisioned externally |
```
kjellp: Suggest: Automatically create an exclusive and non-durable queue; ...

sda-doa/src/main/java/no/uio/ifi/localega/doa/mq/RabbitMQConfiguration.java
joshbaskaran: Java doesn't support named parameters for constructor calls, so I've introduced local variables of the same name to make it explicit. This does add a bit of redundant code, though. Let me know if this works or if I should remove the local variables!

kjellp (approved):
LGTM now :) (Could also consider updating some deps to fix security here instead of dependabot PRs later? Ref last cooldown discussions in sda-collab meeting)

joshbaskaran:
@kjellp, I think it would be better to have that in a separate PR since it's unrelated to this change. I have setup a PR updating springboot version #2568 which resolves many of the security issues. I expect springboot 4.2.0 to be released in the next few days which should resolve another set of issues as well. Will keep an eye out for that as well!

Parisa68 (approved):

PR #2571, opened 2026-09-07
Refactor/intercept broker v2

### LINKED ISSUE #2570: [intercept] Implement the broker v2 package

KarlG-nbis:
## Related issue(s) and PR(s)
This PR closes #2570 

### sda
- intercept:
  - Update to use broker/v2 instead of broker (v1)
  - Refactor and add unit tests
  - Add configuration options for destination routing keys

### sda-svc
- Populate broker_v2 config for intercept

## How to test
- unit & integration tests pass

### DIFF
--- a/sda/CHANGELOG.md
+++ b/sda/CHANGELOG.md
@@ -29,6 +29,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - verify:
   - Update to use broker/v2 instead of broker (v1)
   - Add unit tests
+- intercept:
+  - Update to use broker/v2 instead of broker (v1)
+  - Refactor and add unit tests
+  - Add configuration options for destination routing keys
 
 ### Fixed
 
--- a/sda/cmd/intercept/config/config.go
+++ b/sda/cmd/intercept/config/config.go
@@ -1,0 +1,114 @@
+package config
+
+import (
+	"github.com/neicnordic/sensitive-data-archive/internal/config/v2"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+)
+
+var (
+	sourceQueue         string
+	accessionRoutingKey string
+	cancelRoutingKey    string
+	ingestRoutingKey    string
+	mappingRoutingKey   string
+	releaseRoutingKey   string
+	deprecateRoutingKey string
+)
+
+func init() {
+	config.RegisterFlags(
+		&config.Flag{
+			Name: "source_queue",
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.String(flagName, "from_cega", "The queue where the verify service consumes messages from")
+			},
+			Required: false,
+			AssignFunc: func(flagName string) {
+				sourceQueue = viper.GetString(flagName)
+			},
+		},
+		&config.Flag{
+			Name: "accession_routing_key",
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.String(flagName, "accession", "The routing key the intercept service forwards \"accession\" type messages to")
+			},
+			Required: false,
+			AssignFunc: func(flagName string) {
+				accessionRoutingKey = viper.GetString(flagName)
+			},
+		},
+		&config.Flag{
+			Name: "ingest_routing_key",
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.String(flagName, "ingest", "The routing key the intercept service forwards \"ingest\" type messages to")
+			},
+			Required: false,
+			AssignFunc: func(flagName string) {
+				ingestRoutingKey = viper.GetString(flagName)
+			},
+		},
+		&config.Flag{
+			Name: "cancel_routing_key",
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.String(flagName, "ingest", "The routing key the intercept service forwards \"cancel\" type messages to")
+			},
+			Required: false,
+			AssignFunc: func(flagName string) {
+				cancelRoutingKey = viper.GetString(flagName)
+			},
+		},
+		&config.Flag{
+			Name: "mapping_routing_key",
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.String(flagName, "mappings", "The routing key the intercept service forwards \"mapping\" type messages to")
+			},
+			Required: false,
+			AssignFunc: func(flagName string) {
+				mappingRoutingKey = viper.GetString(flagName)
+			},
+		},
+		&config.Flag{
+			Name: "release_routing_key",
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.String(flagName, "mappings", "The routing key the intercept service forwards \"release\" type messages to")
+			},
+			Required: false,
+			AssignFunc: func(flagName string) {
+				releaseRoutingKey = viper.GetString(flagName)
+			},
+		},
+		&config.Flag{
+			Name: "deprecate_routing_key",
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.String(flagName, "mappings", "The routing key the intercept service forwards \"deprecate\" type messages to")
+			},
+			Required: false,
+			AssignFunc: func(flagName string) {
+				deprecateRoutingKey = viper.GetString(flagName)
+			},
+		},
+	)
+}
+
+func SourceQueue() string {
+	return sourceQueue
+}
+func AccessionRoutingKey() string {
+	return accessionRoutingKey
+}
+func CancelRoutingKey() string {
+	return cancelRoutingKey
+}
+func IngestRoutingKey() string {
+	return ingestRoutingKey
+}
+func MappingRoutingKey() string {
+	return mappingRoutingKey
+}
+func ReleaseRoutingKey() string {
+	return releaseRoutingKey
+}
+func DeprecateRoutingKey() string {
+	return deprecateRoutingKey
+}
--- a/sda/cmd/intercept/intercept.go
+++ b/sda/cmd/intercept/intercept.go
@@ -3,109 +3,147 @@
 package main
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
+	"log/slog"
+	"os"
+	"os/signal"
+	"syscall"
 
-	"github.com/neicnordic/sensitive-data-archive/internal/broker"
-	"github.com/neicnordic/sensitive-data-archive/internal/config"
+	interceptconfig "github.com/neicnordic/sensitive-data-archive/cmd/intercept/config"
+	broker "github.com/neicnordic/sensitive-data-archive/internal/broker/v2"
+	"github.com/neicnordic/sensitive-data-archive/internal/broker/v2/rabbitmq"
+	configv2 "github.com/neicnordic/sensitive-data-archive/internal/config/v2"
 
 	log "github.com/sirupsen/logrus"
 )
 
+type messageType string
+
 const (
-	msgAccession string = "accession"
-	msgCancel    string = "cancel"
-	msgIngest    string = "ingest"
-	msgMapping   string = "mapping"
-	msgRelease   string = "release"
-	msgDeprecate string = "deprecate"
+	messageTypeAccession messageType = "accession"
+	messageTypeCancel    messageType = "cancel"
+	messageTypeIngest    messageType = "ingest"
+	messageTypeMapping   messageType = "mapping"
+	messageTypeRelease   messageType = "release"
+	messageTypeDeprecate messageType = "deprecate"
 )
 
+type intercept struct {
+	broker  broker.Broker
+	routing map[messageType]string
+}
+
 func main() {
-	forever := make(chan bool)
-	conf, err := config.NewConfig("intercept")
-	if err != nil {
+	if err := run(); err != nil {
 		log.Fatal(err)
 	}
-	mq, err := broker.NewMQ(conf.Broker)
-	if err != nil {
-		log.Fatal(err)
+}
+
+func run() error {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	if err := configv2.Load(); err != nil {
+		return fmt.Errorf("failed to load config: %v", err)
 	}
 
-	defer mq.Channel.Close()
-	defer mq.Connection.Close()
+	app := &intercept{
+		routing: map[messageType]string{
+			messageTypeAccession: interceptconfig.AccessionRoutingKey(),
+			messageTypeCancel:    interceptconfig.CancelRoutingKey(),
+			messageTypeIngest:    interceptconfig.IngestRoutingKey(),
+			messageTypeMapping:   interceptconfig.MappingRoutingKey(),
+			messageTypeRelease:   interceptconfig.ReleaseRoutingKey(),
+			messageTypeDeprecate: interceptconfig.DeprecateRoutingKey(),
+		},
+	}
 
-	go func() {
-		connError := mq.ConnectionWatcher()
-		log.Error(connError)
-		forever <- false
-	}()
+	var err error
+	app.broker, err = rabbitmq.NewRabbitMQBroker(ctx)
+	if err != nil {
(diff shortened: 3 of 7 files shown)


KarlG-nbis:
@claude review

sda/cmd/intercept/intercept.go
```diff
@@ -3,117 +3,153 @@
...
+	}
+
+	slog.Info(
+		"Routing message",
+		slog.String("message-key", message.Key),
+		slog.String("message-type", string(msgType)),
+		slog.String("routing-key", routingKey),
+	)
+	if err := app.broker.Publish(ctx, routingKey, *message); err != nil {
+		log.Errorf("failed to publish message, reason: (%v)", err)
```
jbygdell:
```suggestion
		slog.Errorf("failed to publish message, reason: (%v)", err)
```

We should only use one logging library

sda/CHANGELOG.md
```diff
@@ -29,6 +29,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - verify:
   - Update to use broker/v2 instead of broker (v1)
   - Add unit tests
+- intercept:
+  - Update to use broker/v2 instead of broker (v1)
+  - Refactor and add unit tests
+  - Add configuration options for destination routing keys
```
jhagberg: Anchoring here since `intercept.md` is not in the diff. The doc seems to have been missed in the migration: line 49 still documents `BROKER_QUEUE`, which is now the top-level `SOURCE_QUEUE` (config.go:22-30), the six new `*_ROUTING_KEY` options from this changelog line are not mentioned, and the note at line 14 that the service has "no error handling" no longer holds, since intercept.go:138 returns the Publish error and broker/v2 requeues on that.

Could the doc be updated here? It might also be worth exposing the routing-key options in `intercept-secrets.yaml`, which after the earlier thread sets `source_queue` but none of the `*_routing_key` keys. As far as I can see they are not reachable through the chart's normal values, only by supplying a whole config file via `global.vaultSecrets`.

sda/cmd/intercept/intercept.go
```diff
@@ -3,117 +3,155 @@
...
+func typeFromMessage(body []byte) (messageType, error) {
 	message := make(map[string]any)
 	err := json.Unmarshal(body, &message)
 	if err != nil {
 		return "", err
 	}
 
 	msgTypeFetch, ok := message["type"]
-	if !ok {
+	if !ok || msgTypeFetch == "" {
```
jhagberg: Was the `|| msgTypeFetch == ""` here meant as a behaviour change? With it, a message with an empty `type` becomes a parse error that is logged, acked and dropped (lines 106-113). The previous version treated it like any unknown type and forwarded it to the `undeliverable` routing key (base intercept.go:81-91).

If parity with v1 is the goal, dropping the `== ""` condition would keep such messages somewhere an operator can find them.

sda/cmd/intercept/intercept.go
```diff
@@ -3,117 +3,155 @@
...
-				log.Errorf("failed to ack message for reason: %v", err)
-			}
+	sigc := make(chan os.Signal, 1)
+	signal.Notify(sigc, os.Interrupt, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
+
+	select {
+	case sig := <-sigc:
+		slog.Info("received signal, shutting down gracefully", "signal", sig)
+
+		return nil
```
jhagberg: The signal branch returns here without calling `cancel()` first. The deferred `cancel()` at line 47 still runs, but only after the deferred `broker.Close()` at lines 70-77, so the broker is closed while the Subscribe goroutine is still on a live context. verify.go:112 and mapper.go:108 call `cancel()` in the signal branch before returning. Should intercept do the same?

sda/cmd/intercept/intercept_test.go
```diff
@@ -1,127 +1,237 @@
...
-	assert.Nil(ts.T(), err, "Unexpected error from typeFromMessage")
-	assert.Equal(ts.T(), msgType, msgAccession, "message type from message does not match expected")
-}
-
-func (ts *TestSuite) TestMessageSelection_Cancel() {
-	msg := ingest{
-		Type:     "cancel",
-		User:     "foo",
-		FilePath: "/tmp/foo",
+	for _, tc := range []testCase{
```
jhagberg: There is no table case where `Publish` fails, so the new nack-and-requeue path (the Publish error being returned at intercept.go:138) has nothing guarding it; every `newMock` returns nil from `Publish`. Would one be easy to add?

sda/cmd/intercept/config/config.go
```diff
@@ -0,0 +1,114 @@
...
+	releaseRoutingKey   string
+	deprecateRoutingKey string
+)
+
+func init() {
+	config.RegisterFlags(
+		&config.Flag{
+			Name: "source_queue",
+			RegisterFunc: func(flagSet *pflag.FlagSet, flagName string) {
+				flagSet.String(flagName, "from_cega", "The queue where the verify service consumes messages from")
```
jhagberg: Small one: the help text says "the verify service". It shows up in `--help`, so probably worth fixing while here.

sda/cmd/intercept/intercept.go
```diff
@@ -3,117 +3,155 @@
...
+		routingKey = "undeliverable"
+	}
+
+	slog.Info(
+		"Routing message",
+		slog.String("message-key", message.Key),
+		slog.String("message-type", string(msgType)),
+		slog.String("routing-key", routingKey),
+	)
+	if err := app.broker.Publish(ctx, routingKey, *message); err != nil {
```
jhagberg: One behaviour change I wanted to check is intended: `Publish` now relays the incoming message's headers, where the v1 `SendMessage` published an empty header table (base broker.go:202). I do not see a problem with it, just noting it in case something downstream keys on headers.

jhagberg (commented):
Thanks for the migration and for the routing-key options. It is a small diff, so only a handful of comments and most of them questions. Two of them follow on from the earlier threads: the chart got `source_queue` but not the routing keys, and the shutdown ordering in `run()`.

sda/CHANGELOG.md
KarlG-nbis: Yeah will update the docs, but for exposing the previously hardcoded routing keys on the sda-svc chart I think that should be a future issue as tbh dont want to deal with that as part of this pr, given the chart doesnt even have any routing key value configuration defined atm

sda/cmd/intercept/intercept.go
KarlG-nbis: good point, did not think about that behavioural change, will revert it

sda/cmd/intercept/intercept.go
KarlG-nbis: Yeah the idea is that the deferred cancel is intended and the mapper and verify should be updated, the reasoning is that we want to terminate gracefully and by cancelling the context before closing broker mean that any inflight messages, etc fail when they could have completed before closing

sda/cmd/intercept/intercept_test.go
KarlG-nbis: yeah, will add!

sda/cmd/intercept/intercept.go
KarlG-nbis: Yeah I think it makes sense to forward the headers given the function of this service is to forward messages

sda/cmd/intercept/intercept.go
jhagberg: I see the intent, and I agree that cancelling first, as verify and mapper do, cuts an in-flight handler short: amqp091-go checks `ctx.Done()` on entry to `PublishWithContext` (channel.go:1886 in v1.14.0), so the handler's `Publish` fails with `context canceled` and the message is nacked and requeued.

I am less sure the current order gets there either. The two defers run in reverse order, so on the signal path `broker.Close()` (line 70) runs before `cancel()` (line 47), and nothing reads `consumeErr` before `run` returns. I tried it against a real RabbitMQ with a handler blocked mid-flight. With `Close()` first and the ctx still live, the handler's `Publish` goes through `ensureConnected`, reconnects and publishes, the Ack on the old channel fails, and the broker redelivers the message, so it was processed twice. With `cancel()` landing right after `Close()` the Publish fails with `context canceled` and the message is requeued, the same result as verify and mapper. So neither order guarantees that the in-flight message completes; it comes down to timing. Two smaller things I noticed on the way: `Close()` (rabbitmq.go:117-137) does not take `b.mu` while `connect()` replaces the pointers under it, so `Close()` before `cancel()` can run concurrently with a reconnect, and the `consumeChannel.Cancel("", true)` in `consumeMessages` (:199) cancels a consumer tag that does not exist, since `Consume` was given "" and the client generated its own tag (channel.go:1412).

If the goal is to let the in-flight message finish, I think that needs the broker rather than the ordering in `run`. I put a small change for it in #2587 as a draft against main: the handler runs on `context.WithoutCancel(ctx)` with a `broker.shutdown_grace` period, the consumer is cancelled by its real tag, and `Close()` refuses to reconnect afterwards. With that in, this PR and rotatekey would only need `cancel()` before the return and a wait on `consumeErr`, the same as ingest does there. Until then I would keep cancel first in verify and mapper, since it requeues through a live channel and avoids the reconnect race. I may well be missing something in how you see the shutdown though, so feel free to take the draft over or push back on it.

KarlG-nbis: hmm yeah might require a more proper investigation / fix, so yeah Ill update to keep the cancel() before the closing for now.

jhagberg: Sounds good. #2579 (8c416422) and #2563 (8091563b) now have the same `cancel()` before the return, so the three services line up, and the proper drain can wait for #2587.

sda/CHANGELOG.md
jhagberg: Agreed, the chart part is out of scope here. Should it go on the #2459 list so it is not lost, or do you want to open the issue? On the doc: f3d872e3 covers `SOURCE_QUEUE` and the routing keys, but intercept.md:14 still says the service has no error handling.

sda/cmd/intercept/intercept.go
jhagberg: One small thing before I approve: the branch at f5aa027a still returns from the signal case at `intercept.go:89-92` without the explicit `cancel()` that we agreed on, so it relies on the deferred one after `broker.Close()`. Was that intended to come in a later push, or did it get lost in the rebase? Everything else on this PR looks good to me.

nanjiangshu (commented):
This PR has already been reviewed extensively and the current shape looks good to me. However, I do have one question.

The new intercept routing-key options are added in code, but the corresponding helm chart template `intercept-secrets.yaml` does not render them.
That means Helm deployments cannot configure  `accession_routing_key`,  `ingest_routing_key`,  `cancel_routing_key`,  `mapping_routing_key`,  `release_routing_key`, or  `deprecate_routing_key`, and will always use defaults.

 If we want to configure through the helm, it is better to define there as well.

jhagberg:
Thanks @nanjiangshu, that is the same gap I asked about in [this thread](https://github.com/neicnordic/sensitive-data-archive/pull/2571#discussion_r3979711952). @KarlG-nbis and I agreed there that exposing the routing keys in `intercept-secrets.yaml` is out of scope for this PR, since the chart has no routing-key values for any service today and this would be the first one. What was left open is where to track it so it is not lost. #2459 has the "All chart templates are up to date" checkbox, but it may deserve an issue of its own covering all the migrated services, not only intercept. Karl, would you rather open that issue or shall I?

KarlG-nbis:
> This PR has already been reviewed extensively and the current shape looks good to me. However, I do have one question.
> 
> The new intercept routing-key options are added in code, but the corresponding helm chart template `intercept-secrets.yaml` does not render them. That means Helm deployments cannot configure  `accession_routing_key`,  `ingest_routing_key`,  `cancel_routing_key`,  `mapping_routing_key`,  `release_routing_key`, or  `deprecate_routing_key`, and will always use defaults.
> 
> If we want to configure through the helm, it is better to define there as well.



> Thanks @nanjiangshu, that is the same gap I asked about in [this thread](https://github.com/neicnordic/sensitive-data-archive/pull/2571#discussion_r3979711952). @KarlG-nbis and I agreed there that exposing the routing keys in `intercept-secrets.yaml` is out of scope for this PR, since the chart has no routing-key values for any service today and this would be the first one. What was left open is where to track it so it is not lost. #2459 has the "All chart templates are up to date" checkbox, but it may deserve an issue of its own covering all the migrated services, not only intercept. Karl, would you rather open that issue or shall I?

Yeah the added configuration options were just a step along the way to improve this, as we have similar issues in the API service where alot of routing keys are hardcoded, and current in the sda-svc chart you have the ability to configure *SOME* routing key names, but not all (given the hardcoded ones), so Ill create an issue to cover the remaning items that need fixing, as this PR never intended to fix all of this, but given the routing keys were related to the broker_v2, I added to option for configuring them on the application level

nanjiangshu:
> > This PR has already been reviewed extensively and the current shape looks good to me. However, I do have one question.
> > The new intercept routing-key options are added in code, but the corresponding helm chart template `intercept-secrets.yaml` does not render them. That means Helm deployments cannot configure  `accession_routing_key`,  `ingest_routing_key`,  `cancel_routing_key`,  `mapping_routing_key`,  `release_routing_key`, or  `deprecate_routing_key`, and will always use defaults.
> > If we want to configure through the helm, it is better to define there as well.
> 
> > Thanks @nanjiangshu, that is the same gap I asked about in [this thread](https://github.com/neicnordic/sensitive-data-archive/pull/2571#discussion_r3979711952). @KarlG-nbis and I agreed there that exposing the routing keys in `intercept-secrets.yaml` is out of scope for this PR, since the chart has no routing-key values for any service today and this would be the first one. What was left open is where to track it so it is not lost. #2459 has the "All chart templates are up to date" checkbox, but it may deserve an issue of its own covering all the migrated services, not only intercept. Karl, would you rather open that issue or shall I?
> 
> Yeah the added configuration options were just a step along the way to improve this, as we have similar issues in the API service where alot of routing keys are hardcoded, and current in the sda-svc chart you have the ability to configure _SOME_ routing key names, but not all (given the hardcoded ones), so Ill create an issue to cover the remaning items that need fixing, as this PR never intended to fix all of this, but given the routing keys were related to the broker_v2, I added to option for configuring them on the application level

Thanks for the clarification!
I’m okay with this being incremental if the goal of this PR is only to add the application-level routing-key support for the broker/v2 migration. Would be good to open a following issue if not opened yet.

nanjiangshu (approved):

KarlG-nbis:
> > > This PR has already been reviewed extensively and the current shape looks good to me. However, I do have one question.
> > > The new intercept routing-key options are added in code, but the corresponding helm chart template `intercept-secrets.yaml` does not render them. That means Helm deployments cannot configure  `accession_routing_key`,  `ingest_routing_key`,  `cancel_routing_key`,  `mapping_routing_key`,  `release_routing_key`, or  `deprecate_routing_key`, and will always use defaults.
> > > If we want to configure through the helm, it is better to define there as well.
> > 
> > 
> > > Thanks @nanjiangshu, that is the same gap I asked about in [this thread](https://github.com/neicnordic/sensitive-data-archive/pull/2571#discussion_r3979711952). @KarlG-nbis and I agreed there that exposing the routing keys in `intercept-secrets.yaml` is out of scope for this PR, since the chart has no routing-key values for any service today and this would be the first one. What was left open is where to track it so it is not lost. #2459 has the "All chart templates are up to date" checkbox, but it may deserve an issue of its own covering all the migrated services, not only intercept. Karl, would you rather open that issue or shall I?
> > 
> > 
> > Yeah the added configuration options were just a step along the way to improve this, as we have similar issues in the API service where alot of routing keys are hardcoded, and current in the sda-svc chart you have the ability to configure _SOME_ routing key names, but not all (given the hardcoded ones), so Ill create an issue to cover the remaning items that need fixing, as this PR never intended to fix all of this, but given the routing keys were related to the broker_v2, I added to option for configuring them on the application level
> 
> Thanks for the clarification! I’m okay with this being incremental if the goal of this PR is only to add the application-level routing-key support for the broker/v2 migration. Would be good to open a following issue if not opened yet.

Opened #2598 for this now

jhagberg (approved):
Thanks Karl, this looks good to me now. The points from my review are all handled: the explicit `cancel()` with the wait on `consumeErr` and the second-signal escape follows the same pattern as #2587, the routing-key options are in place on the application level, and the chart gap is tracked in #2598. Approving.

PR #2575, opened 2026-09-07
refactor: finish accessionID naming per ADR-0008

### LINKED ISSUE #1854: Consistent with names in db functions
As a developer I want to have the same names for inputs in the functions
so that it is clear what it is used.

For example in the `db_functions.go` the file id (UUID) is used sometimes as `fileID` and some other times as `fileUUID`.
Another example is the stable id (aa-File... for BP) is used sometimes as `accession`, other times as `accessionID` and some other times as `stableID`.

I guess that this inconsistency exists also for other inputs.

jhagberg:
## Related issue(s) and PR(s)

Closes #1854. Implements ADR-0008 (#2574), promoted from RFC-0001 (#2263). This should merge after #2574.

## Description

The last identifiers that did not follow the naming in ADR-0008, one commit per module:

- `sda-download`: `GetDatasetFileStableID`, `getDatasetFileStableID`, `getFileStableID` and the `fileStableID` variables become `...AccessionID`. Nothing here is exposed in JSON or in the API; the S3 handler still passes the value on as the `fileid` parameter.
- `sda-doa`: the `stableId` field on the `Dataset` entity becomes `accessionId`, so the Lombok getter becomes `getAccessionId()`, and the two call sites in the export listener follow. The `@Column(name = "stable_id")` mapping is unchanged, so no schema or query change.
- `sda`: the `fileUUID` parameter in `updateFileEventLog` becomes `fileID` like everywhere else, the bare `accession` loop variable in the dataset files query becomes `accessionID`, and a test literal spelled `stableID-disabled` is renamed so the grep below stays clean. That covers the `fileID` / `fileUUID` example in #1854 as well.
- Docs and dev-tools: the syncapi README and the key-rotation demo output said "stableID"; now "accession ID". Codex review pointed these out.

No behaviour changes and no SQL changes.

## ADR

No new decision. This PR implements ADR-0008 from #2574.

## How to test

`go build ./...`, `go vet ./...` and `golangci-lint run` pass in `sda-download` and `sda`, and `go test ./api/s3/` passes in `sda-download`. The Confirmation check from ADR-0008 comes back empty:

```sh
grep -rn --include='*.go' -e stableID -e StableID .
```

I have no Maven locally, so the `sda-doa` part relies on the CodeQL build and the sda-doa integration tests in CI.


jhagberg:
@claude review always

jbygdell (approved):

Parisa68 (approved):

